Issue #19, opened 2019-06-05
Support Rest Parameters


saelo:
FuzzIL should have support for functions with [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters).

PR #21, opened 2019-06-13
add rest param to functions

wbowling:
fixes https://github.com/googleprojectzero/fuzzilli/issues/19

Changes `defineFunction` to randomly create functions with a rest parameter

### DIFF
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -198,7 +198,12 @@ public class JavaScriptLifter: ComponentBase, Lifter {
                 output = BinaryExpression.new() <> input(0) <> " in " <> input(1)
                 
             case let op as BeginFunctionDefinition:
-                let params = instr.innerOutputs.map({ $0.identifier }).joined(separator: ",")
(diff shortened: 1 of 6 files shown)


Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -198,7 +198,16 @@ public class JavaScriptLifter: ComponentBase, Lifter {
                 output = BinaryExpression.new() <> input(0) <> " in " <> input(1)
                 
             case let op as BeginFunctionDefinition:
```
saelo: I think you can simplify this code a bit, e.g.:
```
var identifiers = instr.innerOutputs.map({ $0.identifier })
if op.hasRestParam, let last = instr.innerOutputs.last {
    identifiers[identifiers.endIndex - 1] = "..." +last.identifier
}
let params = identifiers.joined(separator: ",")``` or something like that
Also no need for `;` at the end or brackets around conditions in an `if` :)

wbowling: Ah cool, haven't really done much swift so good to know!

saelo:
Thanks!

PR #22, opened 2019-06-13
export the corpus on shutdown

wbowling:
If `exportCorpus` is set then save the current state when closing the fuzzer.


Sources/Fuzzilli/Modules/Storage.swift
```diff
@@ -92,4 +81,17 @@ public class Storage: Module {
             logger.error("Failed to write program to disk: \(error)")
         }
     }
+
+    private func saveCorpus() {
```
saelo: This function should now probably be called "saveState" :)

wbowling: Ah yep :)

saelo:
Thanks! :)

PR #25, opened 2019-06-25
Add ChakraCore as a target

wbowling:
Based on the other targets with a couple of changes to ChakraCore to allow for reprl mode.

Not sure what the best options to `ch` are, so picked some that a few of the chakra test cases use but may need to be tweaked.

The `Execs / Second` is pretty low and a higher timeout is required, there might be a few more changed that could be make to the patch to speed this up

### DIFF
--- a/Sources/FuzzilliCli/Profiles/ChakraProfile.swift
+++ b/Sources/FuzzilliCli/Profiles/ChakraProfile.swift
@@ -1,0 +1,64 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


wbowling:
After a bit more fuzzing I'm getting a fair bit of unreproducible flakiness with the patch, might need to change a few things to get it stable.

Also building ChakraCore on Ubuntu with clang-6 was segfaulting when enabling coverage, works fine with clang-8

wbowling:
I think I know what the issue is, I’ll reopen a PR once I’ve tested it a bit more and it’s stable

marckwei:
> I think I know what the issue is, I’ll reopen a PR once I’ve tested it a bit more and it’s stable

Have you done the above? I'm having trouble trying to fuzz ChakraCore with fuzzilli, could you please share your patch file?

Thanks a lot！

Issue #27, opened 2019-06-27
Import corpus without state


wbowling:
Could `evaluatorState` be made optional or add a flag to ignore an incompatible state when importing?

It can be useful to import a previous corpus when testing a new version

saelo:
Hmm good point. So another possible approach I see is to slightly change the logic of Fuzzer.importState to do the following:
* try to import the evaluator state
* if that succeeds, try importing the corpus state
* if any of the above fail, print a info or warning message, potentially revert any imported state, and import the corpus via Fuzzer.importCorpus (which will execute every sample in it)
That way we wouldn't need new cli flags etc.
What do you think?

wbowling:
> * What do you think?

Yeah that sounds like a good solution, then it should work for both cases without the need for another flag

PR #28, opened 2019-07-01
If importing the state fails, fall back to importing the corpus

wbowling:
Fixes https://github.com/googleprojectzero/fuzzilli/issues/27

I think swapping https://github.com/googleprojectzero/fuzzilli/compare/master...wbowling:fallback-to-corpus-import?expand=1#diff-c16198c482fd89831fdde8be546741a8R248 should remove the need to reset any state on a failed import?


saelo:
Thanks!

PR #46, opened 2019-09-07
fixed typo in JS Environment builtins

amarekano:
fixes a minor typo when registering JS Environment builtins.


saelo:
Good catch, thanks!

PR #50, opened 2019-10-19
Update JSC patch to support new HostFunction signature

amarekano:
Webkit commit 622e869db55d9e78435fe55ea11007d0639ff44e updates the HostFunction signature from (ExecState*) to (JSGlobalObject*, CallFrame*). This PR updates fuzzilli's JSC patch to support this change.

The target README was also updated to list the latest webkit commit the patch was tested on.


saelo:
Thanks!

Issue #58, opened 2020-03-26
Use new JSC compilation modes


saelo:
The JSC build scripts should probably use one of these new compilation modes: https://github.com/WebKit/webkit/commit/87bdd5de3d05ff47db02dc8d32aa87495a78ba27 (either release + asserts or optimized debug)

saelo:
Likely best to leave it the way it is now, see https://github.com/googleprojectzero/fuzzilli/pull/62#issuecomment-611991197

PR #62, opened 2020-04-06
Support new JSC compilation mode

amarekano:
Added support for release+assert compilation mode. Reference: #58 
Release builds use O3 optimisation by default and so we don't need to add the optimisation flag to our build script.

I've also looked into generating debug builds with forced optimisation but that isn't any different from what we've currently got in the build script.

I haven't tested this on macOS, but I don't see the flags being too dissimilar. If someone could run a quick build on macOS and post their results here, that would be bril.

### DIFF
--- a/Targets/JavaScriptCore/fuzzbuild.sh
+++ b/Targets/JavaScriptCore/fuzzbuild.sh
@@ -17,10 +17,10 @@
 export WEBKIT_OUTPUTDIR=FuzzBuild
 
 if [ "$(uname)" == "Darwin" ]; then
-    ./Tools/Scripts/build-jsc --jsc-only --debug --cmakeargs="-DENABLE_STATIC_JSC=ON -DCMAKE_CXX_FLAGS='-fsanitize-coverage=trace-pc-guard -O3'"
+    ./Tools/Scripts/build-jsc --jsc-only --release-and-assert --cmakeargs="-DENABLE_STATIC_JSC=ON -DCMAKE_CXX_FLAGS='-fsanitize-coverage=trace-pc-guard -DASSERT_ENABLED=1'"
 elif [ "$(uname)" == "Linux" ]; then
     # Note: requires clang >= 4.0!
-    ./Tools/Scripts/build-jsc --jsc-only --debug --cmakeargs="-DENABLE_STATIC_JSC=ON -DCMAKE_C_COMPILER='/usr/bin/clang' -DCMAKE_CXX_COMPILER='/usr/bin/clang++' -DCMAKE_CXX_FLAGS='-fsanitize-coverage=trace-pc-guard -O3 -lrt'"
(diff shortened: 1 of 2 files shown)


saelo:
Thanks! So judging from the apparent need for `-DASSERT_ENABLED=1`, the cmake build doesn't actually support the Release+Assert mode properly? If that's the case then I guess we'll just leave it like that.

However, I think it would be good to have both commands there for the "testing" (Debug + Optimization) and "release+assert" configuration because iirc, Debug builds do more than just enable Asserts. In the simplest case, maybe just have both commands in fuzzbuild.sh with comments about what they do and have one of them be commented out? Maybe release+assert should be the default?

amarekano:
> the cmake build doesn't actually support the Release+Assert mode properly?

Yes, that's right, I need to add the ASSERT_ENABLED flag or some asserts (for example the fuzzilli crash tests in jsc.cpp) aren't enabled. I imagine there may be other such cases within the code base.

> I think it would be good to have both commands there for the "testing" (Debug + Optimization) and "release+assert" configuration because iirc, Debug builds do more than just enable Asserts.

I did suspect this and I was hoping someone more familiar with the codebase could point that out to me. From the looks of it, Release+assert gets you faster build times and a marginal speed boost while fuzzing. But if we lose out on code coverage for some reason (eg some asserts not being wrapped in ASSERT_ENABLED)  then I feel we are probably better off just using the Debug+Optimisation build. What are you thoughts on this?

saelo:
> I did suspect this and I was hoping someone more familiar with the codebase could point that out to me. 

I think you can find examples by grepping for "NDEBUG" in JavaScriptCore. For example, here is some code that appears to perform additional assertions in DEBUG builds: https://github.com/WebKit/webkit/blob/e668c9e0f3582fa237b3712ca11f3e59de462840/Source/JavaScriptCore/dfg/DFGCSEPhase.cpp#L125

> From the looks of it, Release+assert gets you faster build times and a marginal speed boost while fuzzing. But if we lose out on code coverage for some reason (eg some asserts not being wrapped in ASSERT_ENABLED) then I feel we are probably better off just using the Debug+Optimisation build. What are you thoughts on this?

Agree, so let's make the "testing" (debug + optimize) configuration the default and maybe make a note (maybe including the commented out command) on how to get a "release+assert" build and why we aren't using that as default?

amarekano:
Thank you, that was useful. I've diffed the exported compile commands for Debug, Release, Release+Assert and Testing and here are some observations:

Debug: -g
Release+Assert: -O3 -DNDEBUG
Testing: -O2 -g -DNDEBUG
Release: -O3 -DNDEBUG

Debug is the only one that doesn't define NDEBUG. Release, Release+Assert and Testing define NDEBUG and due to this, as you pointed out, a bunch of asserts aren't enabled. One workaround to this is to add ASSERT_ENABLED to the fuzzbuild script.

Debug builds don't have compiler optimisations set. Testing uses O2. Release and Release+Assert use O3. This again is trivial to workaround by using the --force-opt flag.

The final difference is Debug and Testing are the only builds that include debug symbols. Not a big deal really if you're only gathering crashes.

These builds are on linux and may differ on macOS. I think we should continue with the existing script for now and revisit the other compilation modes at a later date.

saelo:
Ah great, thanks for testing all this!

PR #64, opened 2020-04-22
Update patch for v8

n30m1nd:
Update the patch so that it works with the latest version of v8.

### DIFF
--- a/Targets/V8/v8.patch
+++ b/Targets/V8/v8.patch
@@ -1,8 +1,22 @@
+diff --git a/src/api/api.cc b/src/api/api.cc
+index 915a781f3a..d6bc6689fc 100644
+--- a/src/api/api.cc
++++ b/src/api/api.cc
+@@ -503,7 +503,8 @@ void Utils::ReportOOMFailure(i::Isolate* isolate, const char* location,
+     if (fatal_callback == nullptr) {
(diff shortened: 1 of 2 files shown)


Targets/V8/v8.patch
```diff
@@ -1,8 +1,22 @@
...
+index 915a781f3a..d6bc6689fc 100644
+--- a/src/api/api.cc
++++ b/src/api/api.cc
+@@ -503,7 +503,8 @@ void Utils::ReportOOMFailure(i::Isolate* isolate, const char* location,
+     if (fatal_callback == nullptr) {
+       base::OS::PrintError("\n#\n# Fatal %s OOM in %s\n#\n\n",
+                            is_heap_oom ? "javascript" : "process", location);
+-      base::OS::Abort();
++      exit(0);
++//      base::OS::Abort();
```
saelo, line 11: Ah yeah, that's a good idea!

saelo:
Thanks! Could you also update this file: https://github.com/googleprojectzero/fuzzilli/blob/master/Targets/V8/README.md to include the commit hash please?

n30m1nd:
Done - it applies to the latest build according to `git apply --stat` so it should be good to go.

saelo:
Thanks!

PR #66, opened 2020-04-29
JSC patch update

drtychai:
Patch updated to 2020-04-29 commit [dd7f161](https://github.com/WebKit/webkit/commit/dd7f161).


saelo:
Nice, thanks! Could you please also update https://github.com/googleprojectzero/fuzzilli/blob/master/Targets/JavaScriptCore/README.md to include the new base commit? You can just make another commit for that, I'll squash all commits together into a single one anyway :)

PR #67, opened 2020-05-06
Fix race condition in sanitizer coverage

khanhnt2:
I just added a small check to avoid crashes in `__sanitizer_cov_trace_pc_guard`. Although the race condition still happens but it will not lead to crashes.

### DIFF
--- a/Targets/JavaScriptCore/webkit.patch
+++ b/Targets/JavaScriptCore/webkit.patch
@@ -80,9 +80,8 @@ index 024a83ed25f..31870060c92 100644
 +}
 +
 +extern "C" void __sanitizer_cov_trace_pc_guard(uint32_t *guard) {
-+    if (!*guard) return;
 +    uint32_t index = *guard - 1;
-+    if (index == -1) return;
++    if (index == -1) return;      // Edge has already been discovered
 +    __shmem->edges[index / 8] |= 1 << (index % 8);
 +    *guard = 0;
 +}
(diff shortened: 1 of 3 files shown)


khanhnt2:
@googlebot I signed it!

saelo:
Hi, thanks for that! I assume this is for the race condition mentioned in the v8.patch? 
With the way it is now, can you not also remove the line `if (!*guard) return;` entirely so that `guard` is only dereferenced once? I.e. something like
```
extern "C" void __sanitizer_cov_trace_pc_guard(uint32_t *guard) {
    uint32_t index = *guard - 1;
    if (index == -1) return;      // Edge has already been discovered
    __shmem->edges[index / 8] |= 1 << (index % 8);
    *guard = 0;
}
``` What do you think?

**Comment by khanhnt2** [comment]
Yes, I agree.

**Comment by saelo** [comment]
Thanks 👍

**Comment by eternalsakura** [comment]
```
uint32_t index = *guard - 1;
if(index == -1)
```
error: comparison of integers of different signs: 'uint32_t' and 'int'
[-Wsign-compare]

Samuel Groß <notifications@github.com> 于2020年5月7日周四 下午5:59写道：

**Comment by saelo** [comment]
Good catch, thanks! So I think we should just get rid of the entire `-1` thing. I simplified the coverage instrumentation: https://github.com/googleprojectzero/fuzzilli/commit/4c0c0671159834c5623456a233253b3a69c01040 Let me know if there are any issues with that!

**Comment by eternalsakura** [comment]
cov.cc:62:5: error : use of declared identifier '__shmem'
```
struct shmem_data* shmem;
...
__shmem->edges[index / 8] |= 1 << (index % 8);
```
maybe you mean shmem here.

Samuel Groß <notifications@github.com> 于2020年5月11日周一 下午9:14写道：

eternalsakura:
Target in v8.patch

sakura sakura <eternalsakura13@gmail.com> 于2020年5月12日周二 上午11:24写道：

saelo:
Ah no, I wanted to rename everything to `__shmem`: https://github.com/googleprojectzero/fuzzilli/commit/4f364be8387391a666795c9198b498849ebe96d0 Thanks!

PR #69, opened 2020-05-28
Added check to detect holes in varaible map

amarekano:
This check ensures that we have a compact variable map when importing/generating programs

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -48,6 +48,7 @@ public class ProgramBuilder {
     ///
     /// The builder instance can not be used further after calling this function.
     public func finish() -> Program {
+        program.normalize()
         assert(program.check() == .valid)
         let result = program
         program = Program()
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -47,8 +47,10 @@ public class ProgramBuilder {
     /// Finalizes and returns the constructed program.
     ///
     /// The builder instance can not be used further after calling this function.
-    public func finish() -> Program {
-        assert(program.check() == .valid)
+    public func finish(checkForVariableHoles: Bool = true) -> Program {
+        if checkForVariableHoles {
+            assert(program.check() == .valid)
+        }
         let result = program
         program = Program()
         return result
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -48,7 +48,6 @@ public class ProgramBuilder {
     ///
     /// The builder instance can not be used further after calling this function.
     public func finish() -> Program {
-        program.normalize()
         assert(program.check() == .valid)
         let result = program
         program = Program()
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -47,10 +47,8 @@ public class ProgramBuilder {
     /// Finalizes and returns the constructed program.
     ///
     /// The builder instance can not be used further after calling this function.
-    public func finish(checkForVariableHoles: Bool = true) -> Program {
-        if checkForVariableHoles {
-            assert(program.check() == .valid)
-        }
+    public func finish() -> Program {
+        assert(program.check() == .valid)
         let result = program
         program = Program()
         return result
--- a/Sources/Fuzzilli/FuzzIL/Program.swift
+++ b/Sources/Fuzzilli/FuzzIL/Program.swift
@@ -185,7 +185,7 @@ public class Program: Collection, Codable {
             }
 
             // Ensure that the variable map does not contain holes
-            if(definedVariables.hasHoles()){
+            if definedVariables.hasHoles() {
                 return .invalid("variable map contains holes")
             }
             
--- a/Sources/Fuzzilli/FuzzIL/Program.swift
+++ b/Sources/Fuzzilli/FuzzIL/Program.swift
@@ -128,7 +128,7 @@ public class Program: Collection, Codable {
     }
 
     /// Checks if this program is valid.
-    public func check() -> CheckResult {
+    public func check(checkForVariableHoles: Bool = true) -> CheckResult {
         var definedVariables = VariableMap<Int>()
         var scopeCounter = 0
         var visibleScopes = [scopeCounter]
@@ -195,8 +195,10 @@ public class Program: Collection, Codable {
         }
 
         // Ensure that the variable map does not contain holes
-        guard !definedVariables.hasHoles() else {
-            return .invalid("variable map contains holes")
+        if checkForVariableHoles {
+            guard !definedVariables.hasHoles() else {
+                return .invalid("variable map contains holes")
+            }
         }
 
         return .valid
--- a/Sources/Fuzzilli/FuzzIL/Program.swift
+++ b/Sources/Fuzzilli/FuzzIL/Program.swift
@@ -134,7 +134,7 @@ public class Program: Collection, Codable {
         var visibleScopes = [scopeCounter]
         var blockHeads = [Operation]()
         var phis = VariableSet()
-        
+
         for (idx, instr) in instructions.enumerated() {
             guard idx == instr.index else {
                 return .invalid("instruction \(idx) has wrong index \(String(describing: instr.index))")
@@ -195,7 +195,7 @@ public class Program: Collection, Codable {
         }
 
         // Ensure that the variable map does not contain holes
-        if definedVariables.hasHoles() {
+        guard !definedVariables.hasHoles() else {
             return .invalid("variable map contains holes")
         }
 
--- a/Sources/Fuzzilli/FuzzIL/Program.swift
+++ b/Sources/Fuzzilli/FuzzIL/Program.swift
@@ -18,7 +18,7 @@
 ///
 /// * The empty program is valid
 /// * All input variables must have previously been defined
-/// * Variables have increasing numbers starting at zero and there are no holes (TODO check in isValid())
+/// * Variables have increasing numbers starting at zero and there are no holes
 /// * An instruction always produces a new output variable (SSA form)
 /// * The first input to a Copy operation must be a variable produced by a Phi operation (SSA form)
 ///
@@ -182,6 +182,11 @@ public class Program: Collection, Codable {
                     return .invalid("variable \(output) was already defined")
                 }
                 definedVariables[output] = visibleScopes.last!
+            }
+
+            // Ensure that the variable map does not contain holes
+            if(definedVariables.hasHoles()){
+                return .invalid("variable map contains holes")
             }
             
             // Special handling for Phi and Copy operations
--- a/Sources/Fuzzilli/FuzzIL/Program.swift
+++ b/Sources/Fuzzilli/FuzzIL/Program.swift
@@ -184,11 +184,6 @@ public class Program: Collection, Codable {
                 definedVariables[output] = visibleScopes.last!
             }
 
-            // Ensure that the variable map does not contain holes
-            if definedVariables.hasHoles() {
-                return .invalid("variable map contains holes")
-            }
-            
             // Special handling for Phi and Copy operations
             if instr.operation is Phi {
                 phis.insert(instr.output)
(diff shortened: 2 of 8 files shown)


Sources/Fuzzilli/FuzzIL/Program.swift
```diff
@@ -193,6 +193,12 @@ public class Program: Collection, Codable {
                 }
             }
         }
+
+        // Ensure that the variable map does not contain holes
+        if definedVariables.hasHoles() {
```
saelo, line 198: Nit: can you change this to `guard !definedVariables.hasHoles() else {` to be consistent with the rest of this function?

Sources/Fuzzilli/Util/VariableMap.swift
```diff
@@ -55,6 +55,10 @@ public struct VariableMap<Value>: Sequence {
     public func contains(_ variable: Variable) -> Bool {
         return elements.count > variable.number && elements[variable.number] != nil
     }
+
+    public func hasHoles() -> Bool {
+        return elements.contains(where: {$0 == nil})
```
saelo, line 60: Could this also just be `elements.contains(nil)` instead?

amarekano:
Thanks for the review and I shall add the testcases as soon as I get some time. At the moment, I'm seeing assertion failures when some programs are being minimised. The InlineReducer seems to be generating maps with holes and I'm trying to debug as to why that might be the case. I'm still in the process of familiarising myself with the minimisation phase, would you have thoughts on what some of the potential causes could be? When I get a chance, I'll post some programs (FuzzIL and lifted output) of the misbehaving ones.

saelo:
So the minimizer will delete instructions from a program to see whether they are important. This will leave the resulting program with holes in the variable numbers, and so if the Program is checked at that point, the validation would fail. One option might be to just [normalize](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/FuzzIL/Program.swift#L98) the program after every removal done by the minimizer, it should be cheap enough.

Sources/Fuzzilli/Util/VariableMap.swift
amarekano: I tried using contains(nil) but that throws a compile error: referencing instance method 'contains' on 'Optional' requires that 'Value' conform to 'Equatable' which is why I resorted to using contains(where:)

amarekano:
Thanks @saelo, I've added the new testcases as requested and I've also had to modify one of the existing ones due to the way how shrinkIfNecessary() now works. Adding the normalisation call before ProgramBuilder.finish() returns fixed the issue of holes being generated from program minimisation.
I've also added another comparison operator (unrelated to the varMap checks) but if you'd like that tracked as part of another PR, then I can revert that change.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -48,6 +48,7 @@ public class ProgramBuilder {
     ///
     /// The builder instance can not be used further after calling this function.
     public func finish() -> Program {
+        program.normalize()
```
saelo, line 51: I'm not sure this really fixes the issue. 
I think the main problem with minimization is that it relies on generating programs that are invalid in the new sense (they contain holes in variables), but otherwise correct. As such, this test: https://github.com/googleprojectzero/fuzzilli/blob/5aa6f4ef3724f197a0fc55ffc9b8caf8c809ad56/Sources/Fuzzilli/Minimization/Reducer.swift#L35 should now faill basically all the time, resulting in programs not being minimized. You might see that the average program size keeps increasing or stays very large.
I think you also can't just do `normalize` before that line in the reducer, since some programs will be invalid in other ways, thus probably causing normalize to crash (I think). 
Maybe the correct fix is to add a parameter to the `check` function: `checkForVariableHoles: bool = true` which is true by default but turned off in the Reducer with a comment like `// Due to the way the reducers work, they will produce otherwise valid programs, but with variable holes, so we don't check for those. The holes will be removed after minimization is done`.

Sources/Fuzzilli/Util/VariableMap.swift
saelo: Ah weird, well in that case ignore my comment :)

Tests/FuzzilliTests/VariableMapTest.swift
```diff
@@ -35,7 +35,8 @@ class VariableMapTests: XCTestCase {
         XCTAssert(m.contains(v(1)) && m[v(1)] == 1)
         
         m.remove(v(1))
-        XCTAssert(!m.contains(v(1)) && m[v(1)] == nil)
+
+        XCTAssert(!m.contains(v(42)) && m[v(42)] == nil)
```
saelo, line 39: This was correct the way it was before, i.e. `XCTAssert(!m.contains(v(1)) && m[v(1)] == nil)`

Tests/FuzzilliTests/VariableMapTest.swift
```diff
@@ -59,30 +60,33 @@ class VariableMapTests: XCTestCase {
         XCTAssertEqual(m1, m2)
         
         // Add another 128 elements and compare with a new map built up in the opposite order
-        for i in 128..<256 {
+        for i in 127..<255 {
```
saelo, line 63: Same, should be `128..<256`

Tests/FuzzilliTests/VariableMapTest.swift
```diff
@@ -59,30 +60,33 @@ class VariableMapTests: XCTestCase {
...
             let val = Bool.random()
             m2[v(i)] = val
         }
         
         var m3 = VariableMap<Bool>()
         XCTAssertNotEqual(m1, m3)
-        
-        for i in (0..<256).reversed() {
+
+        for i in (0..<255).reversed() {
```
saelo, line 71: Same, `256`

Tests/FuzzilliTests/VariableMapTest.swift
```diff
@@ -59,30 +60,33 @@ class VariableMapTests: XCTestCase {
...
             m3[v(i)] = m2[v(i)] ?? false
         }
+
```
saelo, line 74: Same, you can just revert all the previously existing test cases to how they were before. After all, the class should still behave the same way it did before :)

Tests/FuzzilliTests/VariableMapTest.swift
```diff
@@ -135,6 +139,58 @@ class VariableMapTests: XCTestCase {
...
+    func testForHolesAfterFirstElementRemoval() {
+        var m = VariableMap<Int>()
+
+        let mapSize = 15
+        for i in 0..<mapSize {
+            m[v(i)] = Int.random(in: 0..<20)
+        }
+        m.remove(v(0))
+
+        XCTAssertEqual(m.hasHoles(), false)
```
saelo, line 180: This should now be `true`

Tests/FuzzilliTests/VariableMapTest.swift
```diff
@@ -135,6 +139,58 @@ class VariableMapTests: XCTestCase {
...
+    func testForHolesAfterArbitraryElementRemoval() {
+        var m = VariableMap<Int>()
+
+        let mapSize = 15
+        for i in 0..<mapSize {
+            m[v(i)] = Int.random(in: 0..<20)
+        }
+        m.remove(v(Int.random(in: 0..<mapSize)))
+
+        XCTAssertEqual(m.hasHoles(), false)
```
saelo, line 192: This should now be `true`

Sources/Fuzzilli/Util/VariableMap.swift
```diff
@@ -31,11 +31,11 @@ public struct VariableMap<Value>: Sequence {
     }
     
     private mutating func shrinkIfNecessary() {
-        while elements.count > 0 && elements.last! == nil {
-            elements.removeLast()
+        while elements.count > 0 && self.hasHoles() {
+            elements = elements.filter { $0 != nil }
```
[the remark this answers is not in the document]
amarekano: Ah I see, I misunderstood the requirement for hasHoles. I assumed, incorrectly, that when a variable is removed, the var map should be compacted to remove holes.
Having re-read the testcases you listed in your first comment to this PR, what I gather is hasHoles should only return true, if there are one or more nil elements at the end of a non empty var map. In all other cases hasHoles should return false.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -463,6 +463,7 @@ public enum Comparator: String {
     case equal              = "=="
     case strictEqual        = "==="
     case notEqual           = "!="
+    case strictNotEqual     = "!=="
```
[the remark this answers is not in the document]
amarekano: No worries, I'll create another PR for this one.

Sources/Fuzzilli/FuzzIL/Program.swift
amarekano: I've fixed this in the most recent commit.

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Thanks for this suggestion, I think this would be the optimal approach and I'll add it to the PR. From what I gather we would need to call `check` with the `checkForVariableHoles` disabled in two places: one in `Reducer.swift` as you highlighted above and the other in `InliningReducer.swift`:

https://github.com/googleprojectzero/fuzzilli/blob/e26e7f78f4b364f5431547f68a412d558e56b07a/Sources/Fuzzilli/Minimization/InliningReducer.swift#L176

amarekano: So disabling the hasHoles check InliningReducer as I mentioned above, turns out, isn't sufficient. Before the assert is called, there's a call to `ProgramBuilder.finish()` which in-turn calls `Program.check()` before it returns.

https://github.com/googleprojectzero/fuzzilli/blob/e26e7f78f4b364f5431547f68a412d558e56b07a/Sources/Fuzzilli/Minimization/InliningReducer.swift#L175-L176 

https://github.com/googleprojectzero/fuzzilli/blob/e26e7f78f4b364f5431547f68a412d558e56b07a/Sources/Fuzzilli/Core/ProgramBuilder.swift#L50-L55

I'm not sure if disabling the hasHoles check in `ProgramBuilder.finish()` would be the right approach here since it's being referenced by several other components. Thoughts?

Sources/Fuzzilli/Util/VariableMap.swift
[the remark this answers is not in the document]
amarekano: Changes to shrinkIfNecessary have been reverted to the original implementation.

Tests/FuzzilliTests/VariableMapTest.swift
amarekano: Changes reverted to the original implementation

Tests/FuzzilliTests/VariableMapTest.swift
amarekano: Changes reverted to the original implementation

Tests/FuzzilliTests/VariableMapTest.swift
amarekano: Changes reverted to the original implementation

Tests/FuzzilliTests/VariableMapTest.swift
amarekano: Done, changes to the existing testcases have now been reverted to the original implementation.

Tests/FuzzilliTests/VariableMapTest.swift
amarekano: Testcase updated

Tests/FuzzilliTests/VariableMapTest.swift
amarekano: Testcase updated

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Cool, I think the InliningReducer should be fixed now: https://github.com/googleprojectzero/fuzzilli/commit/b8e1778ca24d4bfa8aae5932c6f4a04385cf9958#diff-00251930897a0e17dddc9fdd75f5acf2

With that, does it work now when only setting checkForVariableHoles=false in the ReductionVerifier?

amarekano: Brilliant, I think this would now work. Let me run a few sanity checks and update this PR. I'll try to get this done in the next hour or so :)

amarekano: I've updated the PR to have the reducers ignore holes in the variable Map. I've also reverted the changes to ProgramBuilder since they are no longer necessary with your recent updates to the InliningReducer.

saelo:
Merged! Thanks for your work!

PR #72, opened 2020-06-11
Added ecmaversion to profiles

WilliamParks:
Adds ECMAScript version as part of each profile. This is useful for engine targets outside the "big 3" already supported.

In sanity checking against V8, there is a significant performance degradation running it as es5 vs es6 (~4-5x in execs/second), which I believe is due to the additional scope overhead?

### DIFF
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Supported versions of the ECMA standard.
+public enum ECMAScriptVersion {
+    case es5
+    case es6
+}
+
+
 /// Lifts a FuzzIL program to JavaScript.
 public class JavaScriptLifter: ComponentBase, Lifter {
     /// Prefix and suffix to surround the emitted code in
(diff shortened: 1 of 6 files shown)


saelo:
Cool thanks for that! I wouldn't expect any significant performance degrading to occur from changing to es5 mode though. I gave it a quick try in jsc, and there switching to es5 mode didn't result in any noticeable performance degradation (the es5 version actually seemed to be faster, but that's probably just be by chance). That said, two different runs of the same configuration can easily be a factor of 2 or so apart for execs/second.

How are you dealing with new es6 features such as for-of loops and such? Do you think it would make sense to log an error message if these constructs are lifted in es5 mode so the user can then disable the code generators for them?

WilliamParks:
I did some more thorough testing, and it looks just to have been an accident of not running long enough to get good data the first time. Apparently, there was a performance different at one point, but it seems to have been appropriately optimized: https://stackoverflow.com/questions/36623440/let-vs-var-performance-in-nodejs-and-chrome


> How are you dealing with new es6 features such as for-of loops and such? Do you think it would make sense to log an error message if these constructs are lifted in es5 mode so the user can then disable the code generators for them?

To be honest, the target I'm working on supports some features beyond es5: [https://wiki.duktape.org/postes5features](https://wiki.duktape.org/postes5features). I only really needed access to use var/const over let, which was already implemented. My view was that getting the options to precisely target duktape would clutter things up for the main 3 targets.

saelo:
Cool, makes sense!

PR #75, opened 2020-06-14
New TerminalUI

n30m1nd:
The UI now stays on top left of the window, however it is still on an experimental state as the rest of the status messages still overlap. The following has been also added:
* Added colours to the coverage, red being low, blue being mid and green considered good coverage (arbitrarily set by just observing V8 over the course of a week of fuzzing).
* Added red colour to indicate that there's crashes to inspect.
* Added a new stat computed from coverage output: "Coverage Last Update" which indicates roughly the seconds passed since the coverage was last updated. I guess this should be decoupled from the TerminalUI file and added into where the actual stats are computed.


saelo:
Awesome! A few initial thoughts from my side:
1. Upon initialization of the TerminalUI (and maybe before every redraw?), it should clear the entire screen. Currently, it seems to clear the top of the screen but leaves old terminal output below that, which is probably confusing
2. I think the TerminalUI will have to keep a [RingBuffer](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Util/RingBuffer.swift) of log messages and print the newest messages below the stats window every time the UI is redrawn so that all recent messages can be seen. I'm not exactly sure how to best figure out how many messages to show, maybe it'll have to print the newest messages at the top instead and just continue printing log messages as long as there is still space. It might be worth checking how honggfuzz does it, since iirc it has a very similar UI
3. I think we should add a --noui flag to Fuzzilli which disables the UI and just print()s log messages and maybe logs the current statistics as info somewhat frequently. This could be a separate UI class and should probably also be a separate commit
WDYT?

n30m1nd:
Hey all of them valid points - as I said this is still very experimental as I did have a quick look on the logger and I assume I'll need to extend the class in order to do it the right way. Just quickly replying and some extra thoughts:

1. The patch does clear the screen (via ^L) once you start it but if you reattach (screen -r -d || docker attach) it doesn't clean it all, just the stats with the ^K control character that deletes the current line. I was thinking on calculating the amount of rows and doing ^K on each row so I ensure it's clean every time before printing. I already had a look on AFL but without knowledge of the codebase I'll need more time. Finally, the reason why I didn't clear on every redraw with ^L it's because it scrolls down a page every time.
2. I had the exact same thought but thinking with an array where you push and pop; great that there's already a class implementing this sort of behaviour. Again, the amount of messages I guess should be bassed on a calculation on how many rows there are (need to research on using either `tput lines` or `$LINES`).
3. Isn't this what the original UI does? I'm surely missing something here. Regardless, it would make a lot of sense since there might be environments where the "pretty" output won't be possible (I think AFL detects this, even when resizing).

I'll try to tackle the rows thing first, since I think that one will define the amount of message logs that can be on the screen at the same time. Hope that sounds good!

saelo:
Hey, yeah no worries, I'm also just dumping my thoughts here :)

1. So when I first fill the entire screen with A's, then start Fuzzilli, it basically looks like this:
```
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
Fuzzer Statistics
-----------------
Coverage:                     1.89%
Coverage Last Updated:        0
Total Samples:                20
Interesting Samples Found:    13
Valid Samples Found:          13
Corpus Size:                  14
Success Rate:                 65.00%
Timeout Rate:                 0.00%
Crashes Found:                0
Timeouts Hit:                 0
Avg. program size:            34.55
Connected workers:            0
Execs / Second:               0.00
Total Execs:                  208
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
> swift run -c debug FuzzilliCli --profile=jsc ~/webkit/FuzzBuild/Debug/bin/jsc
[Cli] No filesystem storage configured, found crashes will be discarded!
[Coverage] Initialized, 795787 edges
[JavaScriptEnvironment] initialized static JS environment model
[JavaScriptEnvironment] Have 43 available builtins: ["gc", "Infinity", "Promise", "Number", "Int16Array", "Reflect", "WeakMap", "Int32Array", "JSON", "String", "isFinite", "this", "P
roxy", "Array", "Uint32Array", "Uint8Array", "Set", "Map", "transferArrayBuffer", "noInline", "NaN", "parseFloat", "Uint16Array", "Boolean", "Math", "Float32Array", "Object", "undefi
ned", "DataView", "parseInt", "eval", "createGlobalObject", "Uint8ClampedArray", "Symbol", "Float64Array", "RegExp", "isNaN", "noFTL", "arguments", "ArrayBuffer", "Function", "Int8Ar
ray", "WeakSet"]
[JavaScriptEnvironment] Have 144 available method names: ["unshift", "atanh", "cos", "slice", "e", "setInt16", "padEnd", "getPrototypeOf", "forEach", "cbrt", "has", "charAt", "getOwn
PropertyDescriptor", "get", "abs", "getInt8", "asinh", "ownKeys", "every", "splice", "acos", "setFloat32", "clear", "defineProperties", "acosh", "setUint16", "subarray", "raw", "keyF
or", "clz32", "map", "getUint8", "flat", "isArray", "setFloat64", "log10", "min", "preventExtensions", "reverse", "join", "getOwnPropertyNames", "delete", "fround", "imul", "filter",
 "isInteger", "reduce", "freeze", "lastIndexOf", "setInt8", "flatMap", "keys", "log", "set", "includes", "sin", "fill", "getFloat64", "getUint32", "for", "replace", "b", "seal", "isS
afeInteger", "codePointAt", "pop", "expm1", "is", "sinh", "getOwnPropertyDescriptors", "c", "assign", "floor", "findIndex", "d", "setPrototypeOf", "sign", "isSealed", "charCodeAt", "
isNaN", "random", "getUint16", "max", "parse", "pow", "getOwnPropertySymbols", "shift", "construct", "deleteProperty", "getFloat32", "tan", "startsWith", "sort", "concat", "tanh", "t
rim", "create", "sqrt", "log1p", "from", "toString", "of", "isFinite", "atan2", "entries", "call", "toLocaleString", "fromCharCode", "substring", "ceil", "getInt16", "apply", "atan",
 "cosh", "reduceRight", "isFrozen", "copyWithin", "push", "values", "setUint32", "split", "fromEntries", "trunc", "a", "isExtensible", "stringify", "setUint8", "setInt32", "round", "
fromCodePoint", "hypot", "some", "defineProperty", "bind", "repeat", "padStart", "getInt32", "endsWith", "exp", "asin", "find", "indexOf", "log2", "add"]
[JavaScriptEnvironment] Have 42 property names that are available for read access: ["__proto__", "NEGATIVE_INFINITY", "toStringTag", "prototype", "name", "iterator", "split", "specie
s", "byteOffset", "isConcatSpreadable", "e", "toString", "arguments", "caller", "MAX_VALUE", "hasInstance", "description", "replace", "constructor", "EPSILON", "E", "MIN_VALUE", "b",
 "length", "unscopable", "match", "search", "toPrimitive", "MAX_SAFE_INTEGER", "PI", "NaN", "matchAll", "byteLength", "d", "asyncIterator", "c", "POSITIVE_INFINITY", "size", "buffer"
, "valueOf", "a", "MIN_SAFE_INTEGER"]
[JavaScriptEnvironment] Have 10 property names that are available for write access: ["valueOf", "toString", "a", "b", "d", "c", "e", "__proto__", "constructor", "length"]
[JavaScriptEnvironment] Have 5 custom property names: ["c", "d", "a", "b", "e"]
[Fuzzer] Initialized
[Fuzzer] Recommended timeout: at least 140ms. Current timeout: 250ms
[Fuzzer] Startup tests finished successfully
[Fuzzer] Let's go!
```
So the new UI is drawn at the top, followed by A's, followed by the log messages. But I don't know too much about terminal escape codes. Maybe honggfuzz's code is easier to follow than afl's though: https://github.com/google/honggfuzz/blob/master/display.c#L164?

2. Yeah I guess if you know how wide the current window is, you can split log messages that span more than one line and then keep the last Y lines in the buffer, where Y is the number of rows of the log window or something like that. I guess then you'd also need to figure out when the screen is resized though?

3. I think what I mean is to make a simple `LoggingUI` where all it does is print the log level and the message without color or anything like that (then there is minimal code duplication between the two UIs). Basically just this block https://github.com/googleprojectzero/fuzzilli/blob/59780e20830ee8739dfc16fca3a1d3d580e44abe/Sources/FuzzilliCli/TerminalUI.swift#L40 without the coloring. Then, if stats are of interest, a separate "StatsLogger" module could be build which would periodically log the current statistics (and then they would be printed). That might also be useful independently of the UI. But we can figure this out later :)

n30m1nd:
1. Ahhh... true... I forgot to include into the patch the following change since it's on the Fuzzer.swift file, apologies!
```swift
    public func start(runFor maxIterations: Int) {
        precondition(isInitialized)
        assert(OperationQueue.current == queue)
        
        self.maxIterations = maxIterations
        
        self.runStartupTests()
        
        logger.info("Let's go!")

        // Clear screen for good printing of stats
        print("\u{001B}[2J")
        
        // Start fuzzing
        queue.addOperation {
           self.fuzzOne()
        }
    }
```
Regardless, that won't fix the overlapping after attaching or after some time. 

2. +1 at the resizing. Also makes me think about performance impact because so far I've only seen how to do it by calling an OS command and not in a swift-native way.
3. I agree but as you said, this is not a priority I think; I'll try to tackle the above 2 points first. I'm sure that at some point with more knowledge of the code base I'll go like "Oh, that's why he was recommending the separate UI" :laughing:

n30m1nd:
I made something which definitely looks better than all the previous overlapping. Height and width are hardcoded now. Everything is now embedded on the TerminalUI - all screen cleaning and any extra stats are calculated there.

n30m1nd:
@saelo check the latest commit when you have time; I am sure this needs quite a good chunk of refactoring.
The next I have in mind is:
- To store `N` amount of status messages all the time, so that if you resize the window to a bigger one the status messages will still show.

I more and more see the reason behind having a minimal UI... :laughing:

saelo:
Hey! Thanks! So I'd *really* like to avoid having the UI execute a separate binary... :/
I just looked at [honggfuzz's UI code](https://github.com/google/honggfuzz/blob/master/display.c) a bit, and from what I can tell basically what it does is:
* Simply log all messages to stdout via printf basically (so the terminal will scroll normally and you can see older log messages by scrolling up)
* When drawing the UI, it
    * First clears the rows 0 through 13 (the UI takes 13 lines)
    * Then moves the cursor to (0, 0)
    * Then prints the UI screen
    * Then sets only lines 13 to end to be scrollable (I guess so the UI itself isn't scrolled, just the log messages)
    * Then moves the cursor to the start of the row and then moves it down by a large number (500). This will in essence put the cursor to the next empty line or at the bottom of the screen so that the next log message starts right after the last one (and possibly scrolls the screen if necessary)

One documentation I found for the escape sequences used by honggfuzz is http://www.termsys.demon.co.uk/vtansi.htm

This has a couple advantages:
* No need to buffer log messages
* Ability to see older log messages by scrolling up
* No need to keep track of the window size or window resizing

So yeah, looks like I was totally wrong when I said that we'd need to buffer log messages :)
What do you think of this approach? I'd prefer it since it seems simpler and even has a few advantages, like being able to scroll up.

Also I think we should aim to land this under a --newui flag at first (I think I changed my mind in this regard), keeping the old UI around for a bit longer. Then it's easier to test this, fix potential issues, and eventually make this the default UI.

saelo:
Closing due to inactivity. Feel free to reopen

PR #76, opened 2020-06-14
Added two duktape security bugs to the showcase

WilliamParks:
Adds two security bugs found in duktape to the showcase.

Fuzzilli also found the following, which may or many not be security related:
- [Hash handling in enum sorting](https://github.com/svaarala/duktape/pull/2322/files)
- [Assertion failure in CBOR.decode](https://github.com/svaarala/duktape/issues/2316)

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,11 @@ The following is a list of some of the bugs found with the help of Fuzzilli. Onl
 * [CVE-2019-5853](https://bugs.chromium.org/p/chromium/issues/detail?id=976627): Memory corruption in regexp length check
 * [Issue 992914](https://bugs.chromium.org/p/project-zero/issues/detail?id=1923): Map migration doesn't respect element kinds, leading to type confusion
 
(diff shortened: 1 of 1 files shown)


saelo:
Awesome, thanks for your work! 
I think in general the bar for including a bug in the bug showcase is roughly the same as for obtaining a CVE number, i.e. there should be some security impact such as memory corruption and the buggy code should have shipped in some kind of software release. For some bugs (mostly JIT related) it's not always clear whether they could lead to memory corruption, in which case I'd usually include them as well. From a quick look at the other two duktape issues you linked it doesn't look like they would lead to memory corruption, but if you think otherwise I'm happy to add them to the list as well :)

WilliamParks:
> From a quick look at the other two duktape issues you linked it doesn't look like they would lead to memory corruption.

That's what I thought as well. I don't have much JS engine exploitation experience (thus my involvement here!), so I wanted to pass them along in case I was missing something.

Issue #73, opened 2020-06-11
Adding Additional Targets


WilliamParks:
@saelo, what are your thoughts on adding additional targets to Fuzzilli?

I've got it running on [duktape](https://github.com/svaarala/duktape), and have found a handful of bugs (including some to add to the showcase shortly!).

The maintainer there is interested as well, but I didn't want to commit anyone to maintaining the additional target, as I personally cannot. Thanks!

saelo:
Great! Yeah I'll happily merge code that adds support for additional targets into Fuzzilli. That said, I likely won't be maintaining support for targets that am not fuzzing myself either. However, IMO there is definitely value in having the patches etc. in Fuzzilli even if they are outdated, since it's most likely easier to rebase an existing patchset onto a newer version than to develop a new one from scratch.

WilliamParks:
Cool! I'll get something together for another PR.

It looks like it'll be implemented with a make target on the duktape side, rather than a patch in Fuzzilli. The main thing would be making sure the profile continues to compile, if the base profile struct itself is updated.

saelo:
Ah great, that's even better. Yeah the DuktapeProfile will always be compiled into Fuzzilli, so if something in the base Profile struct changes, the build will fail and whoever changed the Profile struct will have to fix the DuktapeProfile :)

WilliamParks:
Closed by #77

PR #77, opened 2020-06-18
Adds duktape JS engine integration to fuzzilli

WilliamParks:
Adds the integration of duktape to fuzzilli. 

Fuzzing thus far has found 4 bugs, 2 of which are security related (1 additional one pending).

Long term fuzzing shows the success rate for duktape is ~50-60%, which is due to [partial integration](https://wiki.duktape.org/postes5features) of es6 and above to duktape. Since fuzzilli still produced good results and specifically targeting of es6 features would greatly complicate the code base, this seems optimal.

Since fuzzilli is being included as a make target for duktape, no specific revision based code checkout is necessary.


WilliamParks:
Should close out issue #73

Cloud/Docker/DuktapeBuilder/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# RUN git clone https://github.com/svaarala/duktape /home/builder/duktape
+RUN git clone https://github.com/svaarala/duktape.git /home/builder/duktape
+WORKDIR /home/builder/duktape
+
+# Docker will attempt to cache the output of every step. That's fine (and useful to speed things up, e.g. by avoiding
+# the need to download the entire source repository again every time!). However, whenever the following ARG is changed
+# (i.e. we are building a new version of the engine), a cache miss occurs (because the build context changed) and all
+# steps from here on are rerun. That, however, means we might be operating on an old checkout of the source code from
+# the cache, and so we update it again before checking out the requested revision.
```
saelo, line 22: I think this part is missing? Since you are on the master branch after the git clone, I think you can just do `git pull && git checkout $rev`. I think the other scripts (e.g. https://github.com/googleprojectzero/fuzzilli/blob/master/Cloud/Docker/JSCBuilder/Dockerfile#L25n) are currently not doing this correctly (they should `git pull`, not `git fetch`), but there it doesn't really matter since they'll check out a specific commit, not master. I'll go fix it anyway

Cloud/Docker/DuktapeBuilder/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+cd $(dirname $0)
+FUZZILLI_ROOT=../../..
+
+# Setup build context
+REV=$(cat $FUZZILLI_ROOT/Targets/duktape/REVISION)
```
saelo, line 9: Did you include this file? It probably should just contain `master`?

Sources/FuzzilliCli/Profiles/DuktapeProfile.swift
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 Google LLC
```
saelo, line 1: Can you make this 2020? =)

Targets/duktape/README.md
```diff
@@ -0,0 +1,8 @@
+# Target: duktape
+
+To build duktape for fuzzing:
+
+1. Download the master branch of https://github.com/svaarala/duktape.
```
saelo, line 5: Nit: please remove the '.' at the end of the two list items to match the style of the other README's

Cloud/Docker/DuktapeBuilder/Dockerfile
saelo: Fixed this now for the other Dockerfiles: https://github.com/googleprojectzero/fuzzilli/commit/5aa6f4ef3724f197a0fc55ffc9b8caf8c809ad56

WilliamParks:
All requested changes should be resolved in the commit I just pushed (I think hitting resolved is the right way of handling that?).

Cloud/Docker/DuktapeBuilder/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN git clone https://github.com/svaarala/duktape.git /home/builder/duktape
+WORKDIR /home/builder/duktape
+
+# Docker will attempt to cache the output of every step. That's fine (and useful to speed things up, e.g. by avoiding
+# the need to download the entire source repository again every time!). However, whenever the following ARG is changed
+# (i.e. we are building a new version of the engine), a cache miss occurs (because the build context changed) and all
+# steps from here on are rerun. That, however, means we might be operating on an old checkout of the source code from
+# the cache, and so we update it again before checking out the requested revision.
+RUN git pull && git checkout $rev
```
saelo, line 22: Aren't you missing the `ARG` here now? I think you can just copy lines 20 through 27 from https://github.com/googleprojectzero/fuzzilli/blob/5aa6f4ef3724f197a0fc55ffc9b8caf8c809ad56/Cloud/Docker/JSCBuilder/Dockerfile#L20 that should work

saelo: Ah hmm thinking about it I guess the caching logic as described in that comment is broken if $rev is master (as it is in the build.sh for this), since then docker won't see a change and just completely use the cached image without updating to the latest source code. Anyway, I think you can just copy+paste those lines from the JSCBuilder, then I'll merge it and then I'll think of some way to fix the caching problem since it'll soon affect v8 as well (v8 can now also be built from master for fuzzing) :)

WilliamParks: How about something like this in build.sh? It's in the commit I just pushed.
```
REV=$(git ls-remote https://github.com/svaarala/duktape.git | grep refs/heads/master | awk '{print $1;}')
```
That'll always get the most recent commit on master, without having to separately download the repo outside the dockerfile.

I'm still pretty new at Docker, but I think that gets the caching behavior we're looking for.

Also, both `HEAD` and `refs/heads/master` have the same commit hash, at least in Duktape. I think the difference would matter if the main branch is no longer named `master`, but should be fine with the repos we're working with?

WilliamParks: I had to update the dockerfile to properly consume the rev argument. Once we figure out what exactly correct looks like here, I can squash the commits before the actual merge.

saelo:
Cool, yeah that's a good solution. I squash-merged the PR now, thanks again! :)

PR #79, opened 2020-06-22
Additional target (JerryScript)

nszetei:
Hey. As the profiles now supports ES5 directly, I added a new target - [JerryScript](https://github.com/jerryscript-project/jerryscript/).

It was very successful to uncover several (e.g. related to the Proxy object) specific bugs which I reported [here](https://github.com/jerryscript-project/jerryscript/issues?q=is%3Aissue+author%3Anszetei ) and all are closed now. The patch for the target is included, currently it uses the latest version. As I wrote the patch about one month ago and I can still apply it without any modification, I believe it should work for future releases in a reasonable time frame too.


nszetei:
@googlebot I signed it!

WilliamParks:
Hi @nszetei, I'm curious as to fuzzilli's performance on Jerryscript (for reference, I'm not the primary guy on this, but am the one who did Duktape integration).

What percentage of successful test cases did you get? The ES5 support I added a couple weeks ago simply triggers using `var`/`const`, rather than `let.` However, the flag, as far as I know, still allows `for...in` loops, `weakmap`, etc., which are not in ES5. Duktape supports some, but not all of these. After running fuzzilli for a couple days, I'd have around 55-60% of my test cases be successful, which is much less than I believe v8/JSC/SpiderMonkey get.

I'd also be interested to know what the test case size parameter ends up at. After a long run, as Duktape's seemed to be much smaller (~70 bytes) than other engines.

Thanks!

nszetei:
Hey @WilliamParks, thanks for your comment, my success rate is around 71%, which is about 8-10% less than when I am fuzzing JavaScriptCore.

Also thanks for your contribution with setting `ECMAScriptVersion` though the Profiles, which finally motivated me to make this pull request.

Before that, I had to patch the Fuzzilli Lifter. I also commented out the SpreadGenerators inside `Settings.swift`, which was not possible via Profiles. JerryScript has a partial support, but some of the constructs are available only in ES2018 and are not implemented, e.g. for object literals. After commenting these Generators out the success rate was pretty much the same and I haven't investigated further how to improve it.

Here are the statistics, I was running Fuzzilli with JerryScript for 2 weeks on a limited virtual server:

```
Fuzzer Statistics
-----------------
Total Samples:                216494565
Interesting Samples Found:    698
Valid Samples Found:          156899373
Corpus Size:                  1024
Success Rate:                 72.47%
Timeout Rate:                 0.84%
Crashes Found:                9908
Timeouts Hit:                 1815259
Coverage:                     61.51%
Avg. program size:            53.43
Connected workers:            0
Execs / Second:               184.03
Total Execs:                  217276575
```

saelo:
Nice! Thanks! Feel free to also add bugs you found that qualify (basically, cause memory corruption and have been there for at least one release of the JS engine) to the [bug showcase](https://github.com/googleprojectzero/fuzzilli#bug-showcase).

@WilliamParks 55%-60% correctness rate sounds about right, I get roughly the same for the big engines as well.

I think the correct approach to deal with unsupported JS features is probably to have a list of unsupported FuzzIL operations in the profile, then have every CodeGenerator report what FuzzIL operations it generates, then automatically disable the ones that generate unsupported operations. What do you think about that? I'll probably create an issue for this, since it sounds like it would be useful to have (but not top priority since currently you can just disable the code generators as you did).

WilliamParks:
@saelo Thanks, I appreciate the additional context. Would you mind sharing what you get for average program size? Running for many core-days against Duktape gets to ~70 bytes, and the results above for Jerryscript show similar results. Even short runs against V8 got much larger.

I'm not (yet) familiar with the internals of FuzzIL, but that seems to make sense to me.

Another thing which I've thought about is someway of getting better insight as to why test cases are unsuccessful, which I ran into a bunch when trying to get the typing right for Duktape builtins. Looking through the interesting samples directory, some of the added builtins were present much more than others, when I'd have expected them to be equally represented. I'm not sure what the best way of easily representing that is, however.

saelo:
@WilliamParks If you are referring to the program size number printed [here](https://github.com/googleprojectzero/fuzzilli/blob/e26e7f78f4b364f5431547f68a412d558e56b07a/Sources/FuzzilliCli/TerminalUI.swift#L96) then that's not the size in bytes but in FuzzIL instructions. Should probably be marked as such... :) An average of 70 instructions per sample is fairly normal I'd say, I don't have current data from my fuzzing runs, but I'd guess it's very roughly around 100 (but those are often JIT related samples which probably require a bit more code than non-JIT samples). It also seems to be normal for that number to be quite large at the beginning (probably because the fuzzer triggers lots of new coverage at once), then slowly get smaller over time.

Indeed, some better debuggability would be great. I'm not sure what the best approach for that would be though. I used to just print out failing samples and their stderr/stdout output, but that's a definitely hack. You could easily write failing samples to disk, but that directory would get pretty big pretty quickly I guess... I think I will create an issue for this also :)
The fact that you are seeing some builtins more often than others might be perfectly normal though, maybe their implementation is just bigger (in terms of code edges), and so Fuzzilli finds more samples using them.

WilliamParks:
Ahh, ok, the output sizes then make sense, especially in regards to engines with/without JIT. Thanks!

The particular example I have in mind is a builtin in Duktape for doing encoding/decoding. Each method takes a string for the encoding format ("hex", "base64", etc), as well as the actual input. At the time, I hadn't added in those strings to the list of candidate strings, and the fuzzer hadn't found them on its own. The functions showed up a couple times in the interesting samples output, but only in assigning the function to something, and not an actual call of the function.

In retrospect, this is pretty easy to see, but I can imagine more complex cases that are harder to catch.

What could be useful is tracking the success rate of each builtin and mutation, and potentially in what context (such as actually calling the function vs assigning it to something). There's going to be a range of successes rates, but could help catch these issues. Thoughts?

nszetei:
> Nice! Thanks! Feel free to also add bugs you found that qualify (basically, cause memory corruption and have been there for at least one release of the JS engine) to the [bug showcase](https://github.com/googleprojectzero/fuzzilli#bug-showcase).

Thanks! And for the showcase, I'll do it shortly.

> I think the correct approach to deal with unsupported JS features is probably to have a list of unsupported FuzzIL operations in the profile, then have every CodeGenerator report what FuzzIL operations it generates, then automatically disable the ones that generate unsupported operations. What do you think about that? I'll probably create an issue for this, since it sounds like it would be useful to have (but not top priority since currently you can just disable the code generators as you did).

Thank you. I agree that it will be helpful and nice to have, although the priority is low.

PR #80, opened 2020-06-23
Added Showcase entries for JerryScript. 

nszetei:
All of the entries were reported as DoS and caused a crash in the normal builds too.


README.md
```diff
@@ -182,6 +182,13 @@ The following is a list of some of the bugs found with the help of Fuzzilli. Onl
 * [Issue 2323](https://github.com/svaarala/duktape/pull/2323): Unstable valstack pointer in putprop
 * [Issue 2320](https://github.com/svaarala/duktape/pull/2320): Memcmp pointer overflow in string builtin
 
+#### [JerryScript](https://github.com/jerryscript-project/jerryscript)
+
+- [CVE-2020-13622](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-13622): DoS via property key query for Proxy objects
```
saelo, line 187: Here and the one below, are you certain these are just DoS and cannot be exploited for e.g RCE? E.g. the stack overflow might be exploitable if it can be collided with another memory region. Also the assertion message here sounds like it could cause memory corruption in release builds (Afaict the JERRY_ASSERT is only active in debug builds and ignored in release builds)? If you are unsure, maybe you can just change the description to something that doesn't include the (worst-case) bug impact, like for example "Unexpected non-string value in ecma_get_string_from_value" and  "Stack overflow via property key for Proxy objects"?

saelo:
Cool stuff! I'm curious, what about some of the other issues you reported that also appear to trigger memory corruption (e.g. https://github.com/jerryscript-project/jerryscript/issues/3837 or maybe https://github.com/jerryscript-project/jerryscript/issues/3784)?

saelo:
Ping @nszetei :)

nszetei:
> Ping @nszetei :)

Thank you for reminding and keeping this pull request open, I wanted to dedicate more time on this to be sure and I'll finalize the list by the end of this week.

saelo:
Awesome, thanks! :)

PR #87, opened 2020-07-08
Fix for directory update in Duktape

WilliamParks:
Updates the Duktape docker container build process, due to changes in the directory structure of that project.


WilliamParks:
Just did a cleanup (removed extraneous sudo), and rebase, so this should be good to go.

PR #89, opened 2020-07-17
Minify generated JS script before executing

samo98:


Sources/Fuzzilli/Lifting/ScriptWriter.swift
```diff
@@ -14,18 +14,51 @@
...
+    private let minifyOutput: Bool
+
+    public init (minifyOutput: Bool, indent: Int = 4) {
+        self.minifyOutput = minifyOutput
+        self.indent = minifyOutput ? 0 : indent
+    }
+
+    /// If minify mode is turned on, remove whitespaces around characters
+    /// which cannot occur in JS variable name
+    func formatLine<S: StringProtocol>(_ line: S) -> String {
```
saelo, line 34: Can you avoid the additional copy in here by directly appending to self.code (including the leading whitespace in case of !minifyOutput) instead of a temporary string instance?

Sources/Fuzzilli/Lifting/ScriptWriter.swift
```diff
@@ -14,18 +14,51 @@
...
+
+    private let minifyOutput: Bool
+
+    public init (minifyOutput: Bool, indent: Int = 4) {
+        self.minifyOutput = minifyOutput
+        self.indent = minifyOutput ? 0 : indent
+    }
+
+    /// If minify mode is turned on, remove whitespaces around characters
+    /// which cannot occur in JS variable name
```
saelo, line 33: `name` => `names`

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -34,13 +34,54 @@ class LifterTests: XCTestCase {
...
+            }
+        }
+        const v7 = v0(true,1);
+        // v7 = .unknown
+
+        """
+
+        let expectedMinifiedCode = "function v0(v1,v2,v3){if(v1){const v4=v1*v2;return v4;}else{return v3;}}const v7=v0(true,1);"
+
+        let prettyCode = fuzzer.lifter.lift(program, withOptions: .dumpTypes)
```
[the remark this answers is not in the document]
samo98: makes sense, agree

Sources/Fuzzilli/Lifting/ScriptWriter.swift
```diff
@@ -14,18 +14,50 @@
...
+
+    public init (minifyOutput: Bool, indent: Int = 4) {
+        self.minifyOutput = minifyOutput
+        self.indent = minifyOutput ? 0 : indent
+    }
+
+    /// If minify mode is turned on, remove whitespaces around characters
+    /// which cannot occur in JS variable names
+    mutating func emitFormattedLine<S: StringProtocol>(_ line: S) {
+        guard self.minifyOutput else {
```
saelo, line 35: Nit: can you make this an `if` instead of a `guard`, since this doesn't just return now but actually does "useful" things

Sources/Fuzzilli/Lifting/ScriptWriter.swift
```diff
@@ -14,18 +14,50 @@
...
+            code += String(repeating: " ", count: currentIndention) + line + "\n"
+            return
+        }
+
+        func canJSNameContain(_ c: Character) -> Bool {
+            c.isLetter || c.isNumber || ["_", "$"].contains(c)
+        }
+
+        for c in line {
+            if c == " " && !code.isEmpty && !canJSNameContain(code.last!){
```
saelo, line 45: Space before `{`

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -34,13 +34,52 @@ class LifterTests: XCTestCase {
...
+        function v0(v1,v2,v3) {
+            if (v1) {
+                const v4 = v1 * v2;
+                return v4;
+            } else {
+                return v3;
+            }
+        }
+        const v7 = v0(true,1);
+
```
saelo, line 65: Is this a double newline at the end? Where does that come from? Can we get rid of that in the lifter?

Sources/Fuzzilli/Lifting/ScriptWriter.swift
samo98: I'm not huge fan of functions not returning what they do, but if we optimize additional copying then sure

Tests/FuzzilliTests/LifterTest.swift
samo98: no there is only single new line at the end, it just needs to be there explicitely (same as we do not have new line at the very beggining)

Sources/Fuzzilli/Lifting/ScriptWriter.swift
samo98: When I browsed swift basics, they used some simple logic in guard before returning 🤷‍♂️ , but sure we can follow this convention.

Tests/FuzzilliTests/LifterTest.swift
saelo: Ah I see, makes sense!

Sources/Fuzzilli/Lifting/ScriptWriter.swift
saelo: Thanks! Yeah I prefer using guards only for cases where the code directly returns or throws an exception and maybe logs an error message before (i.e. for failure cases or exceptional cases)

PR #91, opened 2020-07-17
Added Template literals and EvalGenerator

amarekano:
This PR adds a generator to generate instructions of the type:
let v1 = eval(\` ...list of instructions \`)

The [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) generator doesn't support expression placeholders yet, I shall add this in a future PR.

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -39,6 +39,12 @@ public func StringLiteralGenerator(_ b: ProgramBuilder) {
     b.loadString(b.genString())
 }
 
+public func TemplateLiteralGenerator(_ b: ProgramBuilder) {
+    if !b.isInTemplate {
+        b.loadTemplate(b.genTemplate())
+    }
+}
+
 public func BooleanLiteralGenerator(_ b: ProgramBuilder) {
     b.loadBool(Bool.random())
 }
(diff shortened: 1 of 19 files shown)


saelo:
Hey! Cool idea with the recursive code generation for eval!

Maybe you can do it a little different though. How about you introduce a new block type: "CodeString", which turns its body into a string literal that you can then feed to `eval()`? Then your code generator would look like this:

```
CodeGenerator("EvalGenerator") { b in
    let code = b.codeString() {
        b.generateRecursive()
    }
    let eval = b.loadBuiltin("eval")
    b.callFunction(eval, withArgs: [code])
}
```

Which would emit something like
```
v1 = BeginCodeString
; some instructions here
EndCodeString
v2 = LoadBuiltin "eval"
CallFunction v2, [v1]
```

The JavaScriptLifter could probably just turn `BeginCodeString` into ``vX = ` `` and `EndCodeString` into the closing backtick (i.e. just create multiline string literals out of the bodies).
That way, the body will be minimized and mutated just like all other code and will reference existing variables from the current scope without any additional effort. Also, the BlockReducer should, with a little help (remove the call to eval at the same time as the `BeginCodeString` and `EndCodeString` instructions), be able to turn eval'd code into "normal" code if the `eval` isn't necessary.

The template literals can then probably become a separate PR since they are unrelated to this.

WDYT?

amarekano:
That's brilliant, I like your approach it solves a lot of the mutation challenges that I was trying to work out. Template literals are just strings wrapped in backticks so I could have the template generator emit either regular strings or code strings. I'd like to replace `Being/EndCodeString` with `Being/EndTemplateLiteral` since I don't think there's a need for a separate template literal operation.  Thoughts?

amarekano:
Closing this PR, effort now being tracked under #97

PR #96, opened 2020-07-21
Added support for async arrow functions

amarekano:
and tidied up program.proto

### DIFF
--- a/Sources/Fuzzilli/Protobuf/operations.proto
+++ b/Sources/Fuzzilli/Protobuf/operations.proto
@@ -130,13 +130,6 @@ message BeginArrowFunctionDefinition {
 message EndArrowFunctionDefinition {
 }
 
-message BeginAsyncArrowFunctionDefinition {
-    FunctionSignature signature = 1;
-}
-
-message EndAsyncArrowFunctionDefinition {
-}
-
 message BeginGeneratorFunctionDefinition {
     FunctionSignature signature = 1;
 }
@@ -149,6 +142,13 @@ message BeginAsyncFunctionDefinition {
 }
 
 message EndAsyncFunctionDefinition {
+}
+
+message BeginAsyncArrowFunctionDefinition {
+    FunctionSignature signature = 1;
+}
+
+message EndAsyncArrowFunctionDefinition {
 }
 
 message Return {
--- a/Sources/Fuzzilli/Protobuf/operations.proto
+++ b/Sources/Fuzzilli/Protobuf/operations.proto
@@ -130,6 +130,13 @@ message BeginArrowFunctionDefinition {
 message EndArrowFunctionDefinition {
 }
 
+message BeginAsyncArrowFunctionDefinition {
+    FunctionSignature signature = 1;
+}
+
+message EndAsyncArrowFunctionDefinition {
+}
+
 message BeginGeneratorFunctionDefinition {
     FunctionSignature signature = 1;
 }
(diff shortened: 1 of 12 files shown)


Sources/Fuzzilli/Protobuf/program.proto
```diff
@@ -24,15 +24,13 @@ message Instruction {
     repeated uint32 inouts = 1;
```
saelo, line 24: Ah sorry, but can you undo the reordering here... The idea here is to have them ordered the way they are in Operations.swift and then renumber them (starting from 3 and not 5, then with increasing numbers) once we feel like we have the basic set of operations supported. This would be something to do just before one of the next releases as it will break interoperability.

amarekano: I've reverted the ordering

Sources/Fuzzilli/Protobuf/program.proto
```diff
@@ -99,6 +99,8 @@ message Instruction {
         ThrowException throwException = 63;
         Comment comment = 78;
         Nop nop = 64;
+        BeginAsyncArrowFunctionDefinition beginAsyncArrowFunctionDefinition= 79;
```
saelo, line 102: Please put these below `endAsyncFunctionDefinition`

Sources/Fuzzilli/Protobuf/operations.proto
```diff
@@ -130,6 +130,13 @@ message BeginArrowFunctionDefinition {
 message EndArrowFunctionDefinition {
 }
 
+message BeginAsyncArrowFunctionDefinition {
```
saelo, line 133: Please put these below the `EndAsyncFunctionDefinition` message

Sources/Fuzzilli/Protobuf/program.proto
amarekano: done

Sources/Fuzzilli/Protobuf/operations.proto
amarekano: done

Sources/Fuzzilli/Protobuf/sync.pb.swift
```diff
@@ -71,9 +71,9 @@ public struct Fuzzilli_Protobuf_FuzzerState {
   // `Message` and `Message+*Additions` files in the SwiftProtobuf library for
   // methods supported on all messages.
 
-  public var corpus: Data = Data()
+  public var corpus: Data = SwiftProtobuf.Internal.emptyData
```
saelo, line 74: Where does this change come from? Is this from rerunning the swift protobuf generator on the same sync.proto file?

amarekano: Ah my bad, I've just realised the `--experimental_allow_proto3_optional` flag was dropped from the protoc args. Let me re-run protoc without this flag.

saelo: Ah cool, thanks!

amarekano: So, I've re-run the compiler without the flag and it replaces `Data()` with `SwiftProtobuf.Internal.emptyData`. I'm not sure why this is still `Data()` on the master branch.

saelo: Ok, seems to be the same for me, so let's just take it :)

PR #97, opened 2020-07-22
Eval generator

amarekano:
@saelo I've created a new PR for the EvalGenerator (it was less of an effort than to fix all the merge conflicts with master) based on your comment from #91


saelo:
Thanks! I think that template literals should be separated from this though. The point of this feature is to be able to have FuzzIL/JavaScript code but inside a string literal so that it can be passed to `eval()` (while still being mutated, minimized etc.). A template literal on the other hand is something like `` `foo ${bar} baz` `` (optionally with a "tag" in front). I think those should be implemented by a single `LoadTemplateString` operation:
```
class LoadTemplateLiteral {
    let parts: [String]
}
```
which has `parts.count - 1` inputs and then lifts to e.g. ``v60 = `blabla ${v23} asdf ${v42}`;`` (in this case, `parts = ["blabla ", " asdf ", ""` and `inputs = [v23, v42]`). The template strings will have a variable number of inputs in FuzzIL and one output, while the CodeString blocks only have a single output and no inputs.

amarekano:
Fair, that makes sense and I'll update the PR to reflect that.

amarekano:
I've updated the PR and I think I've met the all requirements that we've discussed. Let me know if there's something that you'd like me to change

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -527,6 +527,16 @@ public let CodeGenerators: [CodeGenerator] = [
         let value = b.randVar()
         b.storeToScope(value, as: b.genPropertyNameForWrite())
     },
+
+    CodeGenerator("EvalGenerator") { b in
+        if !b.context.contains(.codeString) {
```
saelo, line 532: Is this because the lifter currently can't deal with nested codestrings (i.e. JS template literals)? I think you can already get nested codestrings e.g. from splice mutations (which insert snippets of code into the current program), so maybe the lifter anyway needs to be able to deal with nested codestrings?

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -82,6 +86,18 @@ struct BlockReducer: Reducer {
...
         verifier.tryNopping(candidates, in: program)
     }
+
+    private func reduceCodeString(codestring: Block, in program: Program, with verifier: ReductionVerifier){
+        var candidates = [Int]()
+        // Append the begin and end of the code string
+        candidates.append(codestring.head)
+        candidates.append(codestring.tail)
+
+        // Append the callFunction instruction which should immediately follow the EndCodeString instruction
```
saelo, line 96: It doesn't always have to follow the EndCodeString instruction, e.g. if other code was inserted in between by some mutation (also, I think in the current case it's actually the `let eval = b.loadBuiltin("eval")` instruction, so you probably need `.tail + 2`?). However, it's probably not worth the effort to try and reduce all possible CodeString + eval call patterns. But maybe you can just test if the instruction you expect to be the `eval` call is actually a call instruction with the code string as first argument and otherwise bail?

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -732,6 +732,18 @@ class Comment: Operation {
     }
 }
 
+class BeginCodeString: Operation {
```
saelo, line 735: Can you add a small comment here about what this operation does? I.e. that it turns its body into a string literal containing the lifted code, which is then particularly suited to pass to `eval()`

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Yes that's correct and you're also right that the splice mutator could insert instructions that could generate nested codestrings. Either we have the lifter detect nested codestrings and escape backticks correctly or add a check in the splice mutator to prevent splicing code string instructions. I haven't fully evaluated the impact the two approaches would have, do you have a preference on approach?

Sources/Fuzzilli/FuzzIL/Operations.swift
amarekano: done

Sources/Fuzzilli/Minimization/BlockReducer.swift
amarekano: done, I've added the checks as requested and the reducer will now default to generic block reduction if it can't find the call to `eval()`.

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Having given this some more thought, it's better to handle nested codestrings in the lifter that it would be to restrict splicing mutations. I've added/reverted some changes to support nested codestrings.

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -446,6 +449,25 @@ public class JavaScriptLifter: ComponentBase, Lifter {
                 
             case let op as Comment:
                 w.emitComment(op.content)
+
+            case is BeginCodeString:
+                if nestedCodeString > 0 {
+                    w.emit("\(constDecl) \(instr.output) = `")
+                } else {
+                    w.emit("\(constDecl) \(instr.output) = \\`")
```
saelo, line 457: This will only work for one nesting level though, not multiple ones, will it?

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -82,6 +85,24 @@ struct BlockReducer: Reducer {
...
+        candidates.append(codestring.head)
+        candidates.append(codestring.tail)
+
+        // Check if the second instruction that follows EndCodeString is a CallFunction and that the input to the call function is the output of BeginCodeString
+        // TODO: Evaluate and implement a solution that efficiently finds the eval CallFunction
+        let callInstruction = program[codestring.tail+2]
+        if callInstruction.operation is CallFunction && callInstruction.input(1) == program[codestring.head].output {
+            candidates.append(codestring.tail+2)
+            verifier.tryNopping(candidates, in: program)
+        }else {
```
saelo, line 101: Nit: space before `else`

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: Incrementing and decrementing the nesting counter when it encounters Being and End CodeStrings should handle the following cases:
```
` outer block code
  ` inner code block1
   ` inner code block 2
   `
 `
  outer block code
`
```
and cases where you have:
```
` outer code block
  ` code block 1
  `
  ` code block 2
  `
  outer code block
`
```

I'm not sure if there's a case that I've missed here

Sources/Fuzzilli/Minimization/BlockReducer.swift
amarekano: done

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: oh, i see, I've got the order wrong :(

saelo: Hm I can't seem to reproduce that, it gives me an invalid syntax exception. Are you sure escaping this way is possible?
Also I think in any case it would be good to add a test for this to https://github.com/googleprojectzero/fuzzilli/blob/master/Tests/FuzzilliTests/LifterTest.swift that builds up at least three levels of codestring nesting, lifts the resulting program, then compares the output to a constant string.

amarekano: yes, you're probably seeing syntax errors since the first back tick was being incorrectly lifted to `\\`` which was a bug.
I'll also add a test case which is a good idea.

amarekano: I've fixed the escaping bug add added testcases to verify lifting

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -71,13 +71,132 @@ class LifterTests: XCTestCase {
...
+        let program = b.finalize()
+
+        let lifted_program = fuzzer.lifter.lift(program)
+
+        let expected_program = """
+        const v0 = `
+            const v3 = 13.37 * 1337;
+            const v4 = \\`
+                const v7 = 13.37 + 1337;
+                const v8 = \\`
```
saelo, line 115: Shouldn't this now be \\\\\\\\` so that evaling v4 still leaves this as a template literal?

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -82,6 +85,24 @@ struct BlockReducer: Reducer {
         
         verifier.tryNopping(candidates, in: program)
     }
+
+    private func reduceCodeString(codestring: BlockGroup, in program: Program, with verifier: ReductionVerifier){
```
saelo, line 89: I don't think we have a decent mechanism to do this yet, so can you just add a `// TODO write a test for this reduction` here?

amarekano: done

Tests/FuzzilliTests/LifterTest.swift
amarekano: yes, good catch. Turns out escaping is harder than I thought it was :(
I've fixed this bug and updated the testcases to include nested eval statements.

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -464,6 +468,29 @@ public class JavaScriptLifter: ComponentBase, Lifter {
                 
             case let op as Comment:
                 w.emitComment(op.content)
+
+            case is BeginCodeString:
+                if nestedCodeString > 0 {
```
saelo, line 473: Do you even need this `if` now? 2**0 - 1 is zero, so I think you can even do without

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -464,6 +468,29 @@ public class JavaScriptLifter: ComponentBase, Lifter {
...
+                    let count = Int(pow(2, Double(nestedCodeString)))-1
```
saelo, line 474: Nice! Maybe make a small comment here where this series comes from, e.g. just something like "With every additional nesting level, ' becomes \' and every existing \ becomes \\, thus the power series".

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -71,13 +71,147 @@ class LifterTests: XCTestCase {
...
+        XCTAssertEqual(lifted_program, expected_program)
+
+    }
+
+    func testConsecutiveNestedCodeStrings(){
+        let fuzzer = makeMockFuzzer()
+        let b = fuzzer.makeBuilder()
+
+        let code = b.codeString() {
+            let v0 = b.loadInt(1337)
```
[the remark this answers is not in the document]
amarekano: I've renamed some of the variables. The ones that are unused, I've left them as is or the compiler starts generating warnings about unused variables.

saelo:
Great, thanks!

PR #99, opened 2020-07-27
Collect basic runtime types and send it to Fuzzilli

samo98:
* Collect JS runtime types (currently very simple)
* Output info as JSON
* Parse JS output to Type struct
* Save VariableMap<Type> in Program class
* Enable to send Program runtime types in protobuf messages

Whole flow can be enabled with `collectRuntimeTypes` flag


Sources/JS/printCollectedTypes.js
```diff
@@ -0,0 +1,77 @@
+//https://github.com/beatgammit/base64-js/blob/master/index.js
```
saelo, line 1: Is the code below copied from here? We probably can't do that due to license issues... Depending on the license, we might be able to include this code with its original license as comment, but I'm not 100% sure about that. Alternatively, maybe it's possible to avoid the need for this code by using the JSON encoding (see main comment)?

Sources/Fuzzilli/FuzzIL/Program.swift
```diff
@@ -27,6 +27,9 @@
 public final class Program: Collection {
     /// A program is simply a collection of instructions.
     private var instructions: [Instruction] = []
+
+    /// Runtype types of variables if available
+    public var runtimeTypes : [String : Type] = [:]
```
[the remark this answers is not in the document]
samo98: yeah we can, but we need to write proto file for this structure too

Sources/Fuzzilli/FuzzIL/Program.swift
```diff
@@ -43,6 +46,16 @@ public final class Program: Collection {
...
         return size
     }
+
+    /// Save type of given variable
+    public func setRuntimeType(name: String, type: Type) {
+        runtimeTypes[name] = type
+    }
+
+    /// Return type of requested variable if known
+    public func getVarType(name: String) -> Type {
```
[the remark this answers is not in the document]
samo98: sure, just out of curiosity isn't it confusing what is method and what is property? (I would guess `x.runtimeType` is already Type not method)

Sources/Fuzzilli/FuzzIL/Program.swift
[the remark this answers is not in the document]
saelo: Yeah, I think making a VariableMap type in protobuf might be hard since it's a generic class. I guess we let the Program class take care of converting between the protobuf representation (either map: int -> type or just an array of types) and the VariableMap?

Sources/JS/printCollectedTypes.js
samo98: it is under MIT license so that should not be a problem, but yeah we can change JS->Swift communication

Sources/Fuzzilli/FuzzIL/Program.swift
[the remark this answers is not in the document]
saelo: Hm yeah I guess I'm mostly trying to follow the [Swift code style](https://swift.org/documentation/api-design-guidelines/), for example, [`randomElement`](https://developer.apple.com/documentation/swift/string/2997145-randomelement) (or other stdlib collection methods) is somewhat similar, it's also a method even though you may expect it to be a property. I guess the rule of thumb for properties is that they don't require arguments and execute on O(1).

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -65,6 +66,10 @@ public class JavaScriptLifter: ComponentBase, Lifter {
         func expr(for v: Variable) -> Expression {
             return expressions[v] ?? Identifier.new(v.identifier)
         }
+
+        if options.contains(.collectTypes) {
+            w.emitBlock(initializeTypeCollection)
```
saelo, line 71: Maybe we should make it more obvious that these constants contain JavaScript code as string, maybe simply by prefixing them with `JS` or `JSCode` or so?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -493,19 +501,26 @@ public class JavaScriptLifter: ComponentBase, Lifter {
...
                 let v = instr.output
-                if policy.shouldInline(expression) && expression.canInline(instr, analyzer.usesIndices(of: v)) {
+                // Do not inline expressions when collecting runtime types, so we have information about all fuzzilli variables
+                if policy.shouldInline(expression) && expression.canInline(instr, analyzer.usesIndices(of: v)) && !options.contains(.collectTypes) {
                     expressions[v] = expression
                 } else {
                     w.emit("\(constDecl) \(v) = \(expression);")
                     if options.contains(.dumpTypes) {
                         w.emitComment("\(v) = \(interpreter.type(of: v))")
+                    } else if options.contains(.collectTypes) {
```
saelo, line 511: Maybe this should just be an `if` instead of an `else if` because it's not documented/obvious that `.dumpTypes` and `.collectTypes` are mutually exclusive?

Sources/JS/README.md
```diff
@@ -0,0 +1,10 @@
+# Javascript code
+
+We want to use these JS snippets when creating program which collects runtime types.
```
saelo, line 3: "program" -> "programs"

Sources/JS/README.md
```diff
@@ -0,0 +1,10 @@
...
+To make them avaibale for Lifter, we need to save them as strings in swift files.
```
saelo, line 4: "for Lifter" -> "to the Lifter"

Sources/JS/initializeTypeCollection.js
```diff
@@ -0,0 +1,73 @@
...
+        for (const inputType in this.inputTypes) {
+            result.inputTypes.append(inputType.serialize())
+        }
+    }
+    if(this.outputType != null)  result.outputType = this.outputType.serialize()
+
+    return result
+}
+
+var types = {}
```
saelo, line 41: Maybe make a comment here like `// Must be kept in sync with TypeSystem.swift` to explain where these constants come from?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
samo98: I would go with suffix, e.g. `initTypeCollectionScript`?

saelo: SGTM!

PR #102, opened 2020-07-29
Fixes missing xor enum in binary operators

WilliamParks:
Adds missing XOR binary operator


saelo:
Ah good catch, thanks

PR #103, opened 2020-07-31
Adding missing isView method in ArrayBuffer

hotwinter:


hotwinter:
@googlebot I signed it!

Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -339,7 +340,7 @@ public extension Type {
     static let jsRegExpConstructor = Type.jsFunction([.string] => .jsRegExp)
     
     /// Type of the JavaScript ArrayBuffer constructor builtin.
-    static let jsArrayBufferConstructor = Type.constructor([.integer] => .jsArrayBuffer)
+    static let jsArrayBufferConstructor = .functionAndConstructor([.integer] => .jsArrayBuffer) + .object(ofGroup: "ArrayBufferConstructor", withMethods: ["isView"])
```
saelo, line 343: It's only a constructor, is it not? I.e. 
```
> ArrayBuffer(42)
Uncaught TypeError: Constructor ArrayBuffer requires 'new'
    at ArrayBuffer (<anonymous>)
```

hotwinter: ah, you are right, sorry, mybad, I was mimicking the way jsArray was doing, so I think I didn't fully understand the type system at this point

saelo:
Thanks!

Issue #34, opened 2019-07-24
Implement type collection at runtime


saelo:
Currently, variable types are only computed statically with the AbstractInterpreter. This is generally pretty conservative and at times incorrect as the execution semantics of FuzzIL and JavaScript differ slightly (e.g. FuzzIL has no concept of prototypes).
As such it would be nice to be able to also collect precise variable types at runtime, e.g. for programs that are afterwards stored into the corpus.

This will require at least the following:

- A (lightweight) implementation of the type system in JavaScript
- A way to attach existing type information to a program
- A way to instrument programs so that type information is collected and afterwards send to the fuzzer process via the fuzzilli output channel

PR #106, opened 2020-08-10
Add object type collection

samo98:
Collect object properties & methods
Related to https://github.com/googleprojectzero/fuzzilli/issues/34


Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -200,11 +201,19 @@ public class JavaScriptLifter: ComponentBase, Lifter {
                 let dest = MemberExpression.new() <> input(0) <> "." <> op.propertyName
                 let expr = AssignmentExpression.new() <> dest <> " = " <> input(1)
                 w.emit(expr)
+                if options.contains(.collectTypes) {
```
saelo, line 204: How about making a helper function (at the top of `lift`) named something like `maybeUpdateTypes` that contains this entire block so it's not repeated here and below?

Sources/JS/helpers.js
```diff
@@ -0,0 +1,53 @@
...
+    constructor: 1 << 7,
+    unknown: 1 << 8,
+    bigint: 1 << 9,
+    regexp: 1 << 10,
+}
+function setGroup(value, currentType) {
+    for (var i=0;i<possibleGroups.length;i++) {
+        try {
+            if (value instanceof possibleGroups[i]) {
+                currentType.group = possibleGroups[i].name
```
saelo, line 39: Just flagging a potential gotcha: while doing `String.name = "foo"` isn't possible (`.name` is read-only), script code could do `delete String.name` in which case this code here wouldn't work correctly anymore. Since this is an edge-case, I'd be fine with just mentioning this in a comment.

Sources/JS/helpers.js
```diff
@@ -0,0 +1,53 @@
...
+                return
+            }
+        } catch(err) {}
+    }
+}
+function isValidPropName(name) {
+    return /^[a-zA-Z_$][0-9a-zA-Z_$]*$/.test(name)
```
saelo, line 46: Maybe link to the specification here to make it clear where this regex comes from?

Sources/JS/initTypeCollection.js
```diff
@@ -12,89 +12,71 @@
...
-        if (typeof value === 'boolean') {
-            setType(baseTypes.boolean)
-            return
-        }
-        if (typeof value === 'bigint') {
-            setType(baseTypes.bigint)
-            return
-        }
+        if (value == null) return getBaseType(baseTypes.undefined)
+        if (Number.isInteger(value)) return getBaseType(baseTypes.integer)
```
saelo, line 43: Maybe this needs to be wrapped in try-catch as well since script code could redefine `Number.isInteger`?

Sources/JS/initTypeCollection.js
```diff
@@ -12,89 +12,71 @@
...
-            if (value instanceof RegExp) {
-                setType(baseTypes.regexp)
-                return
-            }
+            if (value instanceof RegExp) return getBaseType(baseTypes.regexp)
         } catch(err) {}
-
+        if (typeof value === 'object' || typeof value === 'symbol') {
+            var currentType = getBaseType(baseTypes.object)
+            setGroup(value, currentType)
```
saelo, line 53: I'd probably find something like `currentType.group = determineGroup(value);` (or `currentType.setGroup(value)`) a bit more readable, but I'm fine both ways

Sources/JS/initTypeCollection.js
```diff
@@ -12,89 +12,71 @@
...
+            while (value != null) {
+                var propertyNames = Object.getOwnPropertyNames(value)
```
saelo, line 55: Maybe also needs to be in try-catch, same as above

Sources/JS/initTypeCollection.js
samo98: Then it would match all integers as floats, are we okay with that?

Sources/JS/initTypeCollection.js
samo98: Similar question as above, how do we handle this? return type object without properties? or return unknown and let abstractinterpreter decide?

Sources/JS/initTypeCollection.js
saelo: I think a fairly robust approach would be to
1. "Backup" the `Number.isInteger` function (as well as other required runtime functions) in `initTypeCollectionScript` e.g. by doing something like `var NumberIsInteger = Number.isInteger;`
2. In here, use `NumberIsInteger` in a try catch
3. If it fails, either use unknown or, in this particular case, return .float, which is also fine

Then it should only fail in the rare case that script somehow reassigns the global `NumberIsInteger` variable. 
If you wanted to make it even more robust, you could install all type collection helper variables via `Object.defineProperty(this, '$VAR_NAME', { value: $VALUE, writable: false, configurable: false, enumerable: false}` then they should basically be invisible and immutable but still be available simply via their name:

```
Object.defineProperty(this, 'foobar', {value: 42, writable: false, configurable: false, enumerable: false})
foobar
// 42
foobar = 43
foobar
// 42
delete this.foobar
foobar
// 42
```

You could also group all these global variables together under a `typeCollectionContext` object.

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -346,7 +346,7 @@ public class Fuzzer {
     func collectRuntimeTypes(_ program: Program) {
```
saelo, line 346: I think eventually it would be good to collect some statistics on how frequently type collection fails (the percentage of failed collections vs successful collections could then e.g. just be printed with one of the warning messages logged below). Doesn't have to be in this PR though.

Sources/JS/helpers.js
samo98: Well I think maybe more general mechanism can be used. Instead of array of constructor we define list of groups as pairs (`groupName`, `belongsToGroup`), where belongsToGroup is function which takes object and returns boolean if object belongs to particular group. So we don't rely on having property `name` or that group has to be checked by `instanceof`

saelo: Yup, that also SGTM!

Sources/JS/initTypeCollection.js
samo98: I guess the first one option is enough, so we don'ot overcomplicate it

Sources/Fuzzilli/Fuzzer.swift
samo98: well there should be failures only in very specific cases, so this number maybe makes sense only for time outs

Sources/JS/initTypeCollection.js
samo98: I guess in this situation runtime type does not give us much information, so wide catch for unknown is sufficient

Sources/JS/helpers.js
```diff
@@ -0,0 +1,66 @@
...
+    {name: "Uint8Array", belongsToGroup: function(obj){return obj instanceof Uint8Array}},
+    {name: "Int8Array", belongsToGroup: function(obj){return obj instanceof Int8Array}},
+    {name: "Uint16Array", belongsToGroup: function(obj){return obj instanceof Uint16Array}},
+    {name: "Int16Array", belongsToGroup: function(obj){return obj instanceof Int16Array}},
+    {name: "Uint32Array", belongsToGroup: function(obj){return obj instanceof Uint32Array}},
+    {name: "Int32Array", belongsToGroup: function(obj){return obj instanceof Int32Array}},
+    {name: "Float32Array", belongsToGroup: function(obj){return obj instanceof Float32Array}},
+    {name: "Float64Array", belongsToGroup: function(obj){return obj instanceof Float64Array}},
+    {name: "Uint8ClampedArray", belongsToGroup: function(obj){return obj instanceof String}},
+    {name: "Object", belongsToGroup: function(obj){return obj instanceof Object}},
```
saelo, line 36: Maybe make a comment here that this must be the last entry since it will also match for pretty much all other types?

PR #107, opened 2020-08-10
Adopt types

samo98:
Solves https://github.com/googleprojectzero/fuzzilli/issues/34


Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -78,6 +78,7 @@ public class JavaScriptLifter: ComponentBase, Lifter {
         }
         
         w.emitBlock(prefix)
+        w.increaseIndentionLevel()
```
saelo, line 81: IIUC, this assumes that the prefix and suffix wrap the generated script into a block (e.g. the function `main`)? I'd prefer not to assume that since it's not required (and probably shouldn't be done by the Duktape and Jerryscript profiles since they don't need it)... If we really care about nicely formatted JS, then maybe we can add a `initialScriptIndentionLevel` field to the profiles?

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -0,0 +1,95 @@
...
+        b.loadString("test")
+        let x = b.binary(b.loadInt(1), b.loadInt(2), with: .Add)
+        b.phi(x)
+        let program = b.finalize()
+        var types: [Type?] = [.string, .integer, .integer, .integer, .integer]
+        program.runtimeTypes = VariableMap<Type>(types)
+
+        // Mock mutator to mutate only 3rd instruction
+        let mutator = InputMutator(chooseIndex: {(_) in return 3})
+        let mutatedProgram = mutator.mutate(program, for: fuzzer)!
```
saelo, line 53: Since it's public, how about you directly call `InputMutator.mutate(program[3], b)` here (and manually `adopt` the other instructions)? This would remove the need to change Mutator constructors etc. which isn't that great since e.g. the `chooseIndex` parameter doesn't make sense for some of the mutators like the CombineMutator.

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        // Mock mutator to mutate 1st operation
+        let loadIntMutator = OperationMutator(chooseIndex: {(_) in return 1})
+        let mutatedIntProgram = loadIntMutator.mutate(program, for: fuzzer)!
+
+        // Only int was mutated, so we keep all types
+        XCTAssertEqual(mutatedIntProgram.runtimeTypes, VariableMap<Type>(types))
+
+        // Mock mutator to mutate 3rd operation
+        let binaryOpMutator = OperationMutator(chooseIndex: {(_) in return 3})
```
saelo, line 78: Same here

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -0,0 +1,95 @@
...
+        let x = b.phi(b.loadInt(42))
+        b.beginIf(b.loadBool(true)) {
+            b.copy(b.loadFloat(1.1), to: x)
+        }
+        b.beginElse() {
+            b.copy(b.loadString("test"), to: x)
+        }
+        b.endIf()
+        let program = b.finalize()
+        let types: [Type] = [.integer, .number, .boolean, .float]
```
saelo, line 34: Nit: isn't this missing the type of the variable produced by `loadString` (i.e. `.string` or `.unknown` since it's not executed)?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -319,19 +324,22 @@ public class ProgramBuilder {
...
+    private var varMaps = [VariableMap<Variable>]()
+    private var typeMaps = [VariableMap<Type>]()
     
     /// Prepare for adoption of variables from the given program.
     ///
     /// This sets up a mapping for variables from the given program to the
     /// currently constructed one to avoid collision of variable names.
     public func beginAdoption(from program: Program) {
-        varMaps.append([Variable: Variable]())
+        varMaps.append(VariableMap<Variable>())
```
saelo, line 335: Nit: I don't think you need `<Variable>` here since the compiler should be able to infer it.

Tests/FuzzilliTests/MutationsTest.swift
samo98: Well in runtime this variable does not exist, so I think it is correct it is not there

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
samo98: currently all of profiles wrap it in main, but yeah probably no need to complicate it

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -0,0 +1,111 @@
...
+        program.runtimeTypes = VariableMap<Type>(types)
+
+        let preparedProgram = fuzzer.engine.prepareForMutation(program)
+        XCTAssertEqual(preparedProgram.runtimeTypes, VariableMap<Type>([nil, nil] + types))
+    }
+
+    func mockBaseIntructionMutation(
+        _ program: Program, for fuzzer: Fuzzer, on index: Int, with mutate: (Instruction, ProgramBuilder) -> Void
```
saelo, line 42: You could also let the `with mutate` parameter be an instance of `BaseInstructionMutator`, then you can also call https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Mutators/BaseInstructionMutator.swift#L56 first, which some mutators expect (not the ones you are testing here though).

PR #109, opened 2020-08-11
Fix special cases in runtime types collection

samo98:
Fix cases when generated JS code messes with our runtime types collection, specifically:

Object prototype properties change:
```
const v0 = Object;
updateType(0, v0)
const v1 = v0.prototype;
updateType(1, v1)
v1.c = v1;
updateType(1, v1)
```

Generated code changes global variables:
```
for (v0 in this) {
    this[v0][42] = "foobar";
}
```

Too slow when collecting properties of large arrays:
```
const v0 = 1000000
updateType(0, v0)
const v1 = Uint8Array
updateType(1, v1)
const v2 = new v1(v0)
updateType(2, v2)
```

### DIFF
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -73,6 +73,9 @@ public class JavaScriptLifter: ComponentBase, Lifter {
         }
 
         if options.contains(.collectTypes) {
+            // Wrap type collection to its own main function to avoid using global variables
+            w.emit("function typeCollectionMain() {")
+            w.increaseIndentionLevel()
             w.emitBlock(helpersScript)
             w.emitBlock(initTypeCollectionScript)
         }
@@ -533,6 +536,9 @@ public class JavaScriptLifter: ComponentBase, Lifter {
 
         if options.contains(.collectTypes) {
             w.emitBlock(printTypesScript)
+            w.decreaseIndentionLevel()
+            w.emit("}")
+            w.emit("typeCollectionMain()")
         }
 
         return w.code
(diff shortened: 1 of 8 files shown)


Sources/JS/printTypes.js
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 for (var varNumber in types) {
```
saelo, line 14: I think here you can use `Object.keys` instead of the `for-in` plus `if`. Also, that function should probably be backed up the same way `getObjectPropertyNames` is. Finally, please make a short comment here to explain why we can't just use a for-in loop (because we must ignore properties on the prototype that script code may have set).

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -73,6 +73,8 @@ public class JavaScriptLifter: ComponentBase, Lifter {
         }
 
         if options.contains(.collectTypes) {
+            w.emit("function typeCollectionMain() {")
```
saelo, line 76: Maybe this could also use a short comment about why everything is wrapped into another function (so that type collection variables don't become global variables which would make them accessible to Fuzzilli generated code)?

Sources/JS/printTypes.js
samo98: hmm I don't quite understand how you want to iterate without that **if**, because the additional property is on object prototype -> you encounter this property when iterating any array too (e.g. result of Object.keys too)

saelo: As far as I understand, Object.keys will only return you the names of own properties and ignore properties on the prototype: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys with that it should work, shouldn't it?

samo98: yes as we discussed on call if we iterate over indeces not array itself it will work

PR #111, opened 2020-08-18
Skip checking long arrays properties & small related refactor

samo98:
* Back up `JSON.stringify` function
* move `getBaseType` to constructor
* increase timelimit
* refactor groups
* mark some groups with `slowTypeCollection` and skip their properties collection
  because of slow execution of
  ```
  const v0 = new Uint32Array(1000000) //fast
  updateType(0,v0) //slow because of Object.getOwnPropertyNames(v0)
  ```

### DIFF
--- a/Sources/JS/initTypeCollection.js
+++ b/Sources/JS/initTypeCollection.js
@@ -11,7 +11,15 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-function Type() {}
+function Type(rawValue) {
+    if(rawValue == null) return
+    this.definiteType = rawValue
+    this.possibleType = rawValue
+    if (rawValue === baseTypes.object) {
+        this.properties = []
+        this.methods = []
+    }
+}
 Type.prototype.union = function(otherType) {
     var newType = new Type()
     newType.definiteType = this.definiteType & otherType.definiteType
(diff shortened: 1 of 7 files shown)


Sources/JS/initTypeCollection.js
```diff
@@ -24,50 +32,43 @@ Type.prototype.union = function(otherType) {
...
-            if (value instanceof RegExp) return getBaseType(baseTypes.regexp)
+            if (value instanceof RegExp) return new Type(baseTypes.regexp)
         } catch(err) {}
         if (typeof value === 'object' || typeof value === 'symbol') {
-            var currentType = getBaseType(baseTypes.object)
-            currentType.setGroup(value)
+            var currentType = new Type(baseTypes.object)
+            var group = currentType.setGroup(value)
+            // handle long arrays specially to reduce time spent on properties collection
+            if (group && group.slowTypeCollection && value.length > maxArrayLength) {
```
saelo, line 65: Should this check happen before or inside the `while` loop since you could have something like `Object.prototype.__proto__ = new Uint8Array(10000)` which would not be handled by the current logic? But maybe that edge case happens so rarely that it's not worth bothering

samo98: yeah I thought about that, but it would mean we have to check group for every object in prototype chain which increases number of instructions needed to be done for every object from original code
and I did not encounter such case, so maybe just not worth it?

saelo: SGTM!

PR #113, opened 2020-08-18
Fuzzilli Compiler Support - Initial

WilliamParks:
This commit provides support in Fuzzilli for initial compiler development, such as combining individual program protobufs, and doing a corpus import.

Initial step towards #4 

There are some things I know I need to clean up (such as the arguments file), but i'd appreciate feedback on the commit as a whole first.

Also, I'm not sure why the extraneous commits are showing, but the solutions I got by googling the issue have been ineffective.

Thanks!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,6 +35,7 @@ let package = Package(
         .target(name: "REPRLRun", dependencies: ["libreprl"]),
         .target(name: "FuzzilliCli", dependencies: ["Fuzzilli"]),
         .target(name: "JS", dependencies: []),
+        .target(name: "FuzzILTool", dependencies: ["Fuzzilli"]),
 
         .testTarget(name: "FuzzilliTests", dependencies: ["Fuzzilli"]),
     ],
--- a/Sources/FuzzILTool/main.swift
+++ b/Sources/FuzzILTool/main.swift
@@ -1,0 +1,251 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import Fuzzilli
+
+let jsFileExtension = ".js"
+let protoBufFileExtension = ".il.protobuf"
+
+let corpus = Corpus(minSize: 1000, maxSize: 1000000, minMutationsPerSample: 600)
+let jsPrefix = """
+               """
+let jsSuffix = """
+               """
+let lifter = JavaScriptLifter(prefix: jsPrefix,
+                suffix: jsSuffix,
+                inliningPolicy: NeverInline(),
+                ecmaVersion: ECMAScriptVersion.es6,
+                environment: JavaScriptEnvironment(additionalBuiltins: [:], additionalObjectGroups: []))
+
+
+func importFuzzILState(data: Data) throws {
+    let state = try Fuzzilli_Protobuf_FuzzerState(serializedData: data)
+    try corpus.importState(state.corpus)
+}
+
+// Takes a path, and stores each program to an individual file in that folder
+func dumpProtobufs(dumpPath: String) throws {
+    // Check if folder exists. If not, make it
+    do {
+        try FileManager.default.createDirectory(atPath: dumpPath, withIntermediateDirectories: true)
+    } catch {
+        print("Failed to create directory for protobuf splitting. Is folder \(dumpPath) configured correctly?")
+        exit(-1)
+    }
+    // Write each program as a protobuf to individual files
+    var prog_index = 0
+    for prog in corpus {
+        let name = "prog_\(prog_index).il.protobuf"
+        let progURL = URL(fileURLWithPath: "\(dumpPath)/\(name)")
+        do {
+            let serializedProg = try prog.asProtobuf().serializedData()
+            try serializedProg.write(to: progURL)
+        } catch {
+            print("Failed to serialize program. Skipping")
+        }
+        prog_index += 1
+    }
+}
+
+// Read in a protobuf
+func readProtoToString(path: String) throws -> String {
+    let data = try Data(contentsOf: URL(fileURLWithPath: path))
+    let proto = try Fuzzilli_Protobuf_Program(serializedData: data)
+    var resString = String()
+    dump(proto, to:&resString, maxDepth: 3)
+    return resString
+}
+
+// Convert a serialized protobuf file to a FuzzIL program
+func protobufToProgram(path: String) throws -> Program {
+    let data = try Data(contentsOf: URL(fileURLWithPath: path))
+    let proto = try Fuzzilli_Protobuf_Program(serializedData: data)
+    let program = try Program(from: proto)
+    return program
+}
+
+// Take a program and lifts it to a JS script
+func liftToJS(prog: Program) -> String {
+    let res = lifter.lift(prog)
+    return res.trimmingCharacters(in: .whitespacesAndNewlines)
+}
+
+// Prints a Fuzzilli program as a string
+func getPrettyStringProgram(program: Program) -> String {
+    var resString = String()
+    dump(program, to: &resString)
+    return resString.trimmingCharacters(in: .whitespacesAndNewlines)
+}
+
+// Takes all .il.protobuf files in a directory, and lifts them to JS
+// Returns the number of files successfully converted
+func compileAllProtosToJS(dirPath: String) throws -> Int {
+    let fileEnumerator = FileManager.default.enumerator(atPath: dirPath)
+    var count = 0
+    while let fileName = fileEnumerator?.nextObject() as? String {
+        guard fileName.hasSuffix(protoBufFileExtension) else { continue }
+        let fullPath = dirPath + fileName
+        let prog = try protobufToProgram(path: fullPath)
+        let jsProgString = liftToJS(prog: prog)
+        let newFilePath = dirPath + String(fileName.dropLast(protoBufFileExtension.count)) + jsFileExtension
+        try jsProgString.write(to: URL(fileURLWithPath: newFilePath), atomically: false, encoding: String.Encoding.utf8)
+        count += 1
+    }
+    return count
+}
+
+// Provided a directory with a bunch of protobuf files, combine them all into a corpus file for consumption by Fuzzilli
+func combineProtobufs(dirPath: String, outputFile: String) throws -> Int {
+    let fileEnumerator = FileManager.default.enumerator(atPath: dirPath)
+    var failed_count = 0
+    var progs = [Program]()
+    while let fileName = fileEnumerator?.nextObject() as? String {
+        guard fileName.hasSuffix(protoBufFileExtension) else { continue }
+        let fullPath = dirPath + fileName
+        var prog = Program()
+        do {
+            prog = try protobufToProgram(path: fullPath)
+            progs.append(prog)
+        } catch {
+            print("Failed to convert to program \(fileName) with error \(error)")
+            failed_count += 1
+            continue
+        }
+    }
+
+    let buf = try encodeProtobufCorpus(progs)
+    let url = URL(fileURLWithPath: outputFile)
+    try buf.write(to: url)
+    print("Successfully converted \(progs.count) failed \(failed_count)")
+    return progs.count
+}
+
+let args = Arguments.parse(from: CommandLine.arguments)
+
+if args["-h"] != nil || args["--help"] != nil || args.numPositionalArguments != 0 {
+    print("""
+          Usage:
+          \(args.programName) [options] 
+
+          Options:
+              --fuzzILState=path          : Path of a FuzzIL state file 
+              --splitState=path           : Splits out a fuzzil profile into individual protobuf programs in specified file
+              --combineBuffs=path         : Combines all encoded protobufs in a folder into a single fuzzilli state.
+              --ILToJS=path               : Takes a single protobuf file and converts it to JS     
+              --printProtobufAsProg=path  : Takes a single protobuf file and pretty prints it as a Fuzzilli program
+              --printProtobuf=path        : Takes a single protobuf file and pretty prints it as a Protobuf
+              --dirILToJS=path            : Takes all of the .il.protobuf files in an directory, and produces .js files in that same directory  
+              --combineProtoDir=path      : combines all of the .il.protobuf files in a directory into a corpus.bin file for consumption by Fuzzilli
+          """)
+    exit(0)
+}
+
+
+let fuzzILPath = args["--fuzzILState"]
+let splitStatePath = args["--splitState"]
+let combineBuffsPath = args["--combineBuffs"]
+let ilToJSPath = args["--ILToJS"]
+let printProtoAsProgPath = args["--printProtobufAsProg"]
+let printProtoPath = args["--printProtobuf"]
+let dirILToJS = args["--dirILToJS"]
+let combineDir = args["--combineProtoDir"]
+
+if splitStatePath != nil && fuzzILPath == nil {
+    print("Splitting state requires fuzzILState to be set")
+    exit(-1)
+}
+
+// Split out an already built state
+if let splitPath = splitStatePath, let statePath = fuzzILPath {
+    do {
+        let data = try Data(contentsOf: URL(fileURLWithPath: statePath))
+        try importFuzzILState(data: data)
+        try dumpProtobufs(dumpPath: splitPath)
+    } catch {
+        print("Failed to import FuzzIL State with \(error)")
+        exit(-1)
+    }
+}
+
+// Covert a single IL protobuf file to JS and print to stdout
+else if let ilPath = ilToJSPath {
+    var prog = Program()
+    do {
+        prog = try protobufToProgram(path: ilPath)
+    } catch {
+        print("Failed to load il proto \(ilPath) with error \(error)")
+        exit(-1)
+    }
+    let jsProgString = liftToJS(prog: prog)
+    print(jsProgString)
+}
+
+// Pretty print just the protobuf, without trying to load as a program
+// This allows the debugging of produced programs that are not syntactically valid
+else if let printPath = printProtoPath {
+    let res = try readProtoToString(path: printPath)
+    print(res)
+}
+
+
+// Pretty print a protobuf as a program on stdout
+else if let printPath = printProtoAsProgPath {
+    var prog = Program()
+    do {
+        prog = try protobufToProgram(path: printPath)
+    } catch {
+        print("Failed to load il proto \(printPath) with error \(error)")
+        exit(-1)
+    }
+    let prettyProgString = getPrettyStringProgram(program: prog)
+    print(prettyProgString)
+}
+
+// Produce JS files from protobufs
+else if let dirPath = dirILToJS {
+    var isDir : ObjCBool = false
+    if !FileManager.default.fileExists(atPath: dirPath, isDirectory:&isDir) || !isDir.boolValue {
+        print("Provided directory \(dirPath) is not a valid directory path")
+        exit(-1)
+    }
+    do {
+        let numConverted = try compileAllProtosToJS(dirPath: dirPath)
+        print("Successfully converted \(numConverted) files")
+    } catch {
+        print("Failed to compile protos with error \(error)")
+        exit(-1)
+    }
+}
+
+// Combine a bunch of protobufs into a state json
+else if let dirPath = combineDir {
+    var isDir : ObjCBool = false
+    if !FileManager.default.fileExists(atPath: dirPath, isDirectory:&isDir) || !isDir.boolValue {
+        print("Provided directory \(dirPath) is not a valid directory path")
+        exit(-1)
+    }
+    do {
+        let numConverted = try combineProtobufs(dirPath: dirPath, outputFile: "corpus.bin")
+        print("Successfully combined \(numConverted) files into corpus.bin")
+    } catch {
+        print("Failed to combine protos with error \(error)")
+        exit(-1)
+    }
+}
+
+else {
+    print("Please enter a command to use")
+    exit(-1)
+}
--- a/Sources/Fuzzilli/Core/Corpus.swift
+++ b/Sources/Fuzzilli/Core/Corpus.swift
@@ -27,7 +27,7 @@ import Foundation
 ///
 /// However, once reached, the corpus will never shrink below minCorpusSize again.
 /// Further, once initialized, the corpus is guaranteed to always contain at least one program.
-public class Corpus: ComponentBase {
+public class Corpus: ComponentBase, Collection {
     /// The minimum number of samples that should be kept in the corpus.
     private let minSize: Int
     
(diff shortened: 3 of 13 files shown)


Package.swift
```diff
@@ -35,6 +35,7 @@ let package = Package(
         .target(name: "REPRLRun", dependencies: ["libreprl"]),
         .target(name: "FuzzilliCli", dependencies: ["Fuzzilli"]),
         .target(name: "JS", dependencies: []),
+        .target(name: "ILToJS", dependencies: ["Fuzzilli"]),
```
saelo, line 38: Maybe this tool should have a more generic name, e.g. "FuzzILTool" or so?

Sources/ILToJS/Arguments.swift
```diff
@@ -0,0 +1,111 @@
+// Copyright 2019 Google LLC
```
saelo, line 1: I guess we shouldn't duplicate this file. Maybe just move it into Fuzzilli/Util?

Sources/ILToJS/main.swift
```diff
@@ -0,0 +1,126 @@
+// Provides functionality in support of compiler development
```
saelo, line 1: This will need the copyright header, you can just copy it from one of the other files.

Sources/Fuzzilli/CompilerSupport/CompilerSupport.swift
```diff
@@ -0,0 +1,143 @@
+// Functions in support of js -> il compilter development
```
saelo, line 1: This will need the copyright header, you can just copy it from one of the other files.

Sources/Fuzzilli/CompilerSupport/CompilerSupport.swift
```diff
@@ -0,0 +1,143 @@
...
+
+    //Convert a serialized protobuf to a FuzzIL program
+    public func protobufToProgram(path: String) throws -> Program {
+        let data = try Data(contentsOf: URL(fileURLWithPath: path))
+        let proto = try Fuzzilli_Protobuf_Program(serializedData: data)
+        let program = try Program(from: proto)
+        return program
+    }
+
+    //Take a program and convert it to a string as a JS program
```
saelo, line 70: `// Take a FuzzIL program and lift it to a JS script`?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -172,14 +172,20 @@ public class Fuzzer {
     ///
     /// This must be called after initializing the fuzzer.
     /// Use -1 for maxIterations to run indefinitely.
-    public func start(runFor maxIterations: Int) {
+    public func start(runFor maxIterations: Int, corpusProgs initalCorpusProgs: [Program]) {
```
saelo, line 175: I think the way we should implement this is by having the user call `importCorpus` manually after initializing but before starting the fuzzer. The log messages can just go into `importCorpus`

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -247,10 +253,15 @@ public class Fuzzer {
     ///
     /// This will import each program in the given array into this fuzzer while potentially discarding
     /// some percentage of the programs if dropout is enabled.
-    public func importCorpus(_ corpus: [Program], withDropout applyDropout: Bool = false) {
+    public func importCorpus(_ corpus: [Program], withDropout applyDropout: Bool = false, logStatus: Bool = false) {
```
saelo, line 256: I don't think this function has any other users currently, so I think we can just always log progress without the need for the additional parameter

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -247,10 +253,15 @@ public class Fuzzer {
...
         dispatchPrecondition(condition: .onQueue(queue))
+        var count = 1
         for program in corpus {
-            internalImportProgram(program, withDropout: applyDropout, isCrash: false)
+            internalImportProgram(program, withDropout: applyDropout, isCrash: false, initialCorpusImport: true)
+            if logStatus && count % 500 == 0{
```
saelo, line 261: Nit: space before {

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -247,10 +253,15 @@ public class Fuzzer {
     ///
     /// This will import each program in the given array into this fuzzer while potentially discarding
     /// some percentage of the programs if dropout is enabled.
-    public func importCorpus(_ corpus: [Program], withDropout applyDropout: Bool = false) {
+    public func importCorpus(_ corpus: [Program], withDropout applyDropout: Bool = false, logStatus: Bool = false) {
         dispatchPrecondition(condition: .onQueue(queue))
+        var count = 1
         for program in corpus {
-            internalImportProgram(program, withDropout: applyDropout, isCrash: false)
+            internalImportProgram(program, withDropout: applyDropout, isCrash: false, initialCorpusImport: true)
```
saelo, line 260: Yeah, so I think we should select the import mode via another function parameter of this function - either a boolean or an enum. The two possible modes are:
1. Only programs achieving new coverage are imported, and those are also minimized. This will be fairly slow
2. All given samples are added to the corpus regardless of whether they achieve new coverage. Also, they are *not* minimized (IIUC currently, those programs that trigger new coverage are minimized, the others aren't). This mode should be very fast, but it will still need to execute every sample once to ensure that it is valid and to collect coverage and then again to collect runtime type information.

FuzzilliCli/main.swift should then probably select one of these modes depending on commandline arguments

Sources/Fuzzilli/Lifting/ScriptWriter.swift
```diff
@@ -60,7 +60,7 @@ struct ScriptWriter {
     
     /// Emit one line of code.
     mutating func emit<S: StringProtocol>(_ line: S) {
-        assert(!line.contains("\n"))
+        //assert(!line.contains("\n")) Corpus contains multiline strings
```
saelo, line 63: What exactly is the problem here?

Sources/Fuzzilli/Util/Misc.swift
```diff
@@ -37,3 +37,24 @@ func measureTime<R>(_ operation: () -> R) -> (R, Double) {
     let end = Date()
     return (r, end.timeIntervalSince(start))
 }
+
+public func databuffToProgs(_ buffer: Data) throws -> [Program]{
```
saelo, line 41: This should probably go into Fuzzilli/Protobuf and be named something like `decodeProtobufCorpus` or so?
Also I guess the encoding function should then also be moved out of Corpus.swift into here

Sources/Fuzzilli/CompilerSupport/CompilerSupport.swift
```diff
@@ -0,0 +1,143 @@
...
+
+    // func pretty print state
+    // Really just for dev help, so no need to get super fancy
+    public func getPrettyStringProgram(program: Program) -> String {
+        var resString = String()
+        dump(program, to: &resString)
+        return resString.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    // Takes all .il.protobuf files in a directory, and compiles them to JS
```
saelo, line 88: Nit: `compiled` -> `lifts`

Sources/Fuzzilli/CompilerSupport/CompilerSupport.swift
saelo: Also is this file basically only used by the ILTool? Could this code be moved into that tool then instead?

Sources/Fuzzilli/Core/Corpus.swift
```diff
@@ -36,7 +36,7 @@ public class Corpus: ComponentBase {
     private let minMutationsPerSample: Int
     
     /// The current set of interesting programs used for mutations.
-    private var programs: RingBuffer<Program>
+    public var programs: RingBuffer<Program>
```
saelo, line 39: This probably shouldn't be public... Maybe we can make Corpus conform to [Collection](https://developer.apple.com/documentation/swift/collection) so that you can just do `for program in corpus`? For that, probably Corpus can just forward to the RingBuffer methods, since the RingBuffer is a Collection.

Sources/Fuzzilli/Fuzzer.swift
saelo: I think then we also want to move `self.runStartupTests()` to the end of `initialize()`

saelo (commented):
Exciting! Did a quick first round of review :)

saelo:
Ah yeah and don't worry about the extra commits, I'll just squash+merge this into one commit at the end anyway :)

Sources/Fuzzilli/CompilerSupport/CompilerSupport.swift
WilliamParks: In moving it over, I had to make a number of previously private thing in Sources/Fuzzilli public. Making the changes seemed to make sense to me, but probably should have further review

Sources/Fuzzilli/Lifting/ScriptWriter.swift
WilliamParks: The corpus I've been using has strings in it containing newlines, which was triggering this assert, IIRC

saelo: Ah you mean string literals? But is that even valid JS syntax?

WilliamParks: To be honest, that's what I assumed was the cause, but don't have a test case to verify. I'll track one down

Sources/Fuzzilli/CompilerSupport/CompilerSupport.swift
saelo: Right, I see. I think it should be Ok to make some fields that are now "internal" (the default protection level) public. My gut feeling is that these kinds of tasks (e.g. proto conversion via `asProtobuf` and the constructors) should be possible to perform "from the outside", so then those fields should probably just be public anyway.

Sources/Fuzzilli/Fuzzer.swift
saelo: Actually I think that 2nd comment wouldn't work because `initialize()` is used in tests, which don't have access to a real JS engine. I guess we can worry about the proper placement of that call some other time :)

Sources/Fuzzilli/Fuzzer.swift
WilliamParks: Here's an idea. Have one flag w/ file input for where all successfully executed programs end up in the corpus, and a different one where only the ones that increase coverage get added. The master instance would run both, resulting in a corpus that has all prior regressions for that particular engines, and additional programs that happen to hit additional coverage beyond the provided regressions.

Sources/Fuzzilli/Fuzzer.swift
saelo: I guess that depends on whether main.swift keeps the list alive or not. I think it would happen in this block: https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/FuzzilliCli/main.swift#L223 and so when the block returns, nothing will keep the list alive and the unused programs will be freed. So it should be fine

Sources/Fuzzilli/Fuzzer.swift
WilliamParks: Makes sense. I'm making `runStartupTests` a public function, and explicitly calling it in from main, to ensure everything is ready to go prior to corpus import.

WilliamParks:
Here's another round of updates!

Sources/Fuzzilli/Fuzzer.swift
saelo: Cool, good idea!

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -15,6 +15,17 @@
 import Foundation
 
 public class Fuzzer {
+    //When importing a corpus, this determines how valid samples are added to the corpus
+    public enum CorpusImportMode {
```
saelo, line 19: Yep, this is great! Can you please also mention how the two modes handle minimization of the samples

Sources/Fuzzilli/Protobuf/ProtoUtils.swift
```diff
@@ -64,3 +66,56 @@ class OperationCache {
         operations.append(operation)
     }
 }
+
+public func decodeProtobufCorpus(_ buffer: Data) throws -> [Program]{
```
saelo, line 70: Nit: can you just change the order of `encodeProtobufCorpus` and `decodeProtobufCorpus` here since encode has the nice comment at the top explaining the format =D

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,268 @@
...
+    var count = 0
+    var bad_count = 0
+    var buf = Data()
+    while let fileName = fileEnumerator?.nextObject() as? String {
+        if !fileName.hasSuffix(protoBufFileExtension) { continue }
+        let fullPath = dirPath + fileName
+        var prog = Program()
+        do {
+            prog = try protobufToProgram(path: fullPath)
+        } catch{
```
saelo, line 127: Nit: space before `{`

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,268 @@
...
+    return resString.trimmingCharacters(in: .whitespacesAndNewlines)
+}
+
+// Takes all .il.protobuf files in a directory, and lifts them to JS
+// Returns the number of files successfully converted
+public func compileAllProtosToJS(dirPath: String) throws -> Int {
+    let fileEnumerator = FileManager.default.enumerator(atPath: dirPath)
+    var count = 0
+    while let fileName = fileEnumerator?.nextObject() as? String {
+        if !fileName.hasSuffix(protoBufFileExtension) { continue }
```
saelo, line 104: Nit: this could be `guard fileName.hasSuffix(...) else { continue }`

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,268 @@
...
+        exit(-1)
+    }
+    //Write each program as a protobuf to individual files
+    var prog_index = 0
+    for prog in corpus {
+        let name = "prog_\(prog_index).il.protobuf"
+        let progURL = URL(fileURLWithPath: "\(dumpPath)/\(name)")
+        var buf = Data()
+        do{
+            let serializedProg = try prog.asProtobuf(with: Optional.none).serializedData()
```
saelo, line 53: Can this not use this method: https://github.com/googleprojectzero/fuzzilli/blob/13be8f869b2ffdde50a1a5ae78c7eedccee254ea/Sources/Fuzzilli/FuzzIL/Program.swift#L281 ?

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,268 @@
...
+        print("Failed to create directory for protobuf splitting. Is folder \(dumpPath) configured correctly?")
+        exit(-1)
+    }
+    //Write each program as a protobuf to individual files
+    var prog_index = 0
+    for prog in corpus {
+        let name = "prog_\(prog_index).il.protobuf"
+        let progURL = URL(fileURLWithPath: "\(dumpPath)/\(name)")
+        var buf = Data()
+        do{
```
saelo, line 52: Nit: space before `{`

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,268 @@
...
+        let fullPath = dirPath + fileName
+        var prog = Program()
+        do {
+            prog = try protobufToProgram(path: fullPath)
+        } catch{
+            print("Failed to convert to program \(fileName) with error \(error)")
+            bad_count += 1
+            continue
+        }
+        let proto = prog.asProtobuf(with: Optional.none) //Don't use opcache, to keep things simple
```
saelo, line 132: Same as above, but couldn't you also use the `encodeProtobufCorpus` function to avoid code duplication?

Sources/Fuzzilli/Protobuf/ProtoUtils.swift
```diff
@@ -64,3 +66,56 @@ class OperationCache {
...
+        let data = buffer.subdata(in: offset..<offset+size)
+        offset += size + align(size, to: 4)
+        let proto = try Fuzzilli_Protobuf_Program(serializedData: data)
+        let program = try Program(from: proto, with: opCache)
+        newPrograms.append(program)
+    }
+    return newPrograms
+}
+
+public func encodeProtobufCorpus<T: Collection>(_ programs: T, _ logger: Logger!) -> Data where T.Element == Program {
```
saelo, line 91: This function requiring a `Logger` instance probably prevents its reuse in some cases. I think it'd be fine to either just drop the logging entirely or alternatively to have this function also return how many programs it successfully serialized, then the caller can still log a message based on that.

Sources/FuzzilliCli/main.swift
```diff
@@ -262,6 +278,30 @@ fuzzer.sync {
         }
     }
 
+    // Alternatively, import a corpus file of unrun test cases
+    else{
```
saelo, line 282: Nit: space before `{`

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -69,15 +69,21 @@ public class JavaScriptLifter: ComponentBase, Lifter {
         }
     }
 
-    private func lift(_ program: Program, withOptions options: LiftingOptions, withPolicy policy: InliningPolicy) -> String {
+    //Support function allowing compiler development to lift arbitrary protobufs
+    public func ilSupportLift(_ program: Program) -> String {
```
saelo, line 73: Is the issue here that the JavaScriptLifter normally requires a `Fuzzer` instance to operate? And in this case the `fuzzer` and `logger` members will be `nil` now? That's probably not great since it could lead to crashes when one of those is used in other places in the future...
I think we should fix that properly. For now it should be enough to pass in the Environment to the constructor and turn the `logger.fatal` into a `fatalError`. WDYT?

WilliamParks: Yes, that's the issue I ran into. My goal is minimizing the amount of setup and options that FuzzIlTool has to set up, which don't affect how it operates. I think the change you suggested made sense, and have pushed that up!

Sources/Fuzzilli/Util/Misc.swift
```diff
@@ -26,7 +26,7 @@ func uniqueElements<E>(of list: [E]) -> [E] where E: Hashable {
     return Array(Set(list))
 }
 
- func align(_ v: Int, to desiredAlignment: Int) -> Int {
+public func align(_ v: Int, to desiredAlignment: Int) -> Int {
```
saelo, line 29: This no longer needs to be public, does it?

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,257 @@
...
+    }
+    //Write each program as a protobuf to individual files
+    var prog_index = 0
+    for prog in corpus {
+        let name = "prog_\(prog_index).il.protobuf"
+        let progURL = URL(fileURLWithPath: "\(dumpPath)/\(name)")
+        var buf = Data()
+        do {
+            let serializedProg = try prog.asProtobuf().serializedData()
+            buf.append(serializedProg)
```
saelo, line 54: Isn't `serializedProg` already a Data instance? Or why is this `append` to an empty Data needed here?

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,257 @@
...
+public func protobufToProgram(path: String) throws -> Program {
+    let data = try Data(contentsOf: URL(fileURLWithPath: path))
+    let proto = try Fuzzilli_Protobuf_Program(serializedData: data)
+    let program = try Program(from: proto)
+    return program
+}
+
+//Take a program and lifts it to a JS script
+public func progToJSString(prog: Program) -> String{
+    let lifter = JavaScriptLifter(prefix: jsPrefix,
```
saelo, line 82: Maybe this lifter should be a file-global instance like the corpus?

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,257 @@
...
+
+let jsFileExtension = ".js"
+let protoBufFileExtension = ".il.protobuf"
+
+
+private var corpus: Corpus
+private let jsPrefix: String
+private let jsSuffix: String
+
+corpus = Corpus(minSize: 1000, maxSize: 100000, minMutationsPerSample: 600)
```
saelo, line 26: I think you can do the initialization of these 3 in line with the declaration?

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,257 @@
...
+//Convert a serialized protobuf to a FuzzIL program
+public func protobufToProgram(path: String) throws -> Program {
+    let data = try Data(contentsOf: URL(fileURLWithPath: path))
+    let proto = try Fuzzilli_Protobuf_Program(serializedData: data)
+    let program = try Program(from: proto)
+    return program
+}
+
+//Take a program and lifts it to a JS script
+public func progToJSString(prog: Program) -> String{
```
saelo, line 81: Nit: space before `{`

saelo, line 81: Maybe this function should just be called `liftToJS` or so?

Sources/FuzzILTool/main.swift
```diff
@@ -0,0 +1,257 @@
...
+// limitations under the License.
+
+import Foundation
+import Fuzzilli
+
+let jsFileExtension = ".js"
+let protoBufFileExtension = ".il.protobuf"
+
+
+private var corpus: Corpus
```
saelo, line 22: I don't think these need to be `private`, or else other stuff in this file should probably also be...

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -282,8 +301,16 @@ public class Fuzzer {
             didCrash = true
 
         case .succeeded:
-            if let aspects = evaluator.evaluate(execution) {
+
+            if let aspects = evaluator.evaluate(execution){
```
saelo, line 305: With this logic, when `includeAll` is selected, doesn't it now still minimize programs that find new coverage? That's probably not what we want though, is it?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
saelo: Cool! With that, this function is no longer needed, is it?

Sources/Fuzzilli/Protobuf/ProtoUtils.swift
```diff
@@ -64,3 +66,57 @@ class OperationCache {
         operations.append(operation)
     }
 }
+
+public func encodeProtobufCorpus<T: Collection>(_ programs: T) -> (Int, Data) where T.Element == Program {
```
saelo, line 70: Actually, thinking more about this, I think this function should just throw when an encoding error occurs, like `decodeProtobufCorpus` does. Then the whole `do {} catch` block isn't necessary and the return type again becomes just `Data`. WDYT?

Sources/FuzzilliCli/TerminalUI.swift
```diff
@@ -51,7 +51,7 @@ class TerminalUI {
         fuzzer.registerEventListener(for: fuzzer.events.CrashFound) { crash in
             if crash.isUnique {
                 print("########## Unique Crash Found ##########")
-                print(fuzzer.lifter.lift(crash.program))
+                print(fuzzer.lifter.lift(crash.program, withOptions: []))
```
saelo, line 54: Why is the `withOptions:` added here? It should just call https://github.com/googleprojectzero/fuzzilli/blob/0e4935efd3ba2d860c2d1647be25f5a333909719/Sources/Fuzzilli/Lifting/Lifter.swift#L21 no?

Sources/Fuzzilli/Protobuf/ProtoUtils.swift
WilliamParks: Looking at my current error log for running the compiler against a large JS file corpus, all current errors are caught here: https://github.com/googleprojectzero/fuzzilli/pull/113/files#diff-9492c4d06ed1e3e95cd0b62407fba428R129, and I don't see expect many errors popping up going from a list of Fuzzilli Programs to the final corpus file. So, we should be good here!

Sources/Fuzzilli/Fuzzer.swift
WilliamParks: Would there be value in adding both the minimized and full versions to the corpus? The minimized version that expands coverage would be valuable as input to the combine & concat mutators, while the full, unminimized one is a good starting point for variants. What do you think?

WilliamParks: Thinking about this a bit more, I think keeping only the full version makes sense here, to keep the behavior clear. My current setup is having an engine-unique corpus made of regressions for that engine, and a shared corpus of all regression tests, which are used as input to importCorpusAll and importCorpusNewCov, respectively. Including a sample program in both corpora ensures the behavior I mentioned in my previous comment.

saelo: Yeah, that sounds good!

WilliamParks:
@saelo The latest update should resolve the outstanding issues.

In terms of functionality, the only change is importState and the two importCorpus arguments can now be used together, with corpus processing happening after state import. I don't see reason why they should be exclusive, and it might be useful at some point.

I really appreciate the feedback and help with this!

saelo:
Great, Thanks! I think there are some inconsistencies now in the internal program importing logic, but I think it'll be easiest if I just do a followup commit to clean that up, since I'll need to play around with it a bit myself to figure out the best way to do it. Merging this now, thanks for your work! =)

Issue #101, opened 2020-07-28
Unary Operators in Fuzzilli differ from the JS definition


WilliamParks:
In working on #4, I ran into what appears to be a discrepancy in how Fuzzilli definies a unary operator (++, --), implemented [here](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/FuzzIL/Operations.swift#L456) vs what I understand the JS definition to be (+, 1), described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_negation).

saelo:
Hi! Can you elaborate a bit more on what exactly the issue is that you are running into?
Is it that the `++` and `--` operators in FuzzIL are actually translated to `x + 1` and `x - 1` respectively or that the negation operator is missing (this should probably just be added...)?
For the `++` and `--` operators, it should be possible to translate something like 
```
let i = 42;
i++;
```

into 
```
v0 = LoadInt 42
v1 = Phi v0
v2 = ++ v1
Copy v1 v2
```

But I understand that that's probably not very convenient/intuitive. I'm open for discussions about how to implement these operators "properly" in FuzzIL. The main problem is probably that they could only operate on Phis as they modify their input?

WilliamParks:
I think the best resolution to this issue would be implementing "standard" unary +/-.

Mainly, I was ran into this when the parser I'm using reported `+v0` and `-v0`, as unary plus/minus, which fuzzilli then turned into `v0 + 1` and `v0 - 1` respectively!

saelo:
Ah yeah, I agree, that should be fixable by adding `+` and `-` as unary operators. Feel free to send a PR for that :)

WilliamParks:
Closed by #114

PR #114, opened 2020-08-19
Unary ops

WilliamParks:
Adds +/- unary ops, closing #101 

Like #113, not sure why the merge commits are showing up in the log


saelo:
Are you using the swift-protobuf plugin built from source or installed via homebrew? We should probably use the latest stable release (1.11.0) for consistency, but that doesn't appear to have the "inline closures" stuff yet, so probably the next person to change the protobufs would revert these changes again...

WilliamParks:
@saelo That should fix it!

saelo:
Thanks!

PR #122, opened 2020-08-24
Removed Phis

saelo:
Now, every variable can be reassigned by a Reassign operation, without having to be explicitly marked as a "reassignable" variable. Also, the new Dup instruction can now be used to duplicate a variable.


amarekano:
Many thanks for your effort on this @saelo :)
Quick question on the use of ```let``` and ```var``` is there a reason the fuzzer is restricted to using just one or the other? I was wondering if allowing the use of both ```var``` and ```let``` in generated programs would help with triggering scoping bugs (e.g. when used with BegnBlockStatement or BeginWith statements).
Perhaps there's also the case that allowing both may end generating a higher proportion of invalid samples and is probably not worth the effort?

saelo:
I think the main reason is that it'd be unclear when the Lifter should use `var` and when it should use `let`. Lifting has to be deterministic (otherwise, a crashing sample might not reproduce simply due to lifting indeterminism), so there would have to be clear rules for when to use which. But yeah, I agree that it'd probably be worth experimenting with using `var` some time to see if that triggers any interesting behaviour.

amarekano:
On the subject of inc/dec operators, I came across this interesting JS test case and I couldn't really convert to FuzzIL without losing semantic precision. The test case is as follows: 
```  
var index = 0;
function sideEffect(){
 return index++;
}

sideEffect()
```
The issue here is that ```return``` in FuzzIL only supports a variable as an input and not a UnaryOperation. To convert and lift this testcase to FuzzIL the function would probably need to be transformed to something like:
```
var index = 0;
function sideEffect(){
 let v1 = index;
 let v2 = index++;
 return v1;
}

sideEffect()
```
However, in doing so, i think, we miss out on potentially interesting optimisation bugs around return statements with unary operators? You're more familiar with JIT optimisations than I am so I wanted to get your thoughts on this.

saelo:
One of the ideas behind FuzzIL's design is that the JIT compiler (and probably also the interpreter) generally does not care about syntactical properties of the code, e.g. whether you write something as
```
return index++;
```
Or as 
```
const x = index++;
return x;
```
Because it will translate the code to it's own (usually SSA-based) IR anyway, and that will in both cases basically look like the 2nd option (i.e. a dataflow graph). That's why instructions in FuzzIL only accept variables as inputs and never other, intermediate expressions/instructions.

That said, our Lifter can be smart enough to inline expression and thus get you `return index++` if you want it to.
Does that answer your question?

PR #123, opened 2020-08-25
Add iterable type

### LINKED ISSUE #108: CodeGenerators that spread should be more careful
There are currently three different FuzzIL operations (and corresponding CodeGenerators) that do spreading:
- CreateObjectWithSpread (`v1 = {...v0};`)
- CreateArrayWithSpread (`v1 = [...v0];`)
- CallFunctionWithSpread (`v2 = v1(...v0);`)

in the first case, the spreading will enumerate the properties of the spread value, which is basically possible on every type. However, in the second and third case, it will perform an iteration (and therefore look up the `Symbol.iterator` property), which is only valid on things like Arrays and Strings. As such, the CodeGenerators for them should ensure that spreading is safely possible, for example through a helper function `isIterable` that statically determines whether a variable holds an iterable object.

samo98:
Closes #108

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -86,7 +86,9 @@ public let CodeGenerators: [CodeGenerator] = [
         }
         
         // Pick some random inputs to spread.
-        let spreads = initialValues.map({ _ in Bool.random() })
+        let spreads = initialValues.map({ el in
+            probability(0.75) && b.type(of: el).Is(.iterable)
+        })
         
         b.createArray(with: initialValues, spreading: spreads)
     },
@@ -257,7 +259,9 @@ public let CodeGenerators: [CodeGenerator] = [
         guard let arguments = b.generateCallArguments(for: FunctionSignature.forUnknownFunction) else { return }
         
         // Pick some random arguments to spread.
-        let spreads = arguments.map({ _ in Bool.random() })
+        let spreads = arguments.map({ arg in
+            probability(0.75) && b.type(of: arg).Is(.iterable)
+        })
         
         b.callFunction(f, withArgs: arguments, spreading: spreads)
     },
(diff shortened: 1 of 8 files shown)


saelo (commented):
Great! I'm wondering if the .iterable type should be a "real" type though and not a flag type. Looking at the existing code, the flag types are more or less "pseudo-types" that don't occur in actual variables but only in e.g. function signatures. Thus, I think you'd need to think about how to deal with e.g. [this](https://github.com/googleprojectzero/fuzzilli/blob/8a7f4f36476e9ccd8f32589d203756c863acb0a5/Sources/Fuzzilli/Core/ProgramBuilder.swift#L289) code which, IIUC, would now "strip" the .iterable from some types accidentally. I think if you made .iterable a real type (i.e. 1 << 11, and remove from flagTypes array), then:
- Generators that need to iterate can just say `input: .iterable` which seems intuitive (they shouldn't care about it being a .object also)
- The Environment would contain code like `jsString = .string + .object(...) + .iterable`, which also seems quite intuitive to me
- you would test for iterable with `b.type(of: X).Is(.iterable)` which should be as good as the current code
- You are not subject to [these](https://github.com/googleprojectzero/fuzzilli/blob/8a7f4f36476e9ccd8f32589d203756c863acb0a5/Sources/Fuzzilli/FuzzIL/TypeSystem.swift#L515) arguably somewhat arbitrary restrictions

samo98:
My idea was to avoid having some made up 'real' type, so we avoid some weird behavior in future and also somehow forbid having type .iterable alone, which is missing the real type.
But as I look at it again we don't have to have 1:1 type matching to JS and .iterable alone is probably ok too
And I guess you are right, currently `flagTypes` have very specific use case.
I'll make it 'real' type.

saelo:
Yeah I guess part of the question here is also whether we could consider something to just be a `.iterable` and nothing else (although we won't have anything in our environment that's just .iterable), and I'd be comfortable with a "Yes" here, it would just express that what we have is something that you can iterate over but doesn't necessarily contain properties/methods or behave like any of the primitive types. It would basically be the "pure" array type similar to how .object (probably without a group) is the pure object type.

saelo (commented):
Looks good! One last request, can you add the new type to https://github.com/googleprojectzero/fuzzilli/blob/master/Tests/FuzzilliTests/TypeSystemTest.swift#L827 and https://github.com/googleprojectzero/fuzzilli/blob/master/Tests/FuzzilliTests/TypeSystemTest.swift#L749 please (and, while you are at it, please also add `.bigint` and `.regexp` as I guess those have been forgotten)

PR #124, opened 2020-08-26
Fix DeleteComputedProperty

samo98:
Now
```
let v0 = b.createObject(with: [:])
let v1 = b.loadString("a")
b.deleteComputedProperty(v1, of: v0)
```
lifts to `delete "a"[v0]` but it should be `delete v0["a"]`


saelo:
Ohh good catch!

PR #125, opened 2020-08-28
Deduplicate type extensions

samo98:
In order to save long-term memory we should share TypeExtension instances in corpus.

We implement it as follows: 
* Corpus owns weakSet of currently used TypeExtension instances
* Anytime program is added to corpus it either adds its extension to weakSet (if its typeExtension is not used anywhere at the moment) or change its type to use shared typeExtension from weakSet (if some other program already uses same typeExtension)
* When program is removed, some typeExtensions in weakSet can become `nil`, so we have to remove them from weakSet on `cleanup`


Tests/FuzzilliTests/WeakSetTest.swift
```diff
@@ -0,0 +1,66 @@
...
+
+import XCTest
+@testable import Fuzzilli
+
+class WeakSetTests: XCTestCase {
+
+    func testWeakSetInsert() {
+        var weakSet = WeakSet<TypeExtension>()
+
+        func getTestTypeExtension() -> TypeExtension {
```
saelo, line 23: Nit: maybe name it `makeTestTypeExtension`

Tests/FuzzilliTests/WeakSetTest.swift
```diff
@@ -0,0 +1,66 @@
...
+
+        func addLocalVariable() {
+            let localTypeExtension = getTestTypeExtension()
+            let insertionResult = weakSet.insert(localTypeExtension)
+
+            XCTAssert(weakSet.contains(getTestTypeExtension()))
+            XCTAssert(insertionResult.inserted)
+            XCTAssert(insertionResult.memberAfterInsert === localTypeExtension)
+        }
+        addLocalVariable()
```
saelo, line 35: Could you also test this behaviour without the additional function, by just setting the equivalent of the `localTypeExtension` variable to `nil` or reassigning it, then checking that the set is empty?
Mostly asking because I found the functions name `addLocalVariable` somewhat unintuitive but also don't have a better suggestion for the name =)

Sources/Fuzzilli/Core/Corpus.swift
```diff
@@ -38,6 +38,8 @@ public class Corpus: ComponentBase {
     /// The current set of interesting programs used for mutations.
     private var programs: RingBuffer<Program>
     private var ages: RingBuffer<Int>
+
+    private var typeExtensionDeduplicationSet = WeakSet<TypeExtension>()
```
saelo, line 42: Maybe add a comment here that the Corpus deduplicates the runtime types of its programs to conserve memory.

Sources/Fuzzilli/Core/Corpus.swift
```diff
@@ -84,13 +86,24 @@ public class Corpus: ComponentBase {
...
     /// Adds multiple programs to the corpus.
     public func add(_ programs: [Program]) {
         programs.forEach(add)
     }
+
+    /// Change type extensions for cached ones to save memory
+    private func cacheTypeExtensions(program: Program) {
+        var cachedRuntimeTypes = VariableMap<Type>()
+        for (variable, runtimeType) in program.runtimeTypes {
+            let cachedType = runtimeType.getCachedType(cache: &typeExtensionDeduplicationSet)
```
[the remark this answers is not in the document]
saelo: Mostly just a semantic difference, but I think `uniquify` is the better term for what this achieves, since the goal is not to improve the speed of future lookups or anything like that (as a cache would) but to reduce memory usage by uniquifying instances

saelo (commented):
Really nice! Thanks!

Sources/Fuzzilli/Core/Corpus.swift
```diff
@@ -171,6 +184,9 @@ public class Corpus: ComponentBase {
             }
         }
 
+        // Clean up no longer existing elements
+        defer { typeExtensionDeduplicationSet.removeNils() }
```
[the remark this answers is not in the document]
samo98: I guess it is not necessary cause this is repetitive task.
But I thought that if we do it *after* this function ends there can be some discarded `TypeExtensions`, no matter of `cleanup` implementation

Tests/FuzzilliTests/WeakSetTest.swift
samo98: hmm and isn't it testing different thing? because that test would pass also ordinary `Set` holding pointers.

But I can make it `addLocalVariable(to: inout WeakSet<TypeExtension>)` to be specific

Sources/Fuzzilli/Core/Corpus.swift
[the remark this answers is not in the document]
saelo: Ah I see, good thought. Although I guess then you are assuming that a `defer` executes after local variable destructors (which could free some objects), which I'm not sure is the case. I'm fine both ways, but if you leave the `defer` please make a short comment to explain the rationale :)

Tests/FuzzilliTests/WeakSetTest.swift
saelo: Ah yeah that also sounds like a good solution!

Sources/Fuzzilli/Core/Corpus.swift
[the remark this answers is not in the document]
samo98: Actually I don't really like `defer` at all :D so I get rid of it gladly

Sources/Fuzzilli/Util/WeakSet.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(value)
+    }
+}
+
+public struct WeakSet<Element: AnyObject> where Element: Hashable {
+    private var elements: Set<WeakVar<Element> > = []
+
+    public var count: Int {
```
saelo, line 29: Could it make sense to not even expose a `count` for this since it might be ambiguous whether the count includes nil values or excludes them? The count only seems to be used in tests currently.

Tests/FuzzilliTests/WeakSetTest.swift
```diff
@@ -0,0 +1,66 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+@testable import Fuzzilli
+
+class WeakSetTests: XCTestCase {
+
+    func testWeakSetInsert() {
```
saelo, line 20: Maybe there should be another test that inserts 2 or more separate objects into the set and then all or some of them become `nil` to check that this works correctly (just because a set is normally not supposed to contain two identical values, but I guess in that case it would?)

Sources/Fuzzilli/Util/WeakSet.swift
samo98: yeah I agree, but I still would like to test `removeNils`, so maybe I can expose `nilCount` which returns number of nil elements?

saelo: I see. Yeah that also sounds like a good idea. I think you can also expose it just as `internal` protection level (just remove `public`, `internal` is the default) if its (for now) only used in tests

PR #127, opened 2020-08-28
Get rid of weakSet usage when deduplicating

samo98:


Sources/Fuzzilli/Core/Corpus.swift
```diff
@@ -171,14 +158,26 @@ public class Corpus: ComponentBase {
         ages.removeAll()
         newPrograms.forEach(add)
     }
+
+    /// Change type extensions for cached ones to save memory
+    private func deduplicateTypeExtensions(program: Program, with deduplicationSet: inout Set<TypeExtension>) {
```
saelo, line 163: Nit: can you make it `(in program: Program`? =)

saelo (approved):

PR #128, opened 2020-08-31
Collect runtime types only on variable definition

samo98:
Should help with operations `StoreProperty` & `DeleteProperty` as currently type collection just unions this information away.
This PR modifies it in a way `AbstractInterpreter` adopts runtimeType at definition and handles these intstructions on its own.

Also should save time on `Dup` or `Reassign` operations.

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -245,12 +245,7 @@ public class ProgramBuilder {
     
     /// Type information access.
     public func type(of v: Variable) -> Type {
-        let runtimeType = program.runtimeType(of: v)
-        if runtimeType == .unknown {
-            return interpreter?.type(of: v) ?? .unknown
-        } else {
-            return runtimeType
-        }
+        return interpreter?.type(of: v) ?? .unknown
     }
     
     public func methodSignature(of methodName: String, on object: Variable) -> FunctionSignature {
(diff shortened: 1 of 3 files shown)


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -245,12 +245,7 @@ public class ProgramBuilder {
     
     /// Type information access.
     public func type(of v: Variable) -> Type {
-        let runtimeType = program.runtimeType(of: v)
-        if runtimeType == .unknown {
-            return interpreter?.type(of: v) ?? .unknown
-        } else {
-            return runtimeType
-        }
+        return interpreter?.type(of: v) ?? .unknown
```
saelo, line 248: Nice. With that change, you can probably also remove this code: https://github.com/googleprojectzero/fuzzilli/blob/66de56c39e298e53f2a40c86f7b4607a56a37bc6/Sources/Fuzzilli/Core/ProgramBuilder.swift#L345 ?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -201,7 +201,13 @@ public struct AbstractInterpreter {
         return result
     }
     
-    private mutating func executeEffects(_ instr: Instruction) {
+    private mutating func executeEffects(_ instr: Instruction, with runtimeType: (Variable, Type)?) {
```
saelo, line 204: I'm not 100% happy about this API since the `Variable` parameter seems redundant, it's already in `instr.output`, no? But see also main comment

saelo (commented):
Excellent. I wonder if it would be easier to "inject" the runtime type information into the interpreter in another way: Afaict, the ProgramBuilder only ever gets runtime type information through the `adopt` method. Would it be possible to just add (possibly multiple, depending on the number of outputs) calls to https://github.com/googleprojectzero/fuzzilli/blob/66de56c39e298e53f2a40c86f7b4607a56a37bc6/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift#L125 after this function call: https://github.com/googleprojectzero/fuzzilli/blob/66de56c39e298e53f2a40c86f7b4607a56a37bc6/Sources/Fuzzilli/Core/ProgramBuilder.swift#L358 (or, alternatively, have `internalAppend` accept runtime type information from the adoption typeMap)? I'm also not sure if the other two `adopt` functions (https://github.com/googleprojectzero/fuzzilli/blob/66de56c39e298e53f2a40c86f7b4607a56a37bc6/Sources/Fuzzilli/Core/ProgramBuilder.swift#L338 and https://github.com/googleprojectzero/fuzzilli/blob/66de56c39e298e53f2a40c86f7b4607a56a37bc6/Sources/Fuzzilli/Core/ProgramBuilder.swift#L351) even need to be able to propagate type information, since that seems to be something that you'd only do when copying the entire instruction. Also, you can always manually overwrite the type through one of the type setters if really necessary.

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: The pair is there so we don't reimplement same logic twice https://github.com/googleprojectzero/fuzzilli/pull/128/files#diff-b12fd4c6f60ba74a8c5ba07622efca8dR63, becasue it can be either `output` or `innerOutput`

Sources/Fuzzilli/Core/ProgramBuilder.swift
samo98: hmm and where will runtime types be stored?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -335,27 +330,39 @@ public class ProgramBuilder {
...
     
     /// Adopts an instruction from the program that is currently configured for adoption into the program being constructed.
     public func adopt(_ instruction: Instruction, keepTypes: Bool) {
-        let newInouts = adopt(Array(instruction.inputs), keepTypes: false) + adopt(Array(instruction.allOutputs), keepTypes: keepTypes)
-        internalAppend(Instruction(operation: instruction.operation, inouts: newInouts))
+        let newInouts = adopt(Array(instruction.inputs)) + adopt(Array(instruction.allOutputs))
+        let adoptedInstruction = Instruction(operation: instruction.operation, inouts: newInouts)
+        internalAppend(adoptedInstruction)
+        if keepTypes {
+            adoptTypes(instruction.allOutputs, adoptedInstruction.allOutputs)
```
saelo, line 364: Would it be a bit nicer if the function signature was `adoptTypes(from origInstr: Instruction, to newInstr: Instruction)`?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -335,27 +330,39 @@ public class ProgramBuilder {
...
     /// Maps a list of variables from the program that is currently configured for adoption into the program being constructed.
-    public func adopt(_ variables: [Variable], keepTypes: Bool) -> [Variable] {
-        return variables.map{ adopt($0, keepType: keepTypes) }
+    public func adopt(_ variables: [Variable]) -> [Variable] {
+        return variables.map(adopt)
+    }
+
+    private func adoptTypes(_ originalOutputs: ArraySlice<Variable>, _ adoptedOutputs: ArraySlice<Variable>) {
+        for (originalVariable, adoptedVariable) in zip(originalOutputs, adoptedOutputs) {
+            if let type = typeMaps.last![originalVariable] {
```
saelo, line 348: I guess this could also already be `if let type = typeMaps.last![originalVariable], type != .unknown` (since the program runtime type probably doesn't need to be set to `.unknown` anyway). I'm fine either way though

PR #129, opened 2020-08-31
Add diagnostics mode

carl-smith:
Add a diagnostics mode to save executions that failed or timed-out.
This allows analysis of the output of the fuzzer.
Diagnostics mode also saves all executions made on a REPRL instance
to find scripts that may corrupt the state of the REPRL and saves all
executed scripts if it detects such a REPRL failure.

### DIFF
--- a/Sources/Fuzzilli/Core/Events.swift
+++ b/Sources/Fuzzilli/Core/Events.swift
@@ -64,6 +64,9 @@ public class Events {
     /// Signals that a new interesting program has been found, after the program has been minimized.
     public let InterestingProgramFound = Event<(program: Program, isImported: Bool, newTypeCollectionRun: Bool)>()
 
+    /// Signals that the REPRL instance encountered an error with the attached log of scripts.
+    public let REPRLFail = Event<[String]>()
+
     /// Signals that a program is about to be executed.
     public let PreExecute = Event<Program>()
     
--- a/Sources/Fuzzilli/Core/Events.swift
+++ b/Sources/Fuzzilli/Core/Events.swift
@@ -64,8 +64,8 @@ public class Events {
     /// Signals that a new interesting program has been found, after the program has been minimized.
     public let InterestingProgramFound = Event<(program: Program, isImported: Bool, newTypeCollectionRun: Bool)>()
 
-    /// Signals that the REPRL instance encountered an error with the attached log of scripts.
-    public let REPRLFail = Event<[String]>()
+    /// Signals a diagnostics event
+    public let DiagnosticsEvent = Event<(name: String, content: String)>()
 
     /// Signals that a program is about to be executed.
     public let PreExecute = Event<Program>()
--- a/Sources/Fuzzilli/Execution/REPRL.swift
+++ b/Sources/Fuzzilli/Execution/REPRL.swift
@@ -38,7 +38,7 @@ public class REPRL: ComponentBase, ScriptRunner {
     /// Buffer to hold scripts, this lets us debug issues that arise if
     /// previous scripts corrupted any state which is discovered in
     /// future executions.
-    private var scriptBuffer: [String]? = nil
+    private var scriptBuffer = String()
 
     public init(executable: String, processArguments: [String], processEnvironment: [String: String]) {
         self.processArguments = [executable] + processArguments
@@ -53,11 +53,6 @@ public class REPRL: ComponentBase, ScriptRunner {
         reprlContext = libreprl.reprl_create_context()
         if reprlContext == nil {
             logger.fatal("Failed to create REPRL context")
-        }
-
-        if fuzzer.config.diagnostics {
-            self.scriptBuffer = [String]()
-            self.scriptBuffer!.reserveCapacity(maxExecsBeforeRespawn)
         }
 
         let argv = convertToCArray(processArguments)
@@ -82,7 +77,7 @@ public class REPRL: ComponentBase, ScriptRunner {
     public func run(_ script: String, withTimeout timeout: UInt32) -> Execution {
         // Log the current script into the buffer if diagnostics are enabled.
         if fuzzer.config.diagnostics {
-            self.scriptBuffer!.append(script)
+            self.scriptBuffer += script + "\n"
         }
 
         let execution = REPRLExecution(in: reprlContext)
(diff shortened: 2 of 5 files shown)


Sources/Fuzzilli/Modules/Storage.swift
```diff
@@ -66,6 +77,42 @@ public class Storage: Module {
...
+            for (id, script) in scripts.enumerated() {
+                let filename = "/reprl_\(id).js"
+                // create the file URL
+                let fileURL = URL(fileURLWithPath: self.REPRLFailDir + dirname + filename)
+                self.storeProgram(script, to: fileURL)
+            }
+        }
+
+        fuzzer.registerEventListener(for: fuzzer.events.InvalidProgramFound) { program in
+            if fuzzer.config.diagnostics {
```
saelo, line 99: Can you turn this around and only even register the listener if diagnostics mode is active?
I think you can just do 
```
if fuzzer.config.diagnostics {
    fuzzer.registerEventListener(for: x) {
    ....
    }

    fuzzer.registerEventListener(for: y) {
        ...
```
?

Sources/Fuzzilli/Modules/Storage.swift
```diff
@@ -66,6 +77,42 @@ public class Storage: Module {
...
+                let filename = "failed_\(String(currentMillis())).js"
```
saelo, line 100: Maybe change name to `invalid_`?

Sources/Fuzzilli/Core/Events.swift
```diff
@@ -64,6 +64,9 @@ public class Events {
     /// Signals that a new interesting program has been found, after the program has been minimized.
     public let InterestingProgramFound = Event<(program: Program, isImported: Bool, newTypeCollectionRun: Bool)>()
 
+    /// Signals that the REPRL instance encountered an error with the attached log of scripts.
+    public let REPRLFail = Event<[String]>()
```
saelo, line 68: How about making this a generic `DiagnosticsEvent<name: String, content: String>` with the name/type used as filename and the (opaque) content string used as file content? Then this could be reusable for other things we might want to debug in the future. For example, we could more or less immediately reuse it to also output the entire "REPRL trace" for crashes in case they don't reproduce due to some "state leakage" from a previous program (but probably we'd need to give `config.diagnostics` some kind of verbosity level to not fill the disk with invalid programs, but that's for another time :) )
WDYT?

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -20,71 +20,89 @@ import libreprl
...
     private var recentlyFailedExecutions = 0
-    
+
     private var reprlContext: OpaquePointer? = nil
-    
+
+    /// Buffer to hold scripts, this lets us debug issues that arise if
+    /// previous scripts corrupted any state which is discovered in
+    /// future executions.
+    private var scriptBuffer: [String]? = nil
```
saelo, line 41: I don't think this needs to be an optional. Then those unsafe unwrapping operations (`!`) aren't needed. Maybe depending on the the above comment (`DiagnosticsEvent`), this can also just be a single string which is concatenated to I guess

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -20,71 +20,89 @@ import libreprl
...
     }
-    
+
     override func initialize() {
         reprlContext = libreprl.reprl_create_context()
         if reprlContext == nil {
             logger.fatal("Failed to create REPRL context")
         }
 
+        if fuzzer.config.diagnostics {
```
saelo, line 58: If `scriptBuffer` is just a string, this entire block can be removed

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -20,71 +20,89 @@ import libreprl
...
         }
-        
+
         execsSinceReset += 1
         var freshInstance: Int32 = 0
         if execsSinceReset > maxExecsBeforeRespawn {
             freshInstance = 1
             execsSinceReset = 0
+            if fuzzer.config.diagnostics {
+                scriptBuffer!.removeAll()
```
saelo, line 102: With `scriptBuffer` being a string, this can use https://developer.apple.com/documentation/swift/string/1641517-removeall and do `keepingCapacity: true`

Sources/Fuzzilli/Core/Events.swift
carl-smith: Sounds good! :)

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -20,30 +20,35 @@ import libreprl
...
     private var recentlyFailedExecutions = 0
-    
+
     private var reprlContext: OpaquePointer? = nil
-    
+
+    /// Buffer to hold scripts, this lets us debug issues that arise if
+    /// previous scripts corrupted any state which is discovered in
+    /// future executions.
+    private var scriptBuffer = String()
```
saelo, line 41: Maybe add a comment that this is only used if diagnostics mode is enabled?

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -93,11 +106,15 @@ public class REPRL: ComponentBase, ScriptRunner {
             // to execute this program. If we repeatedly fail to execute any program, we abort.
             if status < 0 {
                 logger.warning("Script execution failed: \(String(cString: reprl_get_last_error(reprlContext))). Retrying in 1 second...")
+                if fuzzer.config.diagnostics {
+                    // Log the buffer to disk
```
saelo, line 110: Maybe remove this comment, since that only happens if the Storage module is active (and also since other users could react to this event)

Sources/Fuzzilli/Modules/Storage.swift
```diff
@@ -66,6 +77,33 @@ public class Storage: Module {
             self.storeProgram(code, to: fileURL)
         }
 
+        if fuzzer.config.diagnostics {
+            fuzzer.registerEventListener(for: fuzzer.events.DiagnosticsEvent) { ev in
+                switch ev.name {
+                    case "REPRLFail":
```
saelo, line 83: This can be more generic, no? I.e. just use name + timestamp as filename and put all of these into a dir named e.g. `diagnostics`.

Sources/Fuzzilli/Modules/Storage.swift
```diff
@@ -22,6 +22,9 @@ public class Storage: Module {
     private let interestingDir: String
     private let statisticsDir: String
     private let stateFile: String
+    private let failedDir: String
+    private let timeOutDir: String
+    private let DiagnosticsDir: String
```
saelo, line 27: Nit: lower case `d` as first char

PR #131, opened 2020-09-02
Improve splicing by re-wiring inputs

carl-smith:


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,33 +388,61 @@ public class ProgramBuilder {
...
+
+        // This maps victim instruction indices to victim : host variable remap
+        // Instead of calling adopt and then using nextvar if the variable is
+        // not in the varMaps map, we do the adoption manually.
+        var rewireMapping: [Int: (Variable, Variable)] = [:]
+
+        func rewireOrKeepInputs(instruction: Instruction) {
+            var inputs: [Variable] = []
+            for input in instruction.inputs {
+                if probability(0.5) {
```
saelo, line 415: My intuition is that this is too high, I'd think more of something in the range 0.10 to 0.25, but maybe this can be evaluated somehow? I'd guess a total replacement rate of maybe 10% would be reasonable

saelo: Also I think it should only be done if `mode != .conservative`

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -433,13 +461,35 @@ public class ProgramBuilder {
...
-        // Insert the slice into the currently mutated program
-        adopting(from: program) {
-            for instr in program {
-                if needs.contains(instr.index) {
-                    adopt(instr, keepTypes: true)
+
+        // This maps victim vars to host vars
+        var localVarMaps: [Variable: Variable] = [:]
+
+        // manual adoption
```
saelo, line 468: I'm not really happy about this due to code duplication...
How about:
* You make a helper getter, `Program.lastVariable`, that gives you the last defined variable in a program (== the one with the highest number), and somewhere at the start of this function do `var placeholderVariable = program.lastVariable + 1` 
* Around line 400, you call `beginAdoption(from: program)`, and call `endAdoption()` at the end of this function
* When rewiring an input, you (1) replace the rewired variable with the next placeholder variable: `placeholderVariable++` and (2) manually create the placeholderVar -> rewiredVar mapping in `varMaps.last!` (feel free to create a new, private API for this)

Then here, you can just do `adopt(instr)` I think
WDYT? Would that work?

saelo (commented):
Very cool!

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Yes, I think we should test different values and evaluate the effect. I think it would be good to be able to increase the aggressiveness over the course of a fuzzing session.

carl-smith: I agree, we should only splice and re-wire if `mode == .aggressive`.

carl-smith: Or like you said `if mode != .conservative` if we intend to add another even more aggressive mode.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,33 +388,61 @@ public class ProgramBuilder {
     /// Append a splice from another program.
     public func splice(from program: Program, at index: Int) {
         var idx = index
-        
+
+        // Get types of the other program
+        var ai = AbstractInterpreter(for: self.fuzzer)
+
+        for instr in program {
+            ai.execute(instr)
```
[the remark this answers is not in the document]
carl-smith: I think it would be easier to just use the AI unconditionally here and then do something like:
```
let type = program.runtimeTypes[input] ?? ai.type(of: input)
```
Currently the `runtimeTypes` map can be accessed by a function which returns the type or `.unknown` in which case I think the AI might provide better type info, which is why I am accessing it here directly to get an optional value.

Then we could also do something like:
```
let replacementVar = randVar(ofConservativeType: type) ?? randVar(ofConservativeType: type.generalize())
if let hostVar = replacementVar {
    // do rewire
}
```
Or we could try to aggregate all the type info in the AI like you said and use the AI's interface for type queries? 
What do you think? :)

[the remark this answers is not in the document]
saelo: Yeah I guess we can leave it like this until we have a proper type information API that incorporates both interpreter-inferred and runtime-collected type information :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,93 @@ public class ProgramBuilder {
...
         var requiredInputs = VariableSet()
-        
+
+        // This maps victim instruction indices to victim : host variable remap
+        // Instead of calling adopt and then using nextvar if the variable is
+        // not in the varMaps map, we do the adoption manually.
+        func rewireOrKeepInputs(instruction: Instruction) {
+            var inputs: [Variable] = Array(instruction.inputs)
+            var needInputs: [Variable] = []
+            for inputIdx in 0..<instruction.inputs.count {
```
saelo, line 431: Here you could use [`.enumerated()`](https://developer.apple.com/documentation/swift/array/1687832-enumerated) to get both the value and the index

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,93 @@ public class ProgramBuilder {
...
+                var didRewire = false
+                if probability(0.2) && mode != .conservative {
+                    // TODO(cffsmith): switch to runtime type information when available.
+                    // In the future, this will return .unknown if there is no type information available, if we are in
```
saelo, line 435: I think you should already do this, since the interpreter already marks quite a few things as `.unknown`. Also, in .conservative mode we currently don't do input rewiring at all :)

Sources/Fuzzilli/FuzzIL/Program.swift
```diff
@@ -32,18 +32,24 @@ public final class Program: Collection {
 
     /// Result of runtime type collection execution, by default there was none
     public var typeCollectionStatus = TypeCollectionStatus.notAttempted
-    
+
+    public var lastVariable: Int {
+        return instructions.reduce(0, {cur, instr in
+            cur + instr.allOutputs.count
```
saelo, line 38: Maybe this function should be called `nextFreeVariable` since this function currently returns 0 for an empty program, which would be incorrect?
Also maybe you can compute the value without traversing all instructions with something like
```
for instr in instructions.reversed() {
    if let r = instr.allOutputs.max() {
        return r.number + 1
    }
}
return 0
```
WDYT?

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: 👍

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Sounds good! I was not sure if we wanted to change this later on, e.g. if we wanted to introduce another level of aggressiveness but let's change it for now :)

Sources/Fuzzilli/FuzzIL/Program.swift
carl-smith: Ah you're right, we should call it `nextFreeVariable` :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
     /// Append a splice from another program.
     public func splice(from program: Program, at index: Int) {
         var idx = index
-        
+
+        // As programs are ref counted, we create copy such that we do not
+        // modify the one in the corpus.
+        let workingCopy = program.copy()
```
saelo, line 399: Nit: maybe rename this to e.g. `source`? and maybe change the comment to something like `The input rewiring algorithm modifies the source program, thus we make a copy of it here` to make it clear who modifies the program?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
...
+                        didRewire = true
+                    }
+                }
+                // If we did not rewire this input, we will need this input from the victim program.
+                if !didRewire {
+                    needInputs.append(input)
+                }
+            }
+            // Rewrite the instruction with the new inputs.
+            workingCopy.replace(instructionAt: instruction.index, with: Instruction(operation: instruction.operation, inouts: inputs + Array(instruction.allOutputs)))
```
saelo, line 449: Can you do this only if any inputs were rewired? Because then, in the case of `mode == .conservative`, the `program.copy()` at the start will essentially be free due to CoW arrays
I think you can determine whether any variables were rewired either by checking if `neededInputs == instruction.inputs` or by checking if `inputs == instruction.inputs`, or by explicitly tracking it via a boolean variable (e.g. the `didRewire`) or so?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -337,6 +337,11 @@ public class ProgramBuilder {
 
         return varMaps.last![variable]!
     }
+
+    public func rewireMap(host hostVariable: Variable, victim victimVariable: Variable) {
```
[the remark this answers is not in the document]
saelo: Also I think the name is not very descriptive. Maybe make it `createVariableMapping(from: Variable, to: Variable)` or so? I think generally the term `victim` is not very clear in this context. Maybe `source` is better?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
...
         var needs = Set<Int>()
         var requiredInputs = VariableSet()
-        
+
+        // This maps victim instruction indices to victim : host variable remap
+        // Instead of calling adopt and then using nextvar if the variable is
+        // not in the varMaps map, we do the adoption manually.
+        func rewireOrKeepInputs(instruction: Instruction) {
+            var inputs: [Variable] = Array(instruction.inputs)
+            var needInputs: [Variable] = []
```
saelo, line 430: Nit: maybe make it `neededInputs`

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
...
+        }
+
+        beginAdoption(from: workingCopy)
+
         // Determine all necessary input instructions for the choosen instruction
         // We need special handling for blocks:
         //   If the choosen instruction is a block instruction then copy the whole block
         //   If we need an inner output of a block instruction then only copy the block instructions, not the content
         //   Otherwise copy the whole block including its content
         var needs = Set<Int>()
```
saelo, line 422: Maybe this should be named `requiredInstructions` to make it more consistent with the following variable?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
...
         var requiredInputs = VariableSet()
-        
+
+        // This maps victim instruction indices to victim : host variable remap
+        // Instead of calling adopt and then using nextvar if the variable is
+        // not in the varMaps map, we do the adoption manually.
+        func rewireOrKeepInputs(instruction: Instruction) {
+            var inputs: [Variable] = Array(instruction.inputs)
```
saelo, line 429: Nit: I think here you can drop the explicit type declaration (`: [Variable]`)?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
...
+            for (idx, input) in instruction.inputs.enumerated() {
+                var didRewire = false
+                if probability(0.2) && mode != .conservative {
+                    // TODO(cffsmith): switch to runtime type information when available.
+                    let type = ai.type(of: input) == .unknown ? .anything : ai.type(of: input)
+                    if let hostVar = randVar(ofConservativeType: type.generalize()) {
+                        let placeholderVariable = nextFreeVariable()
+                        inputs[idx] = placeholderVariable
+                        rewireMap(host: hostVar, victim: placeholderVariable)
+                        didRewire = true
```
saelo, line 440: Maybe instead of using this boolean variable, you can always do `neededInputs.append(input)` first, then, in here, do something like `neededInputs.removeLast()` or so? Would that also work?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -388,43 +393,90 @@ public class ProgramBuilder {
...
+        // Instead of calling adopt and then using nextvar if the variable is
+        // not in the varMaps map, we do the adoption manually.
+        func rewireOrKeepInputs(instruction: Instruction) {
+            var inputs: [Variable] = Array(instruction.inputs)
+            var needInputs: [Variable] = []
+            for (idx, input) in instruction.inputs.enumerated() {
+                var didRewire = false
+                if probability(0.2) && mode != .conservative {
+                    // TODO(cffsmith): switch to runtime type information when available.
+                    let type = ai.type(of: input) == .unknown ? .anything : ai.type(of: input)
```
saelo, line 435: Nit: maybe write this as
```
var type = ai.type(of: input)
if type == .unknown {
    type = .anything
}
```
(and maybe make a comment that we probably should have a method to do this for us =D )

saelo: And I guess you can drop the current comment since the runtime type information access refactoring will probably remove `AbstractInterpreter.type(of:)`, and so this code will have to be updated then anyway :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
carl-smith: 👍

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -337,6 +337,11 @@ public class ProgramBuilder {
 
         return varMaps.last![variable]!
     }
+
+    public func rewireMap(host hostVariable: Variable, victim victimVariable: Variable) {
+        assert(!varMaps.last!.contains(victimVariable))
+        varMaps[varMaps.count - 1][victimVariable] = hostVariable
```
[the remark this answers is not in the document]
carl-smith: It's actually consistent with the `adopt` function just above it :D 
Also accessing a subscript through `.last!` does not immediately work?

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: 👍

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: Ah my bad, `.last` cannot be assigned I guess. Yeah then let's just leave it as it is :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Yeah that sounds good! :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: 👍

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Yes that would sound more intuitive and also consistent.

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Yeah I had it like that but then I switched to this one-liner, was not sure what the best way is :)

saelo: Yeah, I find it a bit more readable like this, but it's also just to avoid calling the `type(of:)` method twice (even though that's fine since it currently has no side effects and does a simple lookup) :)

PR #133, opened 2020-09-03
Collect runtime types per instruction

samo98:
* Adjust runtimeTypes map to store types per (variable, instruction)
* Stop adopting types fom `OperationMutator` (Cases when we did can handle `AbstractInterpreter`)
* Add type collection on changing `__proto__`

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -305,7 +305,7 @@ public class ProgramBuilder {
     /// Required when copying instructions between program.
     ///
     private var varMaps = [VariableMap<Variable>]()
-    private var typeMaps = [VariableMap<Type>]()
+    private var typeMaps = [VariableMap<[Int: Type]>]()
     
     /// Prepare for adoption of variables from the given program.
     ///
@@ -348,10 +348,11 @@ public class ProgramBuilder {
         return variables.map(adopt)
     }
 
-    private func adoptTypes(from origInstr: Instruction, to newInstr: Instruction) {
-        for (originalVariable, adoptedVariable) in zip(origInstr.allOutputs, newInstr.allOutputs) {
-            if let type = typeMaps.last![originalVariable], type != .unknown {
-                program.setRuntimeType(of: adoptedVariable, to: type)
+    private func adoptTypes(from origInstr: Instruction) {
+        let newInstr = program.lastInstruction
+        for (originalVariable, adoptedVariable) in zip(origInstr.inouts, newInstr.inouts) {
+            if let instrMap = typeMaps.last![originalVariable], let type = instrMap[origInstr.index], type != .unknown {
+                program.setRuntimeType(of: adoptedVariable, to: type, at: newInstr.index)
 
                 interpreter?.setType(of: adoptedVariable, to: type)
             }
@@ -360,11 +361,9 @@ public class ProgramBuilder {
     
     /// Adopts an instruction from the program that is currently configured for adoption into the program being constructed.
     public func adopt(_ instruction: Instruction, keepTypes: Bool) {
-        let newInouts = adopt(Array(instruction.inputs)) + adopt(Array(instruction.allOutputs))
-        let adoptedInstruction = Instruction(operation: instruction.operation, inouts: newInouts)
-        internalAppend(adoptedInstruction)
+        internalAppend(Instruction(operation: instruction.operation, inouts: adopt(instruction.inouts)))
         if keepTypes {
-            adoptTypes(from: instruction, to: adoptedInstruction)
+            adoptTypes(from: instruction)
         }
     }
     
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -77,6 +77,8 @@ public class JavaScriptLifter: Lifter {
         // Analyze the program to determine the uses of a variable
         let analyzer = VariableAnalyzer(for: program)
 
+        let typeCollectionAnalyzer = TypeCollectionAnalyzer()
+
         // Need an abstract interpreter if we are dumping type information.
         var interpreter = AbstractInterpreter(for: self.environment)
 
@@ -84,11 +86,6 @@ public class JavaScriptLifter: Lifter {
         var expressions = VariableMap<Expression>()
         func expr(for v: Variable) -> Expression {
             return expressions[v] ?? Identifier.new(v.identifier)
-        }
-
-        func maybeUpdateType(_ variable: Variable) {
-            guard options.contains(.collectTypes) else { return }
-            w.emit("updateType(\(variable.number), \(expr(for: variable)));")
         }
 
         if options.contains(.collectTypes) {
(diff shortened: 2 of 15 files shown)


Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -40,6 +40,10 @@ public class JavaScriptLifter: ComponentBase, Lifter {
     /// For node.js this is "global", otherwise probably just "this".
     let globalObjectIdentifier: String
 
+    /// Special properties which significantly change variable type
+    /// We should recollect type if we change them
+    let specialProperties = ["__proto__"]
```
saelo, line 45: I'm not really happy that the Lifter is now responsible for deciding when to recollect type information. But we can worry about this later (and if other cases come up). In the simplest case, we can just create another Analyzer for that task :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -343,10 +343,11 @@ public class ProgramBuilder {
         return variables.map(adopt)
     }
 
-    private func adoptTypes(from origInstr: Instruction, to newInstr: Instruction) {
+    private func adoptTypes(from origInstr: Instruction) {
+        let newInstr = program.lastInstruction
         for (originalVariable, adoptedVariable) in zip(origInstr.allOutputs, newInstr.allOutputs) {
-            if let type = typeMaps.last![originalVariable], type != .unknown {
-                program.setRuntimeType(of: adoptedVariable, to: type)
+            if let instrMap = typeMaps.last![originalVariable], let type = instrMap[origInstr.index], type != .unknown {
```
saelo, line 349: IIUC, this currently does not make use of updated type information since it only looks at the instructions outputs when setting the type (and so, e.g. a type recollection on `obj.__proto__ = X` would not be processed here as `obj` is not an output)? But this probably also depends on the upcoming type information API refactoring, so I'm fine with leaving it this way for now

saelo: Yes, that would currently fix it I think. But in the future, maybe we'd like to be able to update types of variables not directly affected by an instruction (e.g. if we are clever enough to detect changes to common prototypes or something like that). Then this still wouldn't be enough.
Maybe another way to implement it (possibly after the upcoming refactoring) is to iterate over the runtime type data structure by instruction index instead of by variable. This might require converting the data structure once at the start of an adoption though.

samo98: hmm so the structure will have to be able to answer queries not only on variable but also on instruction, so it will be a bit more complex, so we just leave it to refactoring?

saelo: yes, let's do that. It would also be an option to rebuild a different data structure for `adopt` that can efficiently be queried by instruction instead of by variable. Because I think that query is only needed during adoption. Not 100% sure though :)

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -223,6 +225,11 @@ public class JavaScriptLifter: ComponentBase, Lifter {
                 let dest = MemberExpression.new() <> input(0) <> "." <> op.propertyName
                 let expr = AssignmentExpression.new() <> dest <> " = " <> input(1)
                 w.emit(expr)
+
+                // Recollect changes AbstractInterpreter cannot handle
```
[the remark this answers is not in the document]
samo98: well it is not that simple, because we need to call `matbeUpdateType` with correct arguments which are dependant on operation and therefore easier handled in big switch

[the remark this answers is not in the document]
samo98: What I can imagine is we create `maybeRecollectType` function in the same manner as `maybeUpdateType`
But calling `maybeRecollectType` will still happen in big switch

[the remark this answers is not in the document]
saelo: I guess what I'm suggesting is an interface such as
```
struct TypeRecollectionAnalyzer {
    func analyze(instr: Instruction) -> [Variable]
}
```

And then, at the end of the big switch in the JSLifter:
```
            if let expression = output {
                ...
            }
            for v in typeRecollectionAnalyzer.analyze(instr) {
                maybeUpdateType(of: v, instr)
            }
```
Since the analyzer now decides when to recollect type information, it probably also should decide what to recollect, so this interface would allow that without adding much complexity (as the lifter only has a single point where it deals with type recollection)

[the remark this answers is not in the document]
samo98: yeah so the big switch will be duplicated in analyzer too. I wanted to avoid it, but ok for now it will be simple anyway

Sources/Fuzzilli/Lifting/TypeCollectionAnalyzer.swift
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+// Determine what variable types we should collect given instruction
+struct TypeCollectionAnalyzer {
+    // Special properties which significantly change variable type
+    // We should recollect type if we change them
+    let specialProperties = ["__proto__"]
+
+    func analyze(_ instr: Instruction) -> [Variable] {
+        switch instr.operation {
```
saelo, line 22: Ah ok I see. I was initially suggesting to only let this Analyzer handle **re**collection cases and still have the JSLifter collect types for all outputs. But I think this is actually cooler! Because now all the logic about when to collect type information is in a single place :)
I'm wondering though, can you do something like this instead:
```
swift instr.operation {
case is LoadInt, is LoadBigInt, is LoadString, ...:
    // No need to collect types for literals. The interpreter can handle them
    return []
case is BeginAnyFunctionDefinition:
    // We don't collect parameter types ... (actually, I think maybe we should, but let's discuss some other time :D)
    return ...
case let op as StoreProperty where ...:
    variablesForCollection.append(instr.input(0))
default:
    // By default, collect types for all outputs of an instruction
    return Array(instruction.allOutputs)
}
return variablesForCollection
}
```
Then it's obvious what *isn't* collected, and it's also fail safe for the future, as by default, everything will be collected. WDYT?

Sources/Fuzzilli/Lifting/TypeCollectionAnalyzer.swift
```diff
@@ -0,0 +1,36 @@
...
+            case is LoadString, is LoadRegExp, is CreateObject, is CreateArray, is CreateObjectWithSpread, is Await,
+                 is CreateArrayWithSpread, is LoadBuiltin, is LoadProperty, is LoadElement, is LoadComputedProperty,
+                 is CallFunction, is CallMethod, is Construct, is CallFunctionWithSpread, is UnaryOperation,
+                 is BinaryOperation, is LoadFromScope:
+                return [instr.output]
+            case is BeginFor, is BeginForOf, is BeginCatch:
+                return [instr.innerOutput]
+            case let op as StoreProperty where specialProperties.contains(op.propertyName):
```
saelo, line 30: Oh nice use of `where`!

Sources/Fuzzilli/Lifting/TypeCollectionAnalyzer.swift
samo98: The question is what should be the default behavior.
1. optimistic approach (you suggest) - type collection of newly implemented instruction nearly always result in gathering all output types and it is correctly lifted in usual way (after lifting instruction, lift `updateType` with all output variables one by one). We already catched special cases and if new instruction is somehow special programmer should know about this and fix it himself (otherwise unexpected behavior)

2. conservative approach (originally implemented) - When new instruction is implemented, programmer has to update this function to say how we should collect type information for this new instruction explicitely. If programmer forgets, there is no unexpected behavior, just no benefits from type collecting new instruction

I don't have strong opinion here, so can go with your one

PR #136, opened 2020-09-03
Improve RegExp generation

carl-smith:


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -99,17 +99,35 @@ public class ProgramBuilder {
...
+            }
+            return string
         }, {
             chooseUniform(from: self.fuzzer.environment.interestingRegExps)
         }, {
-            self.genRegExp() + self.genRegExp()
+            withEqualProbability({
+                self.genRegExp() + self.genRegExp()
+            }, {
+                self.genRegExp(withBase: self.genRegExp())
```
saelo, line 128: Cool idea! I think this implementation is not ideal though, because in 2/3 of the cases the base will not be used. I think instead the high level logic should be something like:
withEqualProbability:
* generate a "base regex" (this is essentially the block starting in line 104)
* genRegExp() + genRegExp()
* base = genRegExp(), then add a quantifier or enclose in a group (or maybe also both so you can get `(foo)?` or so?). Maybe here you can also add `^` with a small probability to the start of a character set
WDYT about that?

carl-smith: Yeah that was also something I was worried about, sounds like a good plan to do it like you described :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -99,17 +99,35 @@ public class ProgramBuilder {
...
-    public func genRegExp() -> String {
-        // TODO: add genRegExpPatterns with groups etc.
-        let regexp = withEqualProbability({
-            String.random(ofLength: 2)
+    public func genRegExp(withBase base: String? = nil) -> String {
+        return withEqualProbability({
+            var string = base ?? withEqualProbability({
+                return String.random(ofLength: Int.random(in: 1...3))
+            }, {
+                return chooseUniform(from: self.fuzzer.environment.interestingRegExps)
```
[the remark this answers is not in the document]
carl-smith: good idea!

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Hm but the current code still doesn't really solve this issue, does it? Although the probability that `base` isn't used is now only 1/2 I guess.
But shouldn't something like this work:
```
public func genRegExp() -> String {
    // Generate a "base" regexp
    var regex = ""
    let desiredLength = Int.random(in: 1...4) 
    while string.count < desiredLength {
        regex += withEqualProbability({
            String.random(ofLength: 1)
        }, {
            chooseUniform(from: self.fuzzer.environment.interestingRegExps)
        })
    }
 
    // Now optionally concatenate with another regexp
    if (probability(0.3)) {
        regex += genRegExp()
    }

    // Or add a quantifier
    if (probability(0.2)) {
        regex += chooseUniform(from: self.fuzzer.environment.interestingRegExpQuantifiers)
    }

    // Or wrap in brackets
    if probability(0.1) {
        withEqualProbability({
            // optionally invert the character set
            if probability(0.2) {
                string = "^" + string
            }
            return "[" + string + "]"
        }, {
            return "(" + string + ")"
        })
    }
    return regex        
}
```
? This also makes the final probabilities a bit more obvious I think.

PR #137, opened 2020-09-06
Updates JSC patch and revision

WilliamParks:
This PR updates the JSC patch, as the previous one was failing due to changes in Source/JavaScriptCore/jsc.cpp


saelo:
Thanks!

PR #139, opened 2020-09-09
FuzzEngine interface and Multi-FuzzEngine support

carl-smith:
Currently one test still needs to be fixed, not sure yet if we want to change the test or adapt the `FuzzEngine` interface.

### DIFF
--- a/Sources/Fuzzilli/Core/MultiEngine.swift
+++ b/Sources/Fuzzilli/Core/MultiEngine.swift
@@ -27,7 +27,7 @@ public class MultiEngine: ComponentBase, FuzzEngine {
     private let roundsPerEngine: Int
 
     /// The number of rounds for the current active engine.
-    private var activeFuzzRounds = 0
+    private var currentIteration = 0
 
     public init(engines: WeightedList<FuzzEngine>, initialActive: FuzzEngine? = nil, roundsPerEngine: Int) {
         assert(roundsPerEngine > 0)
@@ -45,11 +45,10 @@ public class MultiEngine: ComponentBase, FuzzEngine {
 
     public func fuzzOne(_ group: DispatchGroup) {
         activeEngine.fuzzOne(group)
-        if activeFuzzRounds == roundsPerEngine {
-            activeFuzzRounds = 0
+        currentIteration += 1
+        if currentIteration % roundsPerEngine == 0 {
+            currentIteration = 0
             activeEngine = engines.randomElement()
-        } else {
-            activeFuzzRounds += 1
         }
     }
 }
--- a/Sources/Fuzzilli/Core/MultiEngine.swift
+++ b/Sources/Fuzzilli/Core/MultiEngine.swift
@@ -23,15 +23,15 @@ public class MultiEngine: ComponentBase, FuzzEngine {
     /// The current active engine.
     private var activeEngine: FuzzEngine
 
-    /// The number of rounds to complete per engine.
-    private let roundsPerEngine: Int
+    /// The number of fuzzing iterations to complete per engine.
+    private let iterationsPerEngine: Int
 
-    /// The number of rounds for the current active engine.
+    /// The number of fuzzing iterations.
     private var currentIteration = 0
 
-    public init(engines: WeightedList<FuzzEngine>, initialActive: FuzzEngine? = nil, roundsPerEngine: Int) {
-        assert(roundsPerEngine > 0)
-        self.roundsPerEngine = roundsPerEngine
+    public init(engines: WeightedList<FuzzEngine>, initialActive: FuzzEngine? = nil, iterationsPerEngine: Int) {
+        assert(iterationsPerEngine > 0)
+        self.iterationsPerEngine = iterationsPerEngine
         self.engines = engines
         self.activeEngine = initialActive ?? engines.randomElement()
         super.init(name: "MultiEngine")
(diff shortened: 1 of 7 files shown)


Sources/Fuzzilli/Core/MultiEngine.swift
```diff
@@ -0,0 +1,38 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+/// Wraps multiple engines into one, which can be initialized given a
+/// WeightedList. This can then switch engines to use, or use a more
+/// complicated heuristic.
+public class MultiEngine: ComponentBase, FuzzEngine {
```
saelo, line 20: Cool! So I think I'd like this to have an `activeEngine` member which is then used for a certain (configurable?) number of `fuzzOne` rounds before it is replaced by `engines.randomElement()`. Moreover, the initial `active` engine should be configurable. This way, you can e.g. say that you want to start with N rounds of fuzzing from the HybridEngine/GeneratorEngine to generate some Corpus before running the MutationEngine. WDYT?

carl-smith: Yeah, I had something similar for testing and actually forgot to remove the `iterationCount` variable :D

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -35,12 +35,23 @@ extension BaseInstructionMutator {
...
+        /// The mutation fuzzer responsible for mutating programs from the corpus and evaluating the outcome.
+        let mutators = WeightedList<Mutator>([
+            (CodeGenMutator(),   1),
+            (OperationMutator(), 1),
+            (InputMutator(),     1),
+            (CombineMutator(),   1),
+            (JITStressMutator(), 1),
+        ])
+
+        let engine = MutationEngine(mutators: mutators, numConsecutiveMutations: 5)
```
saelo, line 47: How about you let `makeMockMutationEngine` return a `MutationEngine`, then you can replace these lines with just a call to `makeMockMutationEngine()`

Sources/Fuzzilli/Core/MultiEngine.swift
```diff
@@ -0,0 +1,54 @@
...
+    let engines: WeightedList<FuzzEngine>
+
+    /// The current active engine.
+    private var activeEngine: FuzzEngine
+
+    /// The number of rounds for the current active engine.
+    private var activeFuzzRounds = 0
+
+    /// The number of rounds to complete per engine.
+    private let roundsPerEngine = 5
```
saelo, line 30: Maybe this should also be a parameter of the constructor?

Sources/Fuzzilli/Core/MultiEngine.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    public init(engines: WeightedList<FuzzEngine>, initialActive: FuzzEngine? = nil) {
+        self.engines = engines
+        self.activeEngine = initialActive ?? engines.randomElement()
+        super.init(name: "MultiEngine")
+    }
+
+    override func initialize() {
+        assert(roundsPerEngine > 0, "roundsPerEngine has to be at least 1")
```
saelo, line 39: With the above, this can go into the constructor

Sources/Fuzzilli/Core/MultiEngine.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    override func initialize() {
+        for engine in engines {
+            engine.initialize(with: self.fuzzer)
+        }
+    }
+
+    public func fuzzOne(_ group: DispatchGroup) {
+        activeEngine.fuzzOne(group)
+        if activeFuzzRounds == roundsPerEngine {
```
saelo, line 48: Maybe
```
currentIteration += 1
if currentIteration % iterationsPerEngine == 0 {
    activeEngine = engines.randomElement()
}
```
But I'm fine either way, whatever you find more readable :)

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -35,12 +35,14 @@ extension BaseInstructionMutator {
 class MutationsTests: XCTestCase {
 
     func testPrepareMutationRuntimeTypes() {
-        let fuzzer = makeMockFuzzer()
+        let engine = makeMockMutationEngine()
+
```
saelo, line 39: Yep nice! Nit: remove this blank line maybe?

Sources/Fuzzilli/Core/MultiEngine.swift
```diff
@@ -0,0 +1,54 @@
...
+        for engine in engines {
+            engine.initialize(with: self.fuzzer)
+        }
+    }
+
+    public func fuzzOne(_ group: DispatchGroup) {
+        activeEngine.fuzzOne(group)
+        currentIteration += 1
+        if currentIteration % roundsPerEngine == 0 {
+            currentIteration = 0
```
saelo, line 50: With the modulo operation in the `if`, you don't need to reset the `currentIteration` variable here I think (Otherwise it should maybe be called `iterationsForActiveEngine` or so instead). If you do that, please update the comment in line 29 as well. Also, please use either `rounds` or `iterations` consistently in this file =)

PR #141, opened 2020-09-11
Program types struct

samo98:
Implement dedicated struct to store types of program variables

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -22,7 +22,7 @@ public class ProgramBuilder {
     
     /// The code and type information of the program that is being constructed.
     private var code = Code()
-    public var runtimeTypes = VariableMap<[Int: Type]>()
+    public var runtimeTypes = ProgramTypes()
     
     public enum Mode {
         /// In this mode, the builder will try as hard as possible to generate semantically valid code.
@@ -71,7 +71,7 @@ public class ProgramBuilder {
         seenPropertyNames.removeAll()
         seenIntegers.removeAll()
         code.removeAll()
-        runtimeTypes.removeAll()
+        runtimeTypes = ProgramTypes()
         scopeAnalyzer = ScopeAnalyzer()
         contextAnalyzer = ContextAnalyzer()
         interpreter?.reset()
@@ -332,7 +332,8 @@ public class ProgramBuilder {
     /// Required when copying instructions between program.
     ///
     private var varMaps = [VariableMap<Variable>]()
-    private var typeMaps = [VariableMap<[Int: Type]>]()
+    /// Formatted ProgramTypes structure for easier adopting of runtimeTypes
+    private var typeMaps = [[[(Variable, Type)]]]()
     
     /// Prepare for adoption of variables from the given program.
     ///
@@ -340,7 +341,18 @@ public class ProgramBuilder {
     /// currently constructed one to avoid collision of variable names.
     public func beginAdoption(from program: Program) {
         varMaps.append(VariableMap())
-        typeMaps.append(program.runtimeTypes)
+        typeMaps.append(formatTypeMap(from: program))
+    }
+
+    // Format ProgramTypes struct so searching for type changes at instruction is easier
+    private func formatTypeMap(from program: Program) -> [[(Variable, Type)]] {
+        var typeMap: [[(Variable, Type)]] = Array(repeating: [], count: program.size)
+        for (variable, typeData) in program.runtimeTypes {
+            for typeIndexPair in typeData {
+                typeMap[typeIndexPair.index].append((variable, typeIndexPair.type))
+            }
+        }
+        return typeMap
     }
     
     /// Finishes the most recently started adoption.
@@ -375,15 +387,11 @@ public class ProgramBuilder {
         return variables.map(adopt)
     }
 
-    private func adoptTypes(from origInstr: Instruction) {
-        let newInstr = code.lastInstruction
-        for (originalVariable, adoptedVariable) in zip(origInstr.inouts, newInstr.inouts) {
-            if let instrMap = typeMaps.last![originalVariable], let type = instrMap[origInstr.index], type != .unknown {
-                // TODO remove code duplication between this and Program.setRuntimeType. Probably by moving this logic into a TypeInformation data structure?
-                if runtimeTypes[adoptedVariable] == nil {
-                    runtimeTypes[adoptedVariable] = [:]
-                }
-                runtimeTypes[adoptedVariable]![newInstr.index] = type
+    private func adoptTypes(at origInstrIndex: Int) {
+        for (variable, type) in typeMaps.last![origInstrIndex] {
+            // No need to keep unknown type nor type of not adopted variable
+            if type != .unknown, let adoptedVariable = varMaps.last![variable] {
+                runtimeTypes.setType(of: adoptedVariable, to: type, at: code.lastInstruction.index)
                 interpreter?.setType(of: adoptedVariable, to: type)
             }
         }
@@ -393,7 +401,7 @@ public class ProgramBuilder {
     public func adopt(_ instr: Instruction, keepTypes: Bool) {
         internalAppend(Instruction(instr.op, inouts: adopt(instr.inouts)))
         if keepTypes {
-            adoptTypes(from: instr)
+            adoptTypes(at: instr.index)
         }
     }
     
--- a/Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
+++ b/Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
@@ -1,0 +1,79 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// Store known types for program variables at specific instructions
+public struct TypeIndexPair: Equatable {
+    public let index: Int
+    public let type: Type
+
+    public init(index: Int, type: Type){
+        self.index = index
+        self.type = type
+    }
+
+    public static func == (lhs: TypeIndexPair, rhs: TypeIndexPair) -> Bool {
+        return lhs.index == rhs.index && lhs.type == rhs.type
+    }
+}
+
+public struct ProgramTypes: Equatable, Sequence {
+    private var types = VariableMap<[TypeIndexPair]>()
+
+    public init () {}
+
+    public init (from types: VariableMap<[TypeIndexPair]>) {
+        self.types = types
+    }
+
+    // Create structure in simple case, when we have only types on definition
+    public init (from types: VariableMap<Type>, in program: Program) {
+        let analyzer = VariableAnalyzer(for: program)
+        for (variable, type) in types {
+            setType(of: variable, to: type, at: analyzer.definition(of: variable).index)
+        }
+    }
+
+    public static func == (lhs: ProgramTypes, rhs: ProgramTypes) -> Bool {
+        return lhs.types == rhs.types
+    }
+
+    // Save type of given variable
+    public mutating func setType(of variable: Variable, to type: Type, at instrIndex: Int) {
+        // Initialize type structure for given variable if not already
+        if types[variable] == nil {
+            types[variable] = []
+        }
+        let typeIndexPair = TypeIndexPair(index: instrIndex, type: type)
+        // Check if we update type of instruction in the middle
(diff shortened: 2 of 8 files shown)


Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -0,0 +1,90 @@
...
+        {
+            return typeData.type
+        }
+
+        return .unknown
+    }
+
+    // Get type of variable at current instruction only if it is explicitely saved (type changes at this instruction)
+    // Used in adoption, so we do not adopt types which are not necessary
+    public func getSavedType(of variable: Variable, at instrIndex: Int) -> Type {
```
saelo, line 77: Maybe call this `getTypeIfModified` or something like that? I think that'd be a bit more intuitive

saelo: And I think this should return `nil` if the type was not modified

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    // Save type of given variable
+    public mutating func setType(of variable: Variable, to type: Type, at instrIndex: Int) {
+        // Initialize type structure for given variable if not already
+        if types[variable] == nil {
+            types[variable] = []
+        }
+        let typeIndexPair = TypeIndexPair(index: instrIndex, type: type)
+        // Check if we update type of instruction in the middle
+        if let insertionIndex = types[variable]!.firstIndex(where: { el in el.index >= instrIndex }) {
```
saelo, line 51: Here and below, if you wanted, I think you could shorten the closure to just `{ $0.index >= instrIndex }`

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -68,16 +77,20 @@ class MutationsTests: XCTestCase {
         let v3 = b.binary(b.loadInt(1), b.loadInt(2), with: .Add)
         b.unary(.BitwiseNot, v3)
         let program = b.finalize()
-        var types: [[Int: Type]?] = [[0: .string], [1: .integer], [2: .integer], [3: .integer], [4: .integer]]
-        program.runtimeTypes = VariableMap<[Int: Type]>(types)
+        var types: [[TypeIndexPair]?] = [
+            [TypeIndexPair(index: 0, type: .string)], [TypeIndexPair(index: 1, type: .integer)],
```
saelo, line 81: I think we should have a convenience constructor for `ProgramTypes` for the (common - at least in tests) case that every variable only has a single type (at the point where it is defined).
That probably requires determining when the variable is defined, which can by done by the [VariableAnalyzer](https://github.com/googleprojectzero/fuzzilli/blob/759e3419159cd1a746f5253b56a99108f673a314/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L62)

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    // Get type of variable at current instruction
+    public func getType(of variable: Variable, at instrIndex: Int) -> Type {
+        if let instrTypes = types[variable],
+           let typeData = instrTypes.last(where: { el in el.index <= instrIndex })
+        {
+            return typeData.type
+        }
+
+        return .unknown
```
saelo, line 72: I wonder if there is something else we should do here in the case where the variable exists, and we have type information for it, but isn't yet defined at the queried index (which would probably be a bug in the user code). We could return `.nothing`, but maybe we should for now just assert that that case doesn't happen (I'm just slightly worried that simply returning `.unknown` here might hide some bugs in the future).

samo98: hmm but how do we decide whether variable does not exist at all or we just don't have data for it? This should not be responsible of this struct. If we want some additional checks related to program, it should be wrapped by Program API

Tests/FuzzilliTests/MutationsTest.swift
samo98: yeah we can have some constructor to make it easier, but analyzer is tied with program and this structure is not

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
samo98: hmm I don't really like having two nils (nil & .unknown).
But there is other way around so we don't need this function at all.

When beginning adoption, we do not append to `typeMaps` this struct but rather inverted one [index] -> [(Var, Type)]
Then it will be easy to adopt just runtime types that were present in original program

Tests/FuzzilliTests/MutationsTest.swift
saelo: The analyzer should also be fine with just `Code` (I guess we just need to change the constructor or add a new one), but yes, we can also remove the name "Program" from this structure, don't have a strong preference :)

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
saelo: I was thinking about something along the lines of
```
// Currently, we disallow querying the type of a variable before it's defined
assert(types[variable] == nil);
```
But yes, that assumes that we will either have no type at all or at least the type at the point of definition (which is probably sensible?). But I'm also fine with adding such a check later in the Program API.

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
saelo: With `nil` you can distinguish in the (probably rare?) case that the type is updated at the instruction but is actually set to .unknown (I guess that can only ever happen at the definition of a variable). But yeah, your suggestion sounds better, I agree :)

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
samo98: well in this struct we store also runtime types and for example we don't gather function types, literal types, ... at all

saelo: Yeah, but for those `types[variable]` will be `nil`, no?
As far as I understand, we either have `types[variable] == nil` if we don't have any information, or we will have an array of at least size 1, and the first entry contains the type at the definition of the variable. So the assert would work to ensure that users don't query types of yet-undefined variables.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -332,15 +332,16 @@ public class ProgramBuilder {
     /// Required when copying instructions between program.
     ///
     private var varMaps = [VariableMap<Variable>]()
-    private var typeMaps = [VariableMap<[Int: Type]>]()
+    /// Inverted ProgramTypes structure for easier adopting of runtimeTypes
+    private var typeMaps = [[Int: [(Variable, Type)]]]()
```
saelo, line 336: Would it be more efficient here to use a simple Array as outer data structure instead of a Map since the integers will be mostly contiguous and starting at 0?

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    // Get type of variable at current instruction
+    public func getType(of variable: Variable, at instrIndex: Int) -> Type {
+        return types[variable]?.last(where: { $0.index <= instrIndex })?.type ?? .unknown
+    }
+
+    // Create inverted structure, easier for getting type changes per instruction
+    public func getInvertedStructure() -> [Int: [(Variable, Type)]] {
```
saelo, line 69: I kind of feel like this code should live in `ProgramBuilder.beginAdoption` since it's somewhat of an implementation detail of the adoption algorithm and not really related to this data structure. WDYT?

samo98: hmm I see it as just swapping order of keys in 2-level dictionary, so anybody who wants to search mainly by second key can request inverted data structure.

But yeah we can move it to `ProgramBuilder` and if there will be someone else searching mainly by instruction we can move it back

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
samo98: There is no explicit rule if there is some type info then there is type on definition (and this struct should not rely on it).
For example we don't collect types on literal definitions, but there are instructions that can change them using side-effect (and in future we can somehow collect them)
```
const v0 = "a"
const v1 = v0++
```

So `We disallow querying the type of a variable before it's defined` would not be true.
Rather `Currently, we disallow querying the type of a variable before we have some type info.`

And still I think this struct should not care about variable definition and other stuff, this should be just storage.
Also these not-explicit rules are good source of bugs as well (somebody changes TypeCollectionAnalyzer and this storage will break, and we cannot enforce it by unit tests as it involves JS execution)

Sources/Fuzzilli/Core/ProgramBuilder.swift
samo98: There are only 40-50% of instructions covered (we don't collect literals,...)

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
saelo: Ok makes sense!

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: But wouldn't an array be more storage efficient even with 25% or less? I'd assume the Map would be overcommitted by roughly a factor of two, and, depending on how it is implemented, would use element structs for its implementation which are probably at least twice the size of an Int. Performance-wise the array should easily win in any case.

samo98: TBH I've never done any experiment to get the constant factor overhead.

But in that case we have a bit more complicated implementation and maybe this deserves own struct too?

saelo: Ah hm, I partly also suggested this because I thought it would make the implementation easier :'D
So I think it should be fine to make this `[[(Variable, Type)]]`, then initialize it fully in `beginAdoption`: `var typeMap = Array(repeating: [], count: program.size)`. That way, no nil handling is required during construction and use of the type maps. This should be pretty much the fastest in terms of execution speed also (which I mostly care about).
Regarding memory consumption, I think at least once the ProgramType information also includes the information from the AbstractInterpreter, we probably will have type information for most lines, so preallocating the array in full shouldn't cause much overhead. I also don't really think that using `nil` values for missing entries would bring much gain, since the size of that `nil` should also be the size of an empty array (the element size of the array must be the size of the largest element) (that assumes that empty arrays don't preallocate their backing memory on the heap though).

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -52,12 +52,17 @@ class MutationsTests: XCTestCase {
...
+        program.runtimeTypes = ProgramTypes(
+            from: VariableMap<Type>([.number, .boolean, .float]),
+            in: program
+        )
 
         let preparedProgram = engine.prepareForMutation(program)
-        XCTAssertEqual(
-            preparedProgram.runtimeTypes, VariableMap<[Int: Type]>([nil, nil, [2: .number], [3: .boolean], [5: .float]])
+        let expectedTypes =  ProgramTypes(
+            from: VariableMap<Type>([nil, nil, .number, .boolean, .float]),
```
saelo, line 62: Nice! Do you think it would be worth adding a VariableMap constructor `VariableMap(_ map: [Int: Value])`, then writing this as
```
from: VariableMap([
    2: .number,
    3: .boolean,
    4: .float,
])
```
Or something like that?

Tests/FuzzilliTests/MutationsTest.swift
```diff
@@ -52,12 +52,17 @@ class MutationsTests: XCTestCase {
         }
         b.endIf()
         let program = b.finalize()
-        program.runtimeTypes = VariableMap<[Int: Type]>([[0: .number], [1: .boolean], [3: .float]])
+        program.runtimeTypes = ProgramTypes(
+            from: VariableMap<Type>([.number, .boolean, .float]),
```
saelo, line 56: Nit: here and elsewhere, I think you can leave off the `<Type>`, the compiler should be able to infer that

Sources/Fuzzilli/Core/ProgramBuilder.swift
samo98: nice this is also implementation-wise easy.
Only one thing. Type information will not cover more instructions. Because there will be 2 separate structures. One for runtime types (for future adoption) and second for types infered by AbstractInterpreter (for queries)

saelo: Yes I think that's correct (if I understand correctly). I think there will the the `typeMap`, which is constructed once during `beginAdoption` for *another* (immutable) program and then only queried (until `endAdoption`) but not updated. Then there will be a `var types: ProgramTypes` member of the ProgramBuilder which will hold the types for the program *under construction* and which will be updated from adopted types and from the interpreter. Then, during `finalize()`, that types data structure will be attached to the produced Program. Is that also how you imagine it will work?

PR #142, opened 2020-09-11
Add documentation for type determination

### LINKED ISSUE #34: Implement type collection at runtime

samo98:
Closes #34 

Document
* New runtime type collection feature
* Abstract interpreter
* Their cooperation

### DIFF
--- a/Sources/FuzzilliCli/main.swift
+++ b/Sources/FuzzilliCli/main.swift
@@ -61,8 +61,7 @@ Options:
     --dontFuzz                  : If used, this instace will not perform fuzzing. Can be useful for master instances.
     --noAbstractInterpretation  : Disable abstract interpretation of FuzzIL programs during fuzzing. See
                                   Configuration.swift for more details.
-    --collectRuntimeTypes       : Collect runtime types of variables in program when interesting one found.
-                                  Currently in development.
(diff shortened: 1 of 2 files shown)


Sources/FuzzilliCli/main.swift
```diff
@@ -62,7 +62,6 @@ Options:
     --noAbstractInterpretation  : Disable abstract interpretation of FuzzIL programs during fuzzing. See
                                   Configuration.swift for more details.
     --collectRuntimeTypes       : Collect runtime types of variables in program when interesting one found.
```
saelo, line 64: Unrelated to this PR, but can you make this something like "Collect runtime type information for programs that are added to the corpus"? I think that's a bit more clear

PR #144, opened 2020-09-12
Improved --importCorpusNewCov documentation

### LINKED ISSUE #143: Importing a Corpus as New Coverage Only Has Unclear Documentation
When importing a corpus using the New Coverage only flag, importing completes with only 1 program in the corpus.

Looking at this closer, the minimization queue run asynchronously, and starts after the importCorpus functions run. Thus, importing will report complete with 1 program, and the programs in the queue will be minimized asynchronously, and show up in the corpus later.

This is as intended, but is a bit  confusing based on the current document. I'll put together a PR to make this behavior clearer.

WilliamParks:
Closes #143. This makes it clear that since minimization is asynchronous, the corpus will show as small after completing the initial import (as intended).

### DIFF
--- a/Sources/FuzzilliCli/main.swift
+++ b/Sources/FuzzilliCli/main.swift
@@ -54,6 +54,8 @@ Options:
                                   This only includes programs that increase coverage.
                                   This is useful for jump starting coverage for a wide range of JavaScript samples.
                                   Can be used alongside importCorpusAll, and will run second.
+                                  Since all imported samples are asynchronously minimized, the corpus will show a smaller
+                                  than expected size until minimization completes.
     --importState=path          : Import a previously exported fuzzer state and resuming fuzzing from it.
     --networkMaster=host:port   : Run as master and accept connections from workers over the network. Note: it is
                                   *highly* recommended to run network fuzzers in an isolated network!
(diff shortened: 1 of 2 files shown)


Sources/FuzzilliCli/main.swift
```diff
@@ -54,6 +54,8 @@ Options:
                                   This only includes programs that increase coverage.
                                   This is useful for jump starting coverage for a wide range of JavaScript samples.
                                   Can be used alongside importCorpusAll, and will run second.
+                                  Since all imported samples are minimized first, and minimization happens asynchronously,
```
saelo, line 57: Nit: It's not "all" here, is it? Only those that are deemed interesting?

Sources/FuzzilliCli/main.swift
```diff
@@ -297,7 +299,7 @@ fuzzer.sync {
             let newProgs = try decodeProtobufCorpus(data)
             logger.info("Starting Cov-only corpus input of size \(newProgs.count). This may take some time")
             fuzzer.importCorpus(newProgs, importMode: .newCoverageOnly)
-            logger.info("Successfully imported coverage only corpus import. Corpus now contains \(fuzzer.corpus.size) elements")
+            logger.info("Successfully imported coverage only corpus import. Samples will appear in corpus as they are asynchronously minimized")
```
saelo, line 302: Nit: maybe remove "import" after "coverage only corpus"? Also, maybe make it "Samples will appear in the corpus once they have been minimized", it's probably not necessary to restate that this happens asynchronously here :)

saelo (commented):
Cool! How about you also log an info message once the import is completed? I think you should be able to do that at the end of `Fuzzer.importCorpus` by using [fuzzGroup.notify](https://developer.apple.com/documentation/dispatch/dispatchgroup/2016066-notify)

PR #145, opened 2020-09-14
implement HybridEngine and CodeTemplates

carl-smith:
Currently the OOM killer kills FuzzilliCli after a longer period of running. It is not yet clear if this is due to some issue in Fuzzilli or if the OOM killer decides to kill FuzzilliCli if the child JS engine process uses too much memory.


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -299,41 +303,125 @@ public class ProgramBuilder {
...
     public func generateCallArguments(forMethod methodName: String, on object: Variable) -> [Variable]? {
         let signature = methodSignature(of: methodName, on: object)
         return generateCallArguments(for: signature)
     }
-    
-    
+
+    /// Generates a sequence of instructions that generate the desired type.
+    func generateVariable(ofType type: Type) -> Variable {
+        switch type.baseType {
```
saelo, line 342: Maybe we can structure this a little different:

- First, a number of checks of this form:
    `if type.Is(.integer) || type.Is(fuzzer.environment. intType { // do the loadInt/loadFloat/createObject/... instruction directly` for all types here: https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Core/Environment.swift#L48
- assert that the type is now `.object` (I think this must be the case now)
- check if a group is present, if yes do the necessary thing
- construct a plain object with properties and methods, as you already do

I think this would get rid of some nesting levels and make the code easier to follow

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -299,41 +303,125 @@ public class ProgramBuilder {
...
-    
+
     public func generateCallArguments(forMethod methodName: String, on object: Variable) -> [Variable]? {
         let signature = methodSignature(of: methodName, on: object)
         return generateCallArguments(for: signature)
     }
-    
-    
+
+    /// Generates a sequence of instructions that generate the desired type.
```
saelo, line 340: Maybe add some comments here about what this function can and cannot generate?

Sources/Fuzzilli/Core/CodeTemplates.swift
```diff
@@ -0,0 +1,227 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// These CodeGenerators create more complex code samples and are better
+// suited for the HybridFuzzing mode.
+public let CodeTemplates: WeightedList<CodeGenerator> = WeightedList([
```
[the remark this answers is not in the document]
saelo: Maybe these stats can be moved into a new `struct`: `ProgramGeneratorStats`, which is then used by the Mutator and the CodeTemplate, and then given to the shared `FuzzEngine.execute` method as `inout` parameter which can update it according to the executions outcome?

Sources/Fuzzilli/Core/CodeTemplates.swift
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        let codeGeneratorDescriptions = ["IntegerGenerator",
+                "FloatGenerator", "BuiltinGenerator", "ArrayGenerator",
+                "ObjectGenerator", "BigIntGenerator", "RegExpGenerator" ]
+
+        // We might want to optimize this, depending on its performance.
+        let codeGenerators = CodeGenerators.get(codeGeneratorDescriptions)
+
+        // Generate a larger function
```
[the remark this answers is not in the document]
carl-smith: I changed it to generate a single function for now, I think we can create a second template for that :)

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    func run(in b: ProgramBuilder) {
+        f(b)
+    }
+
+    /// Generate an array of random function signatures
+    static func generateRandomFunctionSignatures(forFuzzer fuzzer: Fuzzer, n: Int) -> [FunctionSignature] {
+        var signatures = [FunctionSignature]()
+        signatures.append(CodeTemplate.generateSignature(forFuzzer: fuzzer, n: Int.random(in: 0...3)))
```
saelo, line 34: Should probably be `n: n`? :)

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,100 @@
...
+        return signatures
+    }
+
+    /// This function generates and sets property types for the global properties
+    static func generateRandomPropertyTypes(forBuilder b: ProgramBuilder) {
+        for _ in 0..<5 {
```
saelo, line 40: This should probably be a parameter `n` as well

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,100 @@
...
+            let name = chooseUniform(from: b.fuzzer.environment.customPropertyNames.dropLast())
+            b.setType(ofProperty: name, to: CodeTemplate.generateType(forFuzzer: b.fuzzer, forProperty: name))
+        }
+    }
+
+    /// Generate and set random method types for global method names.
+    static func generateRandomMethodTypes(forBuilder b: ProgramBuilder, n: Int) {
+        for _ in 0..<n {
+            b.setSignature(ofMethod: chooseUniform(from:
```
saelo, line 49: Nit: indention/line breaks are a bit weird here

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,100 @@
...
+                return .object(ofGroup: "Array")
+            },
+            // choose a complicated object
+            {
+                var properties: [String] = []
+                var methods: [String] = []
+
+                // Generate random properties
+                for _ in 1..<3 {
+                    let candidates = fuzzer.environment.customPropertyNames.filter({ $0 >= property })
```
saelo, line 74: Is this to avoid cycles? That should probably be explained in a comment :)

saelo: And maybe add a `TODO` that we should not have the "no-cycle" restriction. I think ultimately, the `generateVariable` function should be able to handle cycles (but let's not worry about that for now)

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    /// Generate a random type to use in e.g. function signatures.
+    /// This function should only emit types that can be constructed by ProgramBuilder.generateVariable.
+    static func generateType(forFuzzer fuzzer: Fuzzer, forProperty property: String = "") -> Type {
+        return withEqualProbability(
+            // Choose a basic type
+            { () -> Type in
+                chooseUniform(from: [.integer, .float, .boolean, .bigint])
```
saelo, line 61: What about strings?

Sources/Fuzzilli/Core/CodeTemplates.swift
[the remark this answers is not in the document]
saelo: Yes, nice, let's make a 2nd template!

Sources/Fuzzilli/Core/CodeTemplates.swift
```diff
@@ -0,0 +1,227 @@
...
+        let signature = generateSignature(forFuzzer: b.fuzzer, n: 4)
+        let f = b.definePlainFunction(withSignature: signature) { args in
+            for _ in 0..<2 {
+                b.run(chooseUniform(from: codeGenerators))
```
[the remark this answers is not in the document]
saelo: I think this + lines 41 and 46 can now just be removed since we have the prefix

Sources/Fuzzilli/Core/CodeTemplates.swift
```diff
@@ -0,0 +1,167 @@
...
+            // Generate (larger) function body
+            b.generate(n: 45)
+        }
+
+        // Generate some random instructions now
+        b.generate(n: genSize)
+
+        // trigger JIT
+        b.forLoop(b.loadInt(0), .lessThan, b.loadInt(100), .Add, b.loadInt(1)) { args in
+            // withGeneration == true makes this call infallible.
```
saelo, line 63: Nit: Stale comment

Sources/Fuzzilli/Core/SpliceEngine.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+public class SpliceEngine: ComponentBase, FuzzEngine {
```
saelo, line 17: I think it'd be best to make this file a separate PR

Sources/Fuzzilli/Core/MutationEngine.swift
```diff
@@ -153,69 +136,23 @@ public class MutationEngine: ComponentBase, FuzzEngine {
...
-        let prefixProgram = b.finalize()
+            let (outcome, newCoverage) = self.execute(program, stats: &mutator.stats)
 
-        fuzzer.collectRuntimeTypes(for: prefixProgram)
-        
-        return prefixProgram
+            // If we have new coverage continue mutating the parent as the new program should be in the corpus now.
+            // Moreover, the new program could be empty due to minimization, which would cause problems above.
+            // If we don't have new coverage, we continue mutating this sample.
+            if .succeeded == outcome && !newCoverage {
```
saelo, line 152: This must be a pretty old comment :D I think we can just remove the `&& !newCoverage` bit and just always continue mutating the program if the outcome is `.succeeded`. I think then you can also remove the 2nd return value of `execute`, then it's simpler

Sources/Fuzzilli/Core/MutationEngine.swift
```diff
@@ -142,9 +125,9 @@ public class MutationEngine: ComponentBase, FuzzEngine {
     public func fuzzOne(_ group: DispatchGroup) {
         var parent = prepareForMutation(fuzzer.corpus.randomElement())
         var program = Program()
-        
+
         for _ in 0..<numConsecutiveMutations {
-            var mutator = mutators.randomElement()
+            var mutator = self.fuzzer.mutators.randomElement()
```
saelo, line 130: Nit: here and below, I think you can remove `self`

Sources/Fuzzilli/Core/MutationEngine.swift
```diff
@@ -24,58 +24,41 @@ public enum CrashBehaviour: String {
...
-                self.prefix = self.makePrefix()
+                self.prefix = self.generateProgramPrefix()
             }
         }
-        
+
         if fuzzer.config.logLevel.isAtLeast(.info) {
             fuzzer.timers.scheduleTask(every: 15*Minutes) {
-                let stats = self.mutators.map({ "\($0.name): \(String(format: "%.2f%%", $0.correctnessRate * 100))" }).joined(separator: ", ")
+                let stats = self.fuzzer.mutators.map({ "\($0.name): \(String(format: "%.2f%%", $0.stats.correctnessRate * 100))" }).joined(separator: ", ")
```
saelo, line 56: I think this entire block (lines 54 - 59) should probably be moved into `Fuzzer` since that's where the Mutators are now. That way, with the MultiEngine enabled, these messages won't be printed twice (by the HybridEngine and this one)

Sources/Fuzzilli/Core/CodeTemplates.swift
```diff
@@ -0,0 +1,167 @@
...
+// suited for the HybridFuzzing mode.
+public let CodeTemplates: [CodeTemplate] = [
+    CodeTemplate("JIT1Function") { b in
+        let genSize = 3
+
+        // Generate random function signatures as our helpers
+        var functionSignatures = CodeTemplate.generateRandomFunctionSignatures(forFuzzer: b.fuzzer, n: 2)
+
+        // Generate random property types
+        CodeTemplate.generateRandomPropertyTypes(forBuilder: b)
```
saelo, line 26: I think you should modify the PropertyAssignmentGenerator to take property types into account, i.e. something like
```
...
var propertyType = b.type(ofProperty: propertyName)
// TODO unify the .unknown => .anything conversion
if propertyType == .unknown {
    propertyType = .anything
}
let value = b.randVar(ofType: propertyType)
...
```

Sources/Fuzzilli/Core/HybridEngine.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        let program = generateTemplateProgram(baseTemplate: template)
+
+        let (outcome, _) = execute(program, stats: &template.stats)
+
+        guard outcome == .succeeded else {
+            return
+        }
+
+        for _ in 0..<mutationRounds {
```
saelo, line 72: I think this loop can be removed entirely to reduce the number of "knobs"

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -194,103 +194,107 @@ public class ProgramBuilder {
...
+        interpreter?.setType(of: variable, to: variableType)
+    }
+
     public func setSignature(ofMethod methodName: String, to methodSignature: FunctionSignature) {
         interpreter?.setSignature(ofMethod: methodName, to: methodSignature)
     }
-    
-    public func generateCallArguments(for signature: FunctionSignature) -> [Variable]? {
+
+    private func buildCallArgumentsInternal(for signature: FunctionSignature, block: (_: Type, _: inout [Variable]) -> Bool) -> [Variable]? {
```
saelo, line 294: Hm, maybe we can make this `argumentTypes(for signature: FunctionSignature) -> [Type]`, and just let it produce an array of the types of the arguments, then call it from both `generateCallArguments` and `randCallArguments` and let them actually get/generate the argument values. I think that should simplify things a bit. WDYT?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -299,42 +303,160 @@ public class ProgramBuilder {
...
         return arguments
     }
-    
-    public func generateCallArguments(for function: Variable) -> [Variable]? {
+
+    public func generateCallArguments(for signature: FunctionSignature) -> [Variable] {
+        return buildCallArgumentsInternal(for: signature, block: { param, arguments in
+            // Sometimes we should take the available variables here.
+            // TODO(cffsmith): should we just always generate in this function?
+            if let v = randVar(ofType: param), probability(0.5) {
```
saelo, line 328: I feel like we only want to generate code here if we don't have a match in our current program (otherwise, we again create disjoint dataflows). But I think for that you'll need `randVar(ofConservativeType:)`?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -562,16 +684,21 @@ public class ProgramBuilder {
 
         generator.run(in: self, with: inputs)
     }
-    
+
+    /// Executes a code template.
+    func run(_ template: CodeTemplate) {
```
saelo, line 689: I think we can avoid this additional indirection and just call `template.run(in: b)` directly in the HybridFuzzer.

saelo: And maybe `run` isn't the ideal name, how about `template.generate(in: b)` or something like that?

Sources/Fuzzilli/Core/CodeTemplate.swift
carl-smith: Oh wow actually I wanted to return multiple signatures, the whole loop is missing :D

Sources/Fuzzilli/Core/CodeTemplate.swift
carl-smith: I think `5` is related to `customPropertyNames` if I am not mistaken? i.e. a, b, c, d, e?
We should probably just use `fuzzer.environment.customPropertyNames.count`? and then go through each of the properties and generate a random type instead of `chooseUniform`? WDYT?

saelo: Oh hmm. I guess we'd want at least a few properties to have no types. How about we generate types for half of them?

Sources/Fuzzilli/Core/CodeTemplate.swift
carl-smith: good point!

Sources/Fuzzilli/Core/SpliceEngine.swift
carl-smith: Yes 👍

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -176,7 +176,13 @@ public let CodeGenerators: [CodeGenerator] = [
         } else {
             propertyName = b.genPropertyNameForWrite()
         }
-        let value = b.randVar()
+        var propertyType = b.type(ofProperty: propertyName)
+        // TODO unify the .unknown => .anything conversion
+        if propertyType == .unknown {
+            propertyType = .anything
+        }
+        // As we switch to .anything, we can force unwrap here.
```
saelo, line 184: I don't think this is true since you could have some property type that is not currently present in the generated program, or not?
Since this case is somewhat rare (should only occur in HybridFuzzer in some cases), maybe you can just do `generateVariable` if you get nil here? And make a comment that this is ok since it's guaranteed that property types can be generated by `generateVariable`?

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,105 @@
...
+        var signatures = [FunctionSignature]()
+        for _ in 0..<n {
+            signatures.append(CodeTemplate.generateSignature(forFuzzer: fuzzer, n: Int.random(in: 0...3)))
+        }
+        return signatures
+    }
+
+    /// This function generates and sets property types for the global properties
+    static func generateRandomPropertyTypes(forBuilder b: ProgramBuilder) {
+        for i in 0..<Int(b.fuzzer.environment.customPropertyNames.count/2) {
```
saelo, line 42: Nit: shouldn't the result of the division already be an Int?

Sources/Fuzzilli/Core/FuzzEngine.swift
```diff
@@ -18,3 +18,69 @@ public protocol FuzzEngine: ComponentBase {
...
+            case .timedOut:
+                fuzzer.dispatchEvent(fuzzer.events.TimeOutFound, data: program)
+                stats.producedInvalidSample()
+        }
+
+        return execution.outcome
+    }
+
+    /// Generate some basic Prefix such that samples have some basic types available.
+    public func generateProgramPrefix(mode: ProgramBuilder.Mode = .aggressive) -> Program {
```
saelo, line 62: Do you even need the Mode here since the builder only runs a few trivial CodeGenerators? I guess you can always set it to `.conservative` since it shouldn't make a difference and the prefix should anyway be as correct as possible, since otherwise it would really ruin the MutationEngine's day as it uses it for many mutations in a row

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -176,7 +176,13 @@ public let CodeGenerators: [CodeGenerator] = [
         } else {
             propertyName = b.genPropertyNameForWrite()
         }
-        let value = b.randVar()
+        var propertyType = b.type(ofProperty: propertyName)
```
saelo, line 179: I think we need the same logic in `CreateObject`, since that also defines properties

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -298,42 +306,186 @@ public class ProgramBuilder {
...
+        }
+
+        assert(type.Is(.object()), "Unexpected type encountered \(type)")
+
+        // The variable that we will return.
+        var obj: Variable
+
+        // Fast path for array creation.
+        if type.Is(fuzzer.environment.arrayType) && probability(0.9) {
+            print("took fast path")
```
saelo, line 410: Nit: remove ;)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -298,42 +306,186 @@ public class ProgramBuilder {
...
+            let value = withEqualProbability({
```
saelo, line 411: Maybe you can just do `randVar()` here?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -298,42 +306,186 @@ public class ProgramBuilder {
...
+            return self.createArray(with: Array(repeating: value, count: Int.random(in: 1...5)))
+        }
+
+        if let group = type.group {
+            // We check this during Environment initialization, but let's keep this just in case.
+            assert(self.fuzzer.environment.type(ofBuiltin: group) != .unknown, "We don't know how to construct \(group)")
+            let constructionSignature = self.fuzzer.environment.type(ofBuiltin: group).constructorSignature!
+            let arguments = self.generateCallArguments(for: constructionSignature)
+            let constructor = self.loadBuiltin(group)
+            obj  = self.construct(constructor, withArgs: arguments)
```
saelo, line 427: Nit: double space

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -298,42 +306,186 @@ public class ProgramBuilder {
...
+                for method in type.methods {
+                    // TODO: This should take the method type/signature into account!
+                    let methodVar = self.randVar(ofType: .function())
+                    self.storeProperty(methodVar!, as: method, on: obj)
+                }
+                // These types might have been defined in the interpreter
+                for prop in type.properties {
+                    let type = interpreter?.type(ofProperty: prop)
+                    var value: Variable?
+                    if type != nil && type != .unknown {
```
saelo, line 460: Nit: I think you can write this as `if let type = interpreter?.type(ofProperty: prop), type != .unknown {`

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -298,42 +306,186 @@ public class ProgramBuilder {
...
+                    } else {
+                        value = self.randVar()
+                    }
+                    self.storeProperty(value!, as: prop, on: obj)
+                }
+            }
+
+            // This is needed such that generateCallArguments for later
+            // on finds this variable as the type.
+            self.setType(ofVariable: obj, to: type)
```
saelo, line 474: Is this really needed? Shouldn't the interpreter be able to compute that type?

Sources/Fuzzilli/Core/HybridEngine.swift
```diff
@@ -0,0 +1,78 @@
...
+public class HybridEngine: ComponentBase, FuzzEngine {
+    // The number of mutations to perform to a single sample per round
+    private let numConsecutiveMutations: Int
+
+    public init(numConsecutiveMutations: Int) {
+        self.numConsecutiveMutations = numConsecutiveMutations
+        super.init(name: "HybridEngine")
+    }
+
+    override func initialize() {
```
saelo, line 26: Let's add an `assert(fuzzer.config.useAbstractInterpretation, "The HybridEngine requires abstract interpretation to be enabled")` because I think otherwise, this will just completely fail to do anything useful :D

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Actually, can't you use `ProgramBuilder.type(ofProperty:)` here?

Sources/Fuzzilli/Core/CodeGenerators.swift
carl-smith: Ah I think you're right :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: good catch! :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Well, what's actually interesting is that I hit the assertion [here](https://github.com/carl-smith/fuzzilli-1/blob/HybridEngine/Sources/Fuzzilli/Core/ProgramBuilder.swift#L336) even after explicitly setting the type here. Not sure why that is? I think this now happens after switching from `>=` to `>` [here](https://github.com/carl-smith/fuzzilli-1/blob/HybridEngine/Sources/Fuzzilli/Core/CodeTemplate.swift#L79) but I am not entirely sure.

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: ah cool, was looking for something like that! :)
yeah I added the `ProgramBuilder.type(ofProperty:)` function after changing this I think?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -298,42 +306,186 @@ public class ProgramBuilder {
...
+            }, {
+                self.createObject(with: [:])
+            })
+            return self.createArray(with: Array(repeating: value, count: Int.random(in: 1...5)))
+        }
+
+        if let group = type.group {
+            // We check this during Environment initialization, but let's keep this just in case.
+            assert(self.fuzzer.environment.type(ofBuiltin: group) != .unknown, "We don't know how to construct \(group)")
+            let constructionSignature = self.fuzzer.environment.type(ofBuiltin: group).constructorSignature!
```
saelo, line 424: Nit: you can remove `self` from all of these calls I think

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Hmm, maybe add some logging to print the type that should've been generated

Sources/Fuzzilli/Core/FuzzEngine.swift
carl-smith: Yes I agree :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
carl-smith: Really not sure what the issue here is, I will push my recent changes and then look at this.

carl-smith: ah I actually have not changed from `>=` to `>` yet.

carl-smith: Maybe objects with a `group` have custom properties and if `generateVariable` does not set those as well, `randVar` fails to find them?

saelo: Maybe, I guess you'll see that if you print the type in the assert? :)

Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -172,6 +173,8 @@ public class JavaScriptEnvironment: ComponentBase, Environment {
         }
         
         customPropertyNames = ["a", "b", "c", "d", "e"]
+        // This is needed in ProgramBuilder.generateVariable, such that method names don't collide with property names.
```
saelo, line 176: I think we can just drop this comments, to me it seems intuitive that if we have `customPropertyNames`, then we should also have `customMethodNames`

Sources/Fuzzilli/Util/MockFuzzer.swift
```diff
@@ -54,6 +54,7 @@ class MockEnvironment: ComponentBase, Environment {
     var readPropertyNames = Set(["foo", "bar"])
     var writePropertyNames = Set(["foo", "bar"])
     var customPropertyNames = Set(["foo", "bar"])
+    var customMethodNames = Set(["foo", "bar"])
```
saelo, line 57: Nit: Doesn't this violate the invariant that these two sets should be disjoint? Maybe just call it `m1`, `m2` or so?

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,114 @@
...
+                var methods: [String] = []
+
+                // TODO: add group with probability here as well?
+
+                // Generate random properties.
+                // We filter the candidates to avoid cycles in our objects.
+                // TODO: we should remove this "no-cycle" restriction here, and let `generateVariable`
+                // handle these cases.
+                for _ in 1..<3 {
+                    let candidates = fuzzer.environment.customPropertyNames.filter({ $0 >= property })
```
saelo, line 88: Shouldn't this be `>` instead of `>=`?
Otherwise I think you can end up with
`.b = Type.object(withProperties: ["b"])` 
and I think then generateVariable would recurse until it crashes.

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,114 @@
...
+                    properties.append(chooseUniform(from: candidates))
```
saelo, line 89: I think here we should just check if `candidates.isEmpty`, since I think this case will happen during normal operations

Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -0,0 +1,114 @@
...
+            signatures.append(CodeTemplate.generateSignature(forFuzzer: fuzzer, n: Int.random(in: 0...3)))
+        }
+        return signatures
+    }
+
+    /// This function generates and sets property types for the global properties
+    static func generateRandomPropertyTypes(forBuilder b: ProgramBuilder) {
+        // generate types for half of the available property names.
+        for i in 0..<b.fuzzer.environment.customPropertyNames.count/2 {
+            // if we pass "e" (last customPropertyName) into generateType the filter might fail and we
```
saelo, line 44: I think `generateType` should just be able to handle the case that the filter returns an empty array. Then this workaround isn't necessary.
IIUC, the current approach for preventing cycles works by ensuring that a property X can only be an object with properties whose names are all > X. Then it's guaranteed that you will never get a cycle. However, that implies that the property "e" (the one with the "highest" value) cannot be an object with any other properties.

PR #147, opened 2020-09-16
Cache program types

samo98:
We should save infered types of program variables to avoid executing it by `AbstractInterpreter` more than once

* Attach all known types (not only runtime ones) to `Program` class
* Add `TypeQuality` to `ProgramTypes` struct
* `AbstractInterpreter` now returns type changes after execution
* Recollect static types after runtime types collection (using runtime types as hints)

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -22,7 +22,7 @@ public class ProgramBuilder {
     
     /// The code and type information of the program that is being constructed.
     private var code = Code()
-    public var runtimeTypes = ProgramTypes()
+    public var types = ProgramTypes()
     
     public enum Mode {
         /// In this mode, the builder will try as hard as possible to generate semantically valid code.
@@ -71,7 +71,7 @@ public class ProgramBuilder {
         seenPropertyNames.removeAll()
         seenIntegers.removeAll()
         code.removeAll()
-        runtimeTypes = ProgramTypes()
+        types = ProgramTypes()
         scopeAnalyzer = ScopeAnalyzer()
         contextAnalyzer = ContextAnalyzer()
         interpreter?.reset()
@@ -80,8 +80,7 @@ public class ProgramBuilder {
     
     /// Finalizes and returns the constructed program, then resets this builder so it can be reused for building another program.
     public func finalize() -> Program {
-        let program = Program(with: code)
-        program.runtimeTypes = runtimeTypes
+        let program = Program(code: code, types: types)
         // TODO set type status to something meaningful?
         reset()
         return program
@@ -272,7 +271,7 @@ public class ProgramBuilder {
     
     /// Type information access.
     public func type(of v: Variable) -> Type {
-        return interpreter?.type(of: v) ?? .unknown
+        return types.getType(of: v, at: code.lastInstruction.index) 
     }
     
     public func methodSignature(of methodName: String, on object: Variable) -> FunctionSignature {
@@ -333,7 +332,7 @@ public class ProgramBuilder {
     ///
     private var varMaps = [VariableMap<Variable>]()
     /// Formatted ProgramTypes structure for easier adopting of runtimeTypes
-    private var typeMaps = [[[(Variable, Type)]]]()
+    private var runtimeTypesMaps = [[[(Variable, Type)]]]()
     
     /// Prepare for adoption of variables from the given program.
     ///
@@ -341,24 +340,13 @@ public class ProgramBuilder {
     /// currently constructed one to avoid collision of variable names.
     public func beginAdoption(from program: Program) {
         varMaps.append(VariableMap())
-        typeMaps.append(formatTypeMap(from: program))
-    }
-
-    // Format ProgramTypes struct so searching for type changes at instruction is easier
-    private func formatTypeMap(from program: Program) -> [[(Variable, Type)]] {
-        var typeMap: [[(Variable, Type)]] = Array(repeating: [], count: program.size)
-        for (variable, typeData) in program.runtimeTypes {
-            for typeIndexPair in typeData {
-                typeMap[typeIndexPair.index].append((variable, typeIndexPair.type))
-            }
-        }
-        return typeMap
+        runtimeTypesMaps.append(program.types.onlyRuntimeTypes().indexedByInstruction(for: program))
     }
     
     /// Finishes the most recently started adoption.
     public func endAdoption() {
         varMaps.removeLast()
-        typeMaps.removeLast()
+        runtimeTypesMaps.removeLast()
     }
     
     /// Executes the given block after preparing for adoption from the provided program.
@@ -388,11 +376,13 @@ public class ProgramBuilder {
     }
 
     private func adoptTypes(at origInstrIndex: Int) {
-        for (variable, type) in typeMaps.last![origInstrIndex] {
+        for (variable, type) in runtimeTypesMaps.last![origInstrIndex] {
             // No need to keep unknown type nor type of not adopted variable
             if type != .unknown, let adoptedVariable = varMaps.last![variable] {
-                runtimeTypes.setType(of: adoptedVariable, to: type, at: code.lastInstruction.index)
                 interpreter?.setType(of: adoptedVariable, to: type)
+                // We should save this type even if we do not have interpreter
+                // This way we can use runtime types without interpreter
+                types.setType(of: adoptedVariable, to: type, at: code.lastInstruction.index, quality: .runtime)
             }
         }
     }
@@ -440,13 +430,6 @@ public class ProgramBuilder {
             return Variable(number: nextFreeVariable - 1)
         }
 
-        // Get types of the other program
-        var ai = AbstractInterpreter(for: self.fuzzer.environment)
-
-        for instr in source {
-            ai.execute(instr)
-        }
-
         // We still adopt from the input program, just with slightly modified code :)
         beginAdoption(from: program)
 
@@ -467,7 +450,7 @@ public class ProgramBuilder {
             for (idx, input) in instr.inputs.enumerated() {
                 neededInputs.append(input)
                 if probability(0.2) && mode != .conservative {
-                    var type = ai.type(of: input)
+                    var type = program.types.getType(of: input, at: instr.index)
                     if type == .unknown {
                         type = .anything
                     }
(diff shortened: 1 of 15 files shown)


Sources/Fuzzilli/FuzzIL/TypeInfo.swift
```diff
@@ -0,0 +1,67 @@
...
+            case 0: self = .infered
+            case 1: self = .runtime
+            default: self = .infered
+        }
+    }
+}
+
+// Store known types for program variables at specific instructions
+public struct TypeInfo: Equatable {
+    public let index: Int
```
saelo, line 36: Since we'll probably have quite a lot of these structs in memory, let's make it
```
public let index: UInt16 // Guaranteed to fit into a UInt16, see Instruction.swift
public let quality: TypeQuality // rawValue is UInt16 or UInt8
public let type: Type
```

Tests/FuzzilliTests/InliningTest.swift
```diff
@@ -63,6 +63,9 @@ class InliningTests: XCTestCase {
 
         let referenceProgram = b.finalize()
 
+        // Delete type information to compare only code
+        referenceProgram.types = ProgramTypes()
```
saelo, line 67: How about doing `XCTAssertEqual(inlinedProgram.code, referenceProgram.code)` and defining the comparison for `Code` in the same way as for the program (i.e. via the protobuf?). Would that work? That seems like what this test really wants to accomplish

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -200,36 +220,56 @@ public struct AbstractInterpreter {
...
-    private mutating func executeEffects(_ instr: Instruction) {
+
+    private mutating func executeEffects(_ instr: Instruction) -> [(Variable, Type)] {
+        var typeChanges = [(Variable, Type)]()
+
         switch instr.op {
             
         case let op as LoadBuiltin:
-            set(instr.output, environment.type(ofBuiltin: op.builtinName))
+            typeChanges.append(
```
saelo, line 230: This feels a bit bloated, maybe define a local helper function to reduce all of these back to a single function call/a single line:
```
func set(_ v: Variable, to type; Type) {
    setType(of: v, to: Type)
    typeChanges.append((v, t))
}
```
Then you can also remove the somewhat unintuitive return value of `setType`

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -200,36 +220,56 @@ public struct AbstractInterpreter {
         }
         return result
     }
-    
-    private mutating func executeEffects(_ instr: Instruction) {
+
+    private mutating func executeEffects(_ instr: Instruction) -> [(Variable, Type)] {
```
saelo, line 224: You could let this function take the `typeChanges` array as `inout` parameter since it will anyway be concatenated to it

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -40,54 +40,78 @@ public struct AbstractInterpreter {
         propertyTypes.removeAll()
         methodSignatures.removeAll()
     }
-    
+
+    /// Get variables with new type which have different type in old state
+    /// Used after block end when some state on stack is removed
+    private func getStateChanges(oldState: VariableMap<Type>, newState: VariableMap<Type>) -> [(Variable, Type)] {
```
saelo, line 46: Since the interpreter now has to answer the `type(of)` query less often (the function is private now), I think it would be quite efficient to store the `stack` as diffs instead of having all variables be present in every stack element. This will mostly (significantly) speed up the `merge` operation, but should also help here, since the changes are immediately obvious if the state is just a diff. Only the `type(of)` function will now have to traverse the stack to find the current type. WDYT?

saelo: This can also happen as a separate commit though, and I can also do it

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -73,7 +69,63 @@ public struct ProgramTypes: Equatable, Sequence {
...
+                runtimeTypes.setType(
+                    of: variable, to: typeInfo.type, at: typeInfo.index, quality: .runtime
+                )
+            }
+        }
+        return runtimeTypes
+    }
+
+    // Format ProgramTypes struct so searching for type changes at instruction is easier
+    public func getInvertedStructure(for program: Program) -> [[(Variable, Type)]] {
```
saelo, line 88: Maybe name it `indexedByInstruction` or so, since then it's a bit clearer from the name what this actually does?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -370,6 +370,29 @@ public class Fuzzer {
         return execution
     }
 
+    private func recollectStaticTypes(for program: Program) {
```
saelo, line 373: Maybe call it `inferMissingTypes` or something like that? I think "collect" is a bit confusing here

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -73,7 +69,63 @@ public struct ProgramTypes: Equatable, Sequence {
         return types[variable]?.last(where: { $0.index <= instrIndex })?.type ?? .unknown
     }
 
-    public func makeIterator() -> VariableMap<[TypeIndexPair]>.Iterator {
+    // Filter out only runtime types
+    public func getRuntimeTypes() -> ProgramTypes {
```
saelo, line 73: Nit: maybe `onlyRuntimeTypes` or `removingInferredTypes`/

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -508,7 +499,7 @@ public class JavaScriptLifter: Lifter {
                 } else {
                     w.emit("\(decl(v)) = \(expression);")
                     if options.contains(.dumpTypes) {
-                        w.emitComment("\(v) = \(interpreter.type(of: v))")
+                        w.emitComment("\(v) = \(program.types.getType(of: v, at: instr.index))")
```
saelo, line 502: I think it would be nice to have a convenience accessor: `Program.type(of: Variable, at: Int)`, wdyt?

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -73,7 +69,63 @@ public struct ProgramTypes: Equatable, Sequence {
...
+
+extension ProgramTypes: ProtobufConvertible {
+    public typealias ProtobufType = [UInt32: Fuzzilli_Protobuf_InstrTypes]
+
+    func asProtobuf() -> ProtobufType {
+        var protobuf = ProtobufType()
+        for (variable, instrTypes) in self {
+            protobuf[UInt32(variable.number)] = Fuzzilli_Protobuf_InstrTypes()
+            for typeInfo in instrTypes {
+                protobuf[UInt32(variable.number)]!.typeInfo.append(typeInfo.asProtobuf())
```
saelo, line 111: Nit: how about you keep the `Fuzzilli_Protobuf_InstrTypes()` in a local variable at first, append all of these, then attach it to `protobuf[UInt32(variable.number)]`?

Tests/FuzzilliTests/InliningTest.swift
samo98: yeah well we have to define how to convert `Code` to protobuf and define equality opertor

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: yeah as we talked before I thought it will be done separately as this is large PR itself

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
samo98: yeah and it can do some additional work as we talked before. Like checking if variable is existent at requested instruction and return `.nothing` if it is not (instead of .unknown).

But still I would like to do it separately, if not PR at least separate commit.

saelo: Cool, yeah either is fine for me, whatever you prefer :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -333,32 +332,22 @@ public class ProgramBuilder {
...
-    // Format ProgramTypes struct so searching for type changes at instruction is easier
-    private func formatTypeMap(from program: Program) -> [[(Variable, Type)]] {
-        var typeMap: [[(Variable, Type)]] = Array(repeating: [], count: program.size)
-        for (variable, typeData) in program.runtimeTypes {
-            for typeIndexPair in typeData {
-                typeMap[typeIndexPair.index].append((variable, typeIndexPair.type))
-            }
-        }
-        return typeMap
+        // Append inverted ProgramTypes for easier search for type changes at instruction
```
saelo, line 343: Nit: this comment is outdated, but I think it can be removed entirely now with the new method names

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -908,8 +908,16 @@ extension Type: ProtobufConvertible {
             $0.definiteType = definiteType.rawValue
             $0.possibleType = possibleType.rawValue
             if let typeExtension = ext {
+                // We have to keep these arrays sorted to simplify comparision
                 $0.properties = Array(typeExtension.properties)
                 $0.methods = Array(typeExtension.methods)
+                // Unit tests currently rely on comparing these protobufs, which only
```
saelo, line 914: Is this still true now that we've changed the `==` operator for Programs?

Sources/Fuzzilli/Protobuf/program.proto
```diff
@@ -111,20 +111,32 @@ message Instruction {
...
     ERROR = 1;
     TIMEOUT = 2;
     NOTATTEMPTED = 3;
 }
 
-message TypeMap {
-  map<uint32, Type> typeMap = 1;
+// Keep in sync with ProgramTypes.swift
+enum TypeQuality {
+    INFERED = 0;
```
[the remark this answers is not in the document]
saelo: I think this one is still misspelled :P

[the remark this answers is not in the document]
samo98: weird 🤔 I remember fixing it & generated code is already fixed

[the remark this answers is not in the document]
saelo: yeah I noticed that as well :D

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -467,7 +450,7 @@ public class ProgramBuilder {
             for (idx, input) in instr.inputs.enumerated() {
                 neededInputs.append(input)
                 if probability(0.2) && mode != .conservative {
-                    var type = ai.type(of: input)
+                    var type = program.types.getType(of: input, at: idx)
```
saelo, line 453: I think this should be `instr.index`, since `idx` is the index of the `input`

PR #149, opened 2020-09-17
When processing intereseting program, infer missing types when needed

samo98:
Now after minimization program lose its type info and we recollect it only if runtime types collection is enabled.

This PR changes it in a way, that we infer missing types everytime we have new runtime type hints or program does not have any types

### DIFF
--- a/Sources/Fuzzilli/Fuzzer.swift
+++ b/Sources/Fuzzilli/Fuzzer.swift
@@ -371,8 +371,6 @@ public class Fuzzer {
     }
 
     private func inferMissingTypes(in program: Program) {
-        guard config.useAbstractInterpretation else { return }
-
         var ai = AbstractInterpreter(for: self.environment)
         let runtimeTypes = program.types.onlyRuntimeTypes().indexedByInstruction(for: program)
         var types = ProgramTypes()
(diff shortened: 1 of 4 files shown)


Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -424,17 +421,26 @@ public class Fuzzer {
...
 
-        // Recollect types using AbstractInterpreter as we have new precise runtime type hints
-        inferMissingTypes(in: program)
+    func computeTypeInformation(for program: Program, minimized: Bool) {
+        if config.collectRuntimeTypes {
+            collectRuntimeTypes(for: program)
+        }
+        // New static types are needed either if program was minimized and lost its type info
+        // or we collected runtime types and statical inference can be improved now
+        let newTypesNeeded = minimized || config.collectRuntimeTypes
```
saelo, line 432: Could we make a `Program.typeInformationAvailable` property/method or a `program.types.isEmpty` or so, and then replace `minimized` with that in this condition? I think that would be a bit more intuitive, since I guess there could be other reasons in the future why a program doesn't have type information.

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -424,17 +421,26 @@ public class Fuzzer {
...
+        if config.useAbstractInterpretation, newTypesNeeded {
```
saelo, line 433: Nit: maybe use `&&` instead of `,` to join the conditions

PR #150, opened 2020-09-17
Add type API in program

samo98:
* Rename `at` instruction to `after` instruction
* Add getType to `Program` class
* Forbid runtime types collection mode without abstractInterpreter
* Enforce failure on querying type of variable before definition

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -271,7 +271,7 @@ public class ProgramBuilder {
     
     /// Type information access.
     public func type(of v: Variable) -> Type {
-        return types.getType(of: v, at: code.lastInstruction.index) 
+        return types.getType(of: v, after: code.lastInstruction.index)
     }
     
     public func methodSignature(of methodName: String, on object: Variable) -> FunctionSignature {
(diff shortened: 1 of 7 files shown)


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -450,7 +450,7 @@ public class ProgramBuilder {
             for (idx, input) in instr.inputs.enumerated() {
                 neededInputs.append(input)
                 if probability(0.2) && mode != .conservative {
-                    var type = program.types.getType(of: input, at: instr.index)
+                    var type = program.type(of: input, after: instr.index)
```
saelo, line 453: I think here we actually need `program.type(of: input, before: instr.index)` because we are changing the input of `instr`, so we need the types before that instruction executed. Maybe wen can just have `Program.type(of:, before:)` which simply calls the other function with `index - 1`?

Sources/Fuzzilli/FuzzIL/ProgramTypes.swift
```diff
@@ -64,9 +64,14 @@ public struct ProgramTypes: Equatable, Sequence {
         }
     }
 
-    // Get type of variable at current instruction
-    public func getType(of variable: Variable, at instrIndex: Int) -> Type {
-        return types[variable]?.last(where: { $0.index <= instrIndex })?.type ?? .unknown
+    // Get type of variable after given instruction
+    public func getType(of variable: Variable, after instrIndex: Int) -> Type {
+        if let variableTypes = types[variable] {
+            // There should always be type info after variable definition instruction
```
saelo, line 70: Nit: in case we ever run into this, the crash message will not be very nice (i.e. "unwrapped nil"). How about a
```
assert(variableTypes[0].index <= instrIndex, "Queried type of variable before its definition")
```
?

PR #152, opened 2020-09-18
add JIT2Functions template and add log message

carl-smith:


Sources/Fuzzilli/Core/CodeTemplates.swift
```diff
@@ -139,5 +139,84 @@ public let CodeTemplates: [CodeTemplate] = [
...
+        b.forLoop(b.loadInt(0), .lessThan, b.loadInt(100), .Add, b.loadInt(1)) { args in
+            let arguments = b.generateCallArguments(for: signature2)
+            b.callFunction(f2, withArgs: arguments)
+        }
+
+        // more random instructions
+        b.generate(n: genSize)
+
+        let argumentsBeforeJit2 = b.generateCallArguments(for: signature2)
+        b.callFunction(f2, withArgs: argumentsBeforeJit2)
```
saelo, line 198: Nit: maybe you can just make it `b.callFunction(f2, withArgs: b.generateCallArguments(for: signature2))` here and below instead of creating the temporary variables?

PR #153, opened 2020-09-18
Do not save unknown runtime types

samo98:


Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -425,7 +425,9 @@ public class Fuzzer {
                     for i in stride(from: lineNumber, to: lineNumber + 2 * instrCount, by: 2) {
                         let proto = try Fuzzilli_Protobuf_Type(jsonUTF8Data: lines[i+1].data(using: .utf8)!)
                         let runtimeType = try Type(from: proto)
-                        program.types.setType(of: variable, to: runtimeType, after: Int(lines[i])!, quality: .runtime)
+                        if runtimeType != .unknown {
```
saelo, line 428: Maybe add a small comment here explaining in which case we'd get `.unknown` here?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -553,7 +553,7 @@ public class ProgramBuilder {
     private func adoptTypes(at origInstrIndex: Int) {
         for (variable, type) in runtimeTypesMaps.last![origInstrIndex] {
             // No need to keep unknown type nor type of not adopted variable
-            if type != .unknown, let adoptedVariable = varMaps.last![variable] {
+            if let adoptedVariable = varMaps.last![variable] {
```
saelo, line 556: Since the test is removed here, can we add an `assert(type != .unknown)` here?

PR #155, opened 2020-09-21
AbstractInterpreter - save only state diffs

samo98:
Currently AbstractInterpreter copies all types when creating new internal state, which is quite inefficient.
This PR tries to remember only diffs of states (i.e. types that changed) + current state.

* stateChangesStack - each entry remembers list of sibling states - each of them remembers last known type in that block

* currentState - all current variable types for fast type queries

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -1170,15 +1170,22 @@ public class ProgramBuilder {
 
         currentCodegenBudget -= 1
 
-        // Update type information
-        let typeChanges = interpreter?.execute(instr) ?? []
-        for (variable, type) in typeChanges {
-            types.setType(of: variable, to: type, after: code.lastInstruction.index, quality: .inferred)
-        }
-
         // Update our analyses
         scopeAnalyzer.analyze(instr)
         contextAnalyzer.analyze(instr)
+
+        // Update type information
+        let typeChanges = interpreter?.execute(instr) ?? []
+        for (variable, type) in typeChanges {
+            assert(scopeAnalyzer.visibleVariables.contains(variable))
+            // We should record only changes when type really changes
+            // But we cannot distinguish following changes because .unknown is default type:
+            // 1. nil -> .unknown
+            // 2. .unknwon -> .unknown
+            assert(type != types.getType(of: variable, after: code.lastInstruction.index) || type == .unknown)
+            types.setType(of: variable, to: type, after: code.lastInstruction.index, quality: .inferred)
+        }
+
         updateConstantPool(instr.op)
     }
 
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -11,17 +11,139 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+private struct InterpreterState {
+    // Save variable changes, supports
+    // 1. Nested states (function states, loop states)
+    // 2. Sibling states (if/else branches)
+    // We should start with 2 states, so every used state has parent state
+    private var stateChangesStack = [[VariableMap<Type>()], [VariableMap<Type>()]]
+    
+    // The currently active state.
+    private var currentState = VariableMap<Type>()
+
+    private var activeStateChanges: VariableMap<Type> {
+        get {
+            return stateChangesStack.last!.last!
+        }
+        set {
+            let idx = stateChangesStack[stateChangesStack.count - 1].count - 1
+            stateChangesStack[stateChangesStack.count - 1][idx] = newValue
+        }
+    }
+
+    private var parentStateChanges: VariableMap<Type> {
+        get {
+            return stateChangesStack[stateChangesStack.count - 2].last!
+        }
+        set {
+            let idx = stateChangesStack[stateChangesStack.count - 2].count - 1
+            stateChangesStack[stateChangesStack.count - 2][idx] = newValue
+        }
+    }
+
+    public mutating func reset() {
+        stateChangesStack = [[VariableMap()], [VariableMap()]]
+        currentState = VariableMap()
+    }
+
+    /// Update current variable types and type changes
+    /// Used after block end when some states should be merged to parent
+    public mutating func mergeStates(typeChanges: inout [(Variable, Type)]) {
+        let states = stateChangesStack.removeLast()
+        var numUpdatesPerVariable = VariableMap<Int>()
+        var newTypes = VariableMap<Type>()
+
+        for state in states {
+            for (v, t) in state {
+                // Skip variable types that are out of scope as we do not care about them anymore
+                guard t != .nothing else { continue }
+
+                if newTypes[v] == nil {
+                    newTypes[v] = t
+                    numUpdatesPerVariable[v] = 1
+                } else {
+                    newTypes[v]! |= t
+                    numUpdatesPerVariable[v]! += 1
+                }
+            }
+        }
+
+        for (v, c) in numUpdatesPerVariable {
+            // Skip updating local variables, which go out of scope
+            if activeStateChanges[v] == .nothing {
+                newTypes.remove(v)
+                continue
+            }
+
+            // Not all paths updates this variable, so it must be unioned with the previous type
+            if c != states.count {
+                newTypes[v]! |= activeStateChanges[v]!
+            }
+        }
+
+        // Update currentState and add typeChanges
+        for (v, newType) in newTypes {
+            if currentState[v] != newType {
+                typeChanges.append((v, newType))
+            }
+
+            // Propagate activeStateChanges to parentStateChanges if necessary
+            // We need to keep invariant -> if there is type in active state then there is type in parent
+            if parentStateChanges[v] == nil {
+                parentStateChanges[v] = activeStateChanges[v]!
+            }
+            // Update activeStateChange and currentState
+            activeStateChanges[v] = newType
+            currentState[v] = newType
+        }
+    }
+
+    /// Return the type of the given variable.
+    /// Return .unknown for variables not available in this state.
+    public func type(of variable: Variable) -> Type {
+        return currentState[variable] ?? .unknown
+    }
+
+    public func hasType(variable: Variable) -> Bool {
+        return currentState[variable] != nil
+    }
+
+    /// Set the type of the given variable in the current state.
+    public mutating func setType(of v: Variable, to t: Type) {
+        // Save type in parent state if it is not already
+        if parentStateChanges[v] == nil {
+            parentStateChanges[v] = currentState[v] ?? .nothing
+        }
+        activeStateChanges[v] = t
+        // Update currentState
+        currentState[v] = t
+    }
+
+    public mutating func pushChildState() {
+        stateChangesStack.append([VariableMap()])
+    }
+
+    public mutating func pushSiblingState(typeChanges: inout [(Variable, Type)]) {
+        // Reset last sibling state
+        for (v, t) in activeStateChanges {
+            // Do not save type change if
+            // 1. Variable does not exist in sibling scope (t == .nothing)
+            // 2. Variable is only local in sibling state (parent == .nothing)
+            // 3. No type change happened
+            if t != .nothing && parentStateChanges[v] != .nothing && parentStateChanges[v] != currentState[v] {
+                typeChanges.append((v, parentStateChanges[v]!))
+                currentState[v] = parentStateChanges[v]!
+            }
+        }
+        // Create sibling state
+        stateChangesStack[stateChangesStack.count - 1].append(VariableMap())
+    }
+}
 
 /// Analyzes the types of variables.
 public struct AbstractInterpreter {
-    // States are kept in a stack to support conditional execution.
-    private var ifStack: [VariableMap<Type>] = []
-    private var stack = [VariableMap<Type>()]
-    
-    // The currently active state.
-    private var currentState: VariableMap<Type> {
-        return stack.last!
-    }
+    // The current state
+    private var state = InterpreterState()
     
     // Program-wide property and method types.
     private var propertyTypes = [String: Type]()
@@ -35,83 +157,36 @@ public struct AbstractInterpreter {
     }
     
     public mutating func reset() {
-        ifStack.removeAll()
-        stack = [VariableMap<Type>()]
+        state.reset()
         propertyTypes.removeAll()
         methodSignatures.removeAll()
     }
 
-    /// Get variables with new type which have different type in old state
-    /// Used after block end when some state on stack is removed
-    private func getStateChanges(oldState: VariableMap<Type>, newState: VariableMap<Type>) -> [(Variable, Type)] {
-        var typeChanges = [(Variable, Type)]()
-        for (variable, type) in newState {
-            let oldType = oldState[variable]
-            if oldType == nil || type != oldType  {
-                typeChanges.append((variable, type))
-            }
-        }
-        return typeChanges
-    }
+    // Array for collecting type changes during instruction execution
+    private var typeChanges = [(Variable, Type)]()
 
     /// Abstractly execute the given instruction, thus updating type information.
     /// Return type changes.
     public mutating func execute(_ instr: Instruction) -> [(Variable, Type)] {
-        var typeChanges = [(Variable, Type)]()
+        // Reset type changes array before instruction execution
+        typeChanges = []
+
+        executeOuterEffects(instr)
+
         switch instr.op {
-        case is BeginAnyFunctionDefinition:
-            stack.append(currentState)
-        case is EndAnyFunctionDefinition:
-            let functionState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(functionState, previousState))
-            typeChanges += getStateChanges(oldState: functionState, newState: stack.last!)
         case is BeginIf:
-            stack.append(currentState)
+            state.pushChildState()
         case is BeginElse:
-            let ifState = stack.removeLast()
-            ifStack.append(ifState)
-            typeChanges += getStateChanges(oldState: ifState, newState: stack.last!)
+            state.pushSiblingState(typeChanges: &typeChanges)
         case is EndIf:
-            let ifState = ifStack.removeLast()
-            let elseState = stack.removeLast()
-            stack.append(merge(ifState, elseState))
-            typeChanges += getStateChanges(oldState: elseState, newState: stack.last!)
-        case is BeginWhile:
-            stack.append(currentState)
-        case is EndWhile:
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
-        case is BeginDoWhile:
-            stack.append(currentState)
-        case is EndDoWhile:
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
-        case is BeginFor:
-            stack.append(currentState)
-        case is EndFor:
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
-        case is BeginForIn:
-            stack.append(currentState)
-        case is EndForIn:
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
-        case is BeginForOf:
-            stack.append(currentState)
-        case is EndForOf:
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
+            state.mergeStates(typeChanges: &typeChanges)
+        case is BeginWhile, is BeginDoWhile, is BeginFor, is BeginForIn, is BeginForOf, is BeginAnyFunctionDefinition, is BeginCodeString:
+            // Push empty state representing case when loop/function is not executed at all
+            state.pushChildState()
+            // Push state representing types during loop
+            state.pushSiblingState(typeChanges: &typeChanges)
+        case is EndWhile, is EndDoWhile, is EndFor, is EndForIn, is EndForOf, is EndAnyFunctionDefinition, is EndCodeString:
+            state.mergeStates(typeChanges: &typeChanges)
         case is BeginTry:
             // Ignored for now, TODO
             break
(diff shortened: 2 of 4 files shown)


Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -36,22 +36,47 @@ public struct AbstractInterpreter {
...
+        lastStateDiff: VariableMap<Type>, typeChanges: inout [(Variable, Type)], mergeTypes: Bool = true
+    ) {
+        for (variable, oldType) in lastStateDiff {
+            var newType = oldType
+
+            if mergeTypes {
+                // For now, if variable did not exist in parent state we keep its current type
+                // In future we can get rid of these types when we update types in correct time
+                // i.e. innerOutputs after updating state and outputs before updating state
+                if oldType == .nothing {
```
samo98, line 56: As we discussed we can either follow comment and get rid of this case
or we can use analyzer which hides ceratin types when variable is out of scope

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -36,22 +36,48 @@ public struct AbstractInterpreter {
...
+                // So we won't record any new type changes
+                if oldType == .nothing {
+                    newType = currentState[variable]!
+                } else {
+                    newType |= currentState[variable]!
+                }
+            }
+
+            if currentState[variable] != newType, newType != .nothing {
+                currentState[variable] = newType
```
saelo, line 63: If `oldType == .nothing`, couldn't we remove the variable from the `currentState`?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -60,58 +86,26 @@ public struct AbstractInterpreter {
...
         case is BeginAnyFunctionDefinition:
-            stack.append(currentState)
+            stateDiffsStack.append(VariableMap())
         case is EndAnyFunctionDefinition:
-            let functionState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(functionState, previousState))
-            typeChanges += getStateChanges(oldState: functionState, newState: stack.last!)
+            let functionStateDiff = stateDiffsStack.removeLast()
+            updateStateDiffs(lastStateDiff: functionStateDiff, typeChanges: &typeChanges)
```
saelo, line 92: And this can probably be called `mergeStates` or so, and there could also be a `resetState(to:)` helper function maybe?
I guess overall, it would be nice if the implementation of the state tracking as diffs and the complete `currentState` was somewhat invisible in this code, so that this would just read as if ever entry on the stack was complete. I think that should be possible even with the diffs. The way I see it, the diffs are just an optimization. Every entry on the stack still represents a complete execution state, just its storage is optimized to only store what has changed. WDYT?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -14,14 +14,14 @@
 
 /// Analyzes the types of variables.
 public struct AbstractInterpreter {
-    // States are kept in a stack to support conditional execution.
+    // State diffs are kept in a stacks to support conditional execution.
+    // Save final variable types changed in if block.
     private var ifStack: [VariableMap<Type>] = []
-    private var stack = [VariableMap<Type>()]
+    // Save last variable type before changing it in current block.
```
saelo, line 20: I think here you could explain in a bit more detail how the diffs work, i.e. that they are essentially diffs from `currentState` to be able to reset/merge with a previous state.

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -36,22 +36,48 @@ public struct AbstractInterpreter {
...
+    private mutating func updateStateDiffs(
+        lastStateDiff: VariableMap<Type>, typeChanges: inout [(Variable, Type)], mergeTypes: Bool = true
+    ) {
+        for (variable, oldType) in lastStateDiff {
+            var newType = oldType
+
+            if mergeTypes {
+                // If variable did not exist in parent state we keep its current type
+                // So we won't record any new type changes
+                if oldType == .nothing {
```
saelo, line 55: I think it would be good to explain your trick with `.nothing` expressing that a variable wasn't defined in the outer scope somewhere. I guess either here or in `setType` where you set it

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -35,102 +35,95 @@ public struct AbstractInterpreter {
...
+        for (variable, oldType) in lastStateDiff {
+            var newType = oldType
+
+            switch mergeStrategy {
+            case .union:
+                // If oldType is nothing variable did not exist before, so after merging these states
+                // it won't exit anymore.
+                // If currently variable does not exist but in old block existed
+                // variable will still exist, so we should keep oldType
+                if oldType != .nothing, currentState[variable]! != .nothing {
```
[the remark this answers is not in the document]
samo98: hmm this is probably some old version diff

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: not really, e.g.
```
BeginIf v6
    v28 <- BinaryOperation v2, '/', v9
    // v28 = .integer | .float | .bigint
BeginElse
EndIf
// v28 = .integer | .float | .bigint
```
It would result in additional typechange.

saelo: So in that example, we'd only want to have a single entry (just after the `BinaryOperation `) for `v28` in `Program.types` after interpretation is finished, correct?
So is the problem with `If-Else` that at the time you merge the two branches, you don't know which variables where defined inside the blocks and which ones are visible outside (because you don't merge with the "parent" state but only with a "sibling" state)? Maybe you need to do something like:
```
case BeginElse:
    let lastState = stateStack.last!
    ifStates.append(mergeCurrentState(with: lastState))
    reset(to: lastState)
    pushNewState()
case EndIf:
    ifState = popLastState(from: ifStack)
    lastState = popLastState(from: stateStack)
    mergeCurrentState(with: lastState)    // Merge Else state with state before IfElse
    mergeCurrentState(with: ifState)        // Merge resulting state with state after If to get current state
```
I.e. merge both the If and the Else state with the previous state first (which will remove all variables that are only defined in those blocks), then merge the If and the Else state together to determine the current state after the IfElse. Would something like that work?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: so we should create `InterpreterState` struct?

saelo: Yeah that could be a nice way to implement it!

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -35,102 +35,95 @@ public struct AbstractInterpreter {
...
-        case is EndCodeString:
-            break
-        case is BeginBlockStatement:
-            break
-        case is EndBlockStatement:
-            break
+        case is BeginAnyFunctionDefinition, is BeginCodeString:
+            stateDiffsStack.append(VariableMap())
+        case is EndAnyFunctionDefinition, is EndCodeString:
+            let functionStateDiff = stateDiffsStack.removeLast()
```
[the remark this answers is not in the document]
samo98: I don't really see much difference between `stateDiffsStack.removeLast()` and `popLastState(from: stateDiffsStack)`

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: I don't see the difference between these 2 lines
```
lastState = popLastState(from: stateStack)
mergeCurrentState(with: lastState)    // Merge Else state with state before IfElse
```
and directly adding else branch to current state

samo98: the point is that if you really remove variable then you have to really add it back

saelo: Hm but why do you have to add it back? Once a block ends, if all local variables are removed from `currentState` and from the state stack (when the top entry is popped which has them all set to .nothing), then there should be no more type entries for that variable and that should be correct, shouldn't it? I think it'd be a bug to query the type of a variable after it has gone out of scope, so supporting that shouldn't be necessary. What am I missing?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,108 @@
...
+                if oldType != .nothing {
+                    newType |= currentState[variable]!
+                }
+            }
+
+            if currentState[variable] != newType {
+                // .nothing means variable no longer exists
+                // This information does not need to be forwarded in typeChanges
+                if newType == .nothing {
+                    currentState[variable] = nil
```
saelo, line 63: Nit: I think this should be `currentState.remove(variable)`

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,108 @@
...
+    }
+
+    // Save final variable types in if block
+    public mutating func saveIfState() {
+        ifStack.append(VariableMap())
+        // Iterate over variables that were changed in last if block
+        for (variable, oldType) in stateDiffsStack.last! {
+            // Save only non-local variables
+            if oldType != .nothing {
+                ifStack[ifStack.count - 1][variable] = currentState[variable]
```
saelo, line 84: Nit: maybe it would be nicer to keep the `VariableMap` in a local variable first, then append it to the stack after this loop?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -35,83 +126,28 @@ public struct AbstractInterpreter {
...
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
+            state.mergeStates(from: .conditional, typeChanges: &typeChanges)
+        case is BeginWhile, is BeginDoWhile, is BeginFor, is BeginForIn, is BeginForOf, is BeginAnyFunctionDefinition:
+            state.pushNewState()
+        case is EndWhile, is EndDoWhile, is EndFor, is EndForIn, is EndForOf, is EndAnyFunctionDefinition:
+            state.mergeStates(typeChanges: &typeChanges)
         case is BeginTry:
```
saelo, line 151: Don't you also need to handle those to deal with variables going out of scope now?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -35,83 +126,28 @@ public struct AbstractInterpreter {
...
-            typeChanges += getStateChanges(oldState: functionState, newState: stack.last!)
         case is BeginIf:
-            stack.append(currentState)
+            state.pushNewState()
         case is BeginElse:
-            let ifState = stack.removeLast()
-            ifStack.append(ifState)
-            typeChanges += getStateChanges(oldState: ifState, newState: stack.last!)
+            state.saveIfState()
+            // Exactly one of if/else block is executed. So we should reset types from if block now.
```
saelo, line 143: Maybe the comment should also mention that the else block is executed from the state *before* the if block, so that's why the reset is necessary?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,108 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
-/// Analyzes the types of variables.
-public struct AbstractInterpreter {
-    // States are kept in a stack to support conditional execution.
+private struct InterpreterState {
```
saelo, line 14: Yeah I think using a separate struct here is very nice!

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -35,83 +126,28 @@ public struct AbstractInterpreter {
...
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
-        case is BeginForOf:
-            stack.append(currentState)
-        case is EndForOf:
-            let loopState = stack.removeLast()
-            let previousState = stack.removeLast()
-            stack.append(merge(loopState, previousState))
-            typeChanges += getStateChanges(oldState: loopState, newState: stack.last!)
+            state.mergeStates(from: .conditional, typeChanges: &typeChanges)
```
saelo, line 146: Don't you also need to merge with current (main) state here somehow? AFAIU, this wouldn't be handled correctly now:
```
let v0 = 41;
let v1 = 42;
if (x) {
    v0 = "foo";
} else {
    v1 = "bar";
}
```
Because you merge the current state with the state after the if (which I think would just be "v0 is .string"), so you'd miss the change of `v1` at the `EndIf`? Maybe?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: well yeah it was the original idea, but I kind of lost track what is the benefit of doing that. We do not want to forward this info outside interpreter and we never ask for such type inside interpreter.
So what is the point to make our life harder?

saelo: Ok I thought that would be required to not leak type updates for dead variables into `Program.types`, but I guess if its not, then let's not do that! =)

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: added test for this

Tests/FuzzilliTests/InterpreterTest.swift
```diff
@@ -153,6 +153,26 @@ class AbstractInterpreterTests: XCTestCase {
...
+
+        let v0 = b.loadInt(42)
+        let v1 = b.loadInt(42)
+        XCTAssertEqual(b.type(of: v0), .integer)
+        XCTAssertEqual(b.type(of: v1), .integer)
+        b.beginIf(v0) {
+            b.reassign(v0, to: b.loadString("foo"))
+            b.reassign(v1, to: b.loadString("foo"))
+        }
+
```
saelo, line 168: Nit: remove empty line

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,116 @@
...
+    private mutating func setTypeChange(of v: Variable, to t: Type, at index: Int) {
+        stateChangesStack[index][stateChangesStack[index].count - 1][v] = t
+    }
+
+    /// Set the type of the given variable in the current state.
+    public mutating func setType(of v: Variable, to t: Type) {
+        // Save type in parent state if it is not already
+        if stateChangesStack[stateChangesStack.count - 2].last![v] == nil {
+            setTypeChange(of: v, to: currentState[v] ?? .nothing, at: stateChangesStack.count - 2)
+        }
```
saelo, line 92: Can you assert here that if `stateChangesStack[stateChangesStack.count - 2].last![v]` is not nil that it's then `== currentState[v]`?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,116 @@
...
+                    numUpdatesPerVariable[v]! += 1
+                }
+            }
+        }
+
+        for (v, c) in numUpdatesPerVariable {
+            // Skip updating:
+            // 1. Fallback type for grandchild (no record in prevState)
+            // 2. Local variables, which go out of scope (prevState is .nothing)
+            if prevState[v] == nil || prevState[v] == .nothing {
```
saelo, line 56: When does the `prevState[v] == nil` case occur? Shouldn't `setType` ensure that the parent state always has the old type? Probably related: shouldn't this function also ensure that the new parent state has the previous type for things that changed, e.g. how is this handled:
```
let x = 42;
// diff: x = int
if (c1) {
    // diff: <empty> // but should it be int here?
    if (c2) {
        // diff: x = int (when setting x in the child)
        if (c3) {
            x = "foobar";
            // diff: x = string
        } else {}
        // diff: x = int | string
    } else {}
} else {}
// does x have the correct type here?
```

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,116 @@
...
-/// Analyzes the types of variables.
-public struct AbstractInterpreter {
-    // States are kept in a stack to support conditional execution.
-    private var ifStack: [VariableMap<Type>] = []
-    private var stack = [VariableMap<Type>()]
+private struct InterpreterState {
+    // Save variable changes, supports
+    // 1. Nested states (function states, loop states)
+    // 2. Sibling states (if/else branches)
+    private var stateChangesStack = [[VariableMap<Type>()], [VariableMap<Type>()]]
```
saelo, line 18: Can you make a short comment here why the stack starts with two states? I assume it's because you always need the "parent" state and the active state?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,116 @@
...
+
+    // Set type change on stateChangeStack at given index
+    private mutating func setTypeChange(of v: Variable, to t: Type, at index: Int) {
+        stateChangesStack[index][stateChangesStack[index].count - 1][v] = t
+    }
+
+    /// Set the type of the given variable in the current state.
+    public mutating func setType(of v: Variable, to t: Type) {
+        // Save type in parent state if it is not already
+        if stateChangesStack[stateChangesStack.count - 2].last![v] == nil {
```
saelo, line 90: Nit: Maybe you can make convenience accessors for `activeState` and `parentState` or enum values that represent the indices `stateChangesStack.count - 1` and `stateChangesStack.count - 2` or something like that?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,116 @@
...
+            setTypeChange(of: v, to: currentState[v] ?? .nothing, at: stateChangesStack.count - 2)
```
saelo, line 91: Same here, maybe `stateChangesStack.count - 2` could become some kind of enum value `parentState` or so?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: well when you set type to parentState then later on parentState becomes current and sudden of all parent state does not have any type

samo98: but yeah you're right I should probably use `setType` function

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: yeah the problem is that, this getter you can't use for setting, but at least index getter would be nice

saelo: Ah hm, does it work if you define a setter as well, i.e.
```
var activeState: VariableMap<Type> {
    get {
        let idx = stateChangesStack[stateChangesStack.count - 1].count - 1
        return stateChangesStack[stateChangesStack.count - 1][idx]
    }
    set {
        let idx = stateChangesStack[stateChangesStack.count - 1].count - 1
        stateChangesStack[stateChangesStack.count - 1][idx] = newValue
    }
}
```
?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
samo98: it is not true, type can change multiple times in same block, but we need to update parent state only first time (to save last type in parent block if it is not already)

saelo: Ah yes of course, makes sense

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1170,15 +1170,17 @@ public class ProgramBuilder {
...
         currentCodegenBudget -= 1
 
+        // Update our analyses
+        scopeAnalyzer.analyze(instr)
+        contextAnalyzer.analyze(instr)
+
         // Update type information
         let typeChanges = interpreter?.execute(instr) ?? []
         for (variable, type) in typeChanges {
+            assert(scopeAnalyzer.visibleVariables.contains(variable))
```
saelo, line 1180: Can you also add an assert here that `type != typeOf(variable)`?

samo98: This is a bit complicated because current API returns .unknown when type not set. So we cannot distinguish between (unknown -> unknown and nil->unknown change).
So do we want to extend API of InterpreterState and ProgramTypes just because this assert with:

maybeType(variable) -> Type?
type(variable) -> Type (unknown as default)
?

saelo: Ah right, makes sense. How about `assert(type != typeOf(variable) || type == .unknown)` and a short comment explaining that we can't distinguish no type from unknown type here?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,135 @@
...
+        currentState[v] = t
     }
+
+    public mutating func pushChildState() {
+        stateChangesStack.append([VariableMap()])
+    }
+
+    public mutating func pushSiblingState(typeChanges: inout [(Variable, Type)]) {
+        // Reset last sibling state
+        for (v, t) in stateChangesStack.last!.last! {
```
saelo, line 124: Nit: can this be `activeState`?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -11,17 +11,135 @@
...
+            // Do not save type change if
+            // 1. Variable does not exist in sibling scope (t == .nothing)
+            // 2. Variable is only local in sibling state (parent == .nothing)
+            // 3. No type change happened
+            if t != .nothing, parentStateChanges[v] != .nothing, parentStateChanges[v] != currentState[v] {
```
saelo, line 129: Nit: please use `&&` instead of `,`

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -198,39 +262,48 @@ public struct AbstractInterpreter {
...
         }
         
         return .unknown
     }
     
-    /// Sets the type of the given variable in the current state.
-    /// Return changed variable with its new type
     public mutating func setType(of v: Variable, to t: Type) {
-        stack[stack.count - 1][v] = t
+        state.setType(of: v, to: t)
```
saelo, line 273: Since this is now an single-line function, maybe we can remove it?

samo98: We need to keep this function to expose this functionality for `ProgramBuilder` (setting runtime types)

saelo:
Awesome work!

Issue #154, opened 2020-09-21
Improve ImportCorpusNewCov Speed


WilliamParks:
Currently, ImportCorpusNewCov is particularly slow to finish minimizing all samples. This is primarily due to minimization itself being slow (see multiple comments throughout the codebase, such as [here](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Fuzzer.swift#L165)).

Currently, I've been fuzzing by throwing a huge number of samples into importCorpusNewCov, looking for snippets that add new coverage at the start of a fuzzing run. However, the FuzzilliCli process doing the import takes a very long time to minimize all samples (at least > 10 hours, for an 80k file corpus).

One simple heuristic to improve speed would be to sort samples by size on import. This would decrease the chances that a large sample needs to be minimized, as hopefully the same coverage is achieved by a smaller sample.

Thoughts?

PR #156, opened 2020-09-22
Added corpus sort

WilliamParks:
Fixes issue #154

### DIFF
--- a/Sources/FuzzilliCli/main.swift
+++ b/Sources/FuzzilliCli/main.swift
@@ -336,7 +336,11 @@ fuzzer.sync {
     if let path = corpusImportCovOnlyFile {
         do {
             let data = try Data(contentsOf: URL(fileURLWithPath: path))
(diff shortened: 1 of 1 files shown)


Sources/Fuzzilli/FuzzIL/Program.swift
```diff
@@ -89,6 +89,18 @@ public final class Program {
     public func clearParent() {
         parent = nil
     }
+
+    /// Simple comparison by size, used to sort compiled corpora prior to importing.
+    public static func < (lhs: Program, rhs: Program) -> Bool {
+        return lhs.size < rhs.size
+    }
+
+    /// We consider two programs to be equal if their code is equal
```
saelo, line 98: I'd prefer keeping this code only in FuzzilliTests since it shouldn't be relied on by any other code. Can you use the [`sort(by:)`](https://developer.apple.com/documentation/swift/array/2296801-sort) method to sort the input programs without having Program be Comparable?

PR #162, opened 2020-10-01
Refactored FuzzILTool

saelo:


saelo:
@WilliamParks I did some refactoring of the FuzzILTool, please take a look and see what you think. Also I'm thinking about making a corpus on disk a directory of multiple .fuzzil.protobuf files (as Fuzzilli now generates on disk after https://github.com/googleprojectzero/fuzzilli/commit/1ff0e9c05fa4257c5f973012050e05c35de1871f) instead of a single file and having all the import corpus mechanisms (the new --resume as well as the existing --importCorpus*) simply load all .fuzzil.protobuf files from that directory. This should also simplify the FuzzILTool a bit. What do you think about that?

WilliamParks:
Yeah, that makes sense to me. One useful part of my current workflow is the combine step forces a static check of all the compiled protobufs, which continues to help catch issues in the compiler. Some sort of equivalent would be useful in the updated tool, such as checking all the fuzzil.protobuf files in the specified directory.

saelo:
Cool! I did another revision of this and added a `--checkCorpus` flag which checks all programs in the given directory. Also I've now switched all corpus imports to take a directory and search for ".fuzzil.protobuf" files in them. PTAL and let me know if this works for you! Thanks!

Sources/FuzzILTool/main.swift
```diff
@@ -16,235 +16,153 @@ import Foundation
...
-               """
-let lifter = JavaScriptLifter(prefix: jsPrefix,
+let protoBufFileExtension = ".fuzzil.protobuf"
+
+let jsPrefix = ""
+let jsSuffix = ""
+
+let jsLifter = JavaScriptLifter(prefix: jsPrefix,
                 suffix: jsSuffix,
                 inliningPolicy: NeverInline(),
```
saelo, line 26: Should this be "InlineOnlyLiterals()" to match the default policy used by the fuzzer?

WilliamParks: I specifically chose NeverInline() so that the produced JS matches the provided IL input as closely as possible, to aid sanity-checking the compiler's output. My automated tests are only JS -> IL, so there aren't any issues with breaking what I have currently.

Sources/FuzzILTool/main.swift
```diff
@@ -16,235 +16,153 @@ import Foundation
...
-    var isDir : ObjCBool = false
-    if !FileManager.default.fileExists(atPath: dirPath, isDirectory:&isDir) || !isDir.boolValue {
-        print("Provided directory \(dirPath) is not a valid directory path")
-        exit(-1)
+// Combine multiple protobuf programs into a single corpus file
+else if args.has("--checkCorpus") {
+    var numPrograms = 0, numInvalidPrograms = 0
+    for (filename, program) in loadAllPrograms(in: path) {
+        numPrograms += 1
+        if !program.code.isStaticallyValid() {
```
WilliamParks, line 155: Would it be possible to provide additional debugging information, as to why the program is invalid?

WilliamParks:
@saelo I added two comments regarding what I've found useful thus far. Looks good to me!

Sources/FuzzILTool/main.swift
saelo: Makes sense. There is now also --liftToFuzzIL, which gives you a textual representation of the FuzzIL code. Does that solve that problem as well? In that case I'd change this to `InlineOnlyLiterals()`, if not, I'll leave it `NeverInline()`

Sources/FuzzILTool/main.swift
saelo: Yup! I implemented more verbose error reporting for program decoding. If you run --checkCorpus now with invalid programs, you'll see messages like
```
Failed to load program fuzzdir/corpus/program_20201012104930_99664821-5D15-4AF7-9EB5-A66898F96DC0.fuzzil.protobuf: decoded code is not statically valid: variable v13 was already defined
```
Does that work for you? Also let me know should you ever run into the "Failed to load program \(path) due to unexpected error" case (because that shouldn't happen)

WilliamParks: Great! That should greatly help in debugging

Sources/FuzzILTool/main.swift
WilliamParks: That works!

saelo:
Merged! Thanks for your input @WilliamParks!

Issue #163, opened 2020-10-02
Consider Supporting Variable Deletion


WilliamParks:
Fuzzilli currently supports deleting elements, and properties (both normal and computed). Some regression tests, such as  [regress-1132.js](https://chromium.googlesource.com/v8/v8.git/+/7.3.9/test/mjsunit/regress/regress-1132.js) directly delete an identifier, which Fuzzilli cannot do currently. 

Thoughts? I'm willing to implement it as a new operator, but want to make sure this doesn't cause an issue elsewhere I'm not currently considering.

WilliamParks:
Closed, as I don't think this is actually useful (described in #164)

PR #164, opened 2020-10-02
Implements variable deletion

WilliamParks:
Closes issue #163 

Two questions:
- What's the proper type in CodeGenerators.swift? .anything seems too loose and .object overly restrictive
- I added the required functions at the bottom of the relevant sections across the various files. Is there an order I should be following?

The weight assigned to DeleteVarGenerator in codeGeneratorWeights.swift was chosen to be the same as the similar functions PropertyRemovalGenerator, ElementRemovalGenerator, and ComputedPropertyRemovalGenerator

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -394,6 +394,10 @@ public let CodeGenerators: [CodeGenerator] = [
         b.throwException(v)
     },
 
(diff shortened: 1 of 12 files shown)


WilliamParks:
Looking into this more, I'm not actually sure this provides utility. The examples that I was testing against are all declared without var/let, which allows them to be deleted, but these aren't generate by Fuzzilli currently.

PR #167, opened 2020-10-24
Updates the SPM patch and SPM/JSC/V8 revisions

WilliamParks:


Cloud/Docker/build.sh
```diff
@@ -40,7 +40,7 @@ do
             BUILD_JSC=true
             BUILD_SPIDERMONKEY=true
             BUILD_V8=true
-            BUILD_DUKTAPE=true
+            BUILD_DUKTAPE=false
```
saelo, line 43: WDYT about renaming the option from `all` to `major`, then dropping this line entirely?

Targets/JavaScriptCore/REVISION
```diff
@@ -1 +1 @@
-85041c84e3ed873d6a9d77e276f08eed41dfa648
+3da169520a4cc62ff0ea01e8cc36a8d6a15a809f
```
saelo, line 1: Do the previous patches apply cleanly (i.e. without offsets etc.) to the new revision? Otherwise, please also include the new webkit.patch :)

Cloud/Docker/build.sh
```diff
@@ -36,11 +36,10 @@ do
...
             ;;
-        all)
+        major)
             BUILD_JSC=true
             BUILD_SPIDERMONKEY=true
             BUILD_V8=true
-            BUILD_DUKTAPE=true
             ;;
         *)
             echo "Usage: $0 [fuzzilli|jsc|spidermonkey|v8|duktape|jerryscript|all]"
```
saelo, line 45: Nit: please s/all/major/ here :)

WilliamParks: I changed it in the script, and in the Readme!

PR #168, opened 2020-10-26
Added constructables function and probabilistic selection of object g…

carl-smith:
…roups

### DIFF
--- a/Sources/Fuzzilli/Core/CodeTemplate.swift
+++ b/Sources/Fuzzilli/Core/CodeTemplate.swift
@@ -74,7 +74,7 @@ public class CodeTemplate {
 
                 var group: String? = nil
                 if probability(0.2) {
-                    group = chooseUniform(from: fuzzer.environment.constructables())
+                    group = chooseUniform(from: fuzzer.environment.constructables)
                 }
 
                 // Generate random properties.
@@ -93,11 +93,7 @@ public class CodeTemplate {
                     methods.append(chooseUniform(from: fuzzer.environment.customMethodNames))
                 }
 
-                if let group = group {
-                    return .object(ofGroup: group, withProperties: properties, withMethods: methods)
-                } else {
-                    return .object(withProperties: properties, withMethods: methods)
-                }
+                return .object(ofGroup: group, withProperties: properties, withMethods: methods)
             })
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Fuzzilli/Core/CodeTemplate.swift
```diff
@@ -90,9 +93,12 @@ public class CodeTemplate {
                     methods.append(chooseUniform(from: fuzzer.environment.customMethodNames))
                 }
 
-                return .object(withProperties: properties, withMethods: methods)
+                if let group = group {
+                    return .object(ofGroup: group, withProperties: properties, withMethods: methods)
```
saelo, line 97: Nit: I think the .object type constructor takes a String? as ofGroup parameter, so you should be able to remove the if-else and just leave line 97

Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -195,20 +195,27 @@ public class JavaScriptEnvironment: ComponentBase, Environment {
         logger.info("Have \(customPropertyNames.count) custom property names: \(customPropertyNames)")
         logger.info("Have \(customMethodNames.count) custom method names: \(customMethodNames)")
     }
-    
+
+    public func constructables() -> [String] {
```
saelo, line 199: I think this should be a property and not a method, so that the content is computed just once.

Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -195,20 +195,27 @@ public class JavaScriptEnvironment: ComponentBase, Environment {
         logger.info("Have \(customPropertyNames.count) custom property names: \(customPropertyNames)")
         logger.info("Have \(customMethodNames.count) custom method names: \(customMethodNames)")
     }
-    
+
+    public var constructables: [String] {
+        // These are basically the groups that we already filter for in the
+        // init function and some that can not be constructed with the new
+        // keyword
+        groups.keys.filter({ !$0.contains("Constructor") && !["Math", "JSON", "Reflect", "Symbol"].contains($0) })
```
saelo, line 203: Can you compute this list at the end of the initialize function? It doesn't have to be computed on every access since it won't change afterwards, or can it?

PR #169, opened 2020-11-09
WIP: Fuzzilli compiler

WilliamParks:
This is very much a work in progress, but is at a point where I'd appreciate outside input. The amount of time I've been able to work on this has significantly decreased from when I started, so I'd appreciate help getting it across the finish line.

Current major issues:
    - The code needs a general cleanup, including overall documentation.
    - The current handling of variable declarations results in some compiled .fuzzil.protobuf programs to not be statically valid.
    - The tests I've written are decent, but not great. The test suite here includes a number of JS -> fuzzil samples, which I built by hand
    - The process of getting the flow_ast dependency is not straightforward, but was the least bad option (see notes in Readme).
    - I have not tested on any OS other than Ubuntu 20.04


WilliamParks:
I just added a small documentation update in the Readme. I plan to continue pushing through the documentation and code quality issues over the next several days. This PR exists so we can tackle the algorithmic/structure issues

saelo:
Thanks *a ton* for implementing and sharing this! Since I'm not really familiar with OCaml, and because this is kind of a large PR :), I'll probably need some time to get a rough understanding of how the compiler works. But I'll make comments as I go through it.

WilliamParks:
One additional useful feature would be to filter out known problematic buitlins (example: the Spidermonkey oom test functions) but leaving the other builtins that are not hardcoded in the compiler (not using the "placeholder" flag).

saelo:
I made a Dockerfile to build the compiler:

```
FROM ubuntu:latest

ENV DEBIAN_FRONTEND=noninteractive
ENV SHELL=bash

RUN apt-get -y update && apt-get -y upgrade
RUN apt-get install -y curl ocaml opam npm
RUN npm install -g --unsafe-perm esy

RUN useradd -m builder
WORKDIR /home/builder
USER builder

RUN opam init -a --disable-sandboxing

ADD --chown=builder:builder Compiler Compiler
WORKDIR Compiler

# Pin the flow package...
RUN opam pin add -y flow_parser https://github.com/facebook/flow.git

# ... and add its path to package.json
RUN sed -i 's/.*REPLACE ME.*/    "flow_parser": "link:\/home\/builder\/.opam\/default\/.opam-switch\/sources\/flow_parser\/flow_parser.opam"/' package.json

# Install dependencies
RUN esy install

# And build!
RUN esy build

# Run the tests to verify that the compiler works correctly
RUN esy x test

# Finally, copy the compiler binary into the current directory for easy access
RUN cp _esy/default/build/default/bin/fuzzilli_compiler.exe .
```
Feel free to just add that to this PR :)

Unfortunately, the compiler binary can't currently be added to the [main Fuzzilli docker image](https://github.com/googleprojectzero/fuzzilli/tree/master/Cloud/Docker) because it is compiled on Ubuntu 20.04 while the fuzzilli container uses Ubuntu 18.04 (but there the opam version is too old). 
So maybe this Dockerfile should just be kept separate in the Compiler/ directory for now (and mentioned in the README.md)? It can then probably also be extended to take a directory of JS files, then build the compiler (if necessary) and output a directory of .proto files. But we can see what most sense later on.

Compiler/src/translate.ml
```diff
@@ -0,0 +1,1554 @@
...
+    let (obj_temp_tuple, create_obj_inst) = List.split (List.map temp_func props) in
+    let (obj_temp_list, obj_key_list_unflattened) = List.split obj_temp_tuple in
+    let obj_key_list_flat = List.flatten obj_key_list_unflattened in
+    let flat_inst = List.flatten create_obj_inst in
+    let return_temp = Context.get_new_intermed_temp tracker in
+    let inouts =  (obj_temp_list @ [return_temp]) in
+    let op = if List.length obj_key_list_flat == List.length obj_temp_list then
+            Program_types.Create_object{property_names = obj_key_list_flat} 
+        else
+            Program_types.Create_object_with_spread{property_names = obj_key_list_flat}
```
saelo, line 672: I think I'll refactor the CreateObject and CreateArray operations with the upcoming type system refactoring so that they
- Contain the resulting output type (so that the AbstractInterpreter can be smarter)
- Support Get, Set, and Method syntax (for `CreateObject`) which automatically expose `this`
- Spreading (not more `WithSpread` variants)
- Holes (in the case of arrays)

That should hopefully make this logic a bit simpler. WDYT? Anything else that's missing?

Compiler/src/translate.ml
```diff
@@ -0,0 +1,1554 @@
...
+                | _ ->
+                    let (callee_reg, callee_inst) = proc_expression call_exp.callee tracker in
+                    let arguments = call_exp.arguments in  
+                    let (is_spread_list, arg_regs, arg_inst) = proc_arg_list arguments tracker in
+                    let is_spread = List.fold_left (||) false is_spread_list in
+                    let result_reg = Context.get_new_intermed_temp tracker in
+                    let inst : Program_types.instruction = if is_spread
+                    then
+                        (let op : Operations_types.call_function_with_spread = Operations_types.{spreads = is_spread_list} in
+                        let inst_op : Program_types.instruction_operation = Program_types.Call_function_with_spread op in
```
saelo, line 551: Probably same here, this would probably be a bit easier if we removed those `WithSpread` variants in FuzzIL for calls as well? Then Method calls and constructors would automatically support them.

Compiler/src/translate.ml
```diff
@@ -0,0 +1,1554 @@
...
+    let inst : Program_types.instruction = Program_types.{
+        inouts = [temp_reg];
+        operation = Program_types.Load_integer op;
+    } in
+    (temp_reg, inst)
+
+(* Adds the id of a declared function to the tracker *)
+let id_to_func_type id tracker = 
+    let temp = Context.get_new_intermed_temp tracker in
+    let type_ext = Typesystem_types.{
```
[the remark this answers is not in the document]
saelo: I think there are some other issues around types in the FuzzIL output. I'll write down some notes in the main comment.

saelo (commented):
Had a first (incomplete) look at the code, great stuff!
Apart from the code comments, here are some more general thoughts:

## FuzzIL
I think some of FuzzIL's operations should be refactored to make the job of the compiler easier/possible in some cases. I made some notes in the code, but let me know if there are other areas that are needlessly complex due to FuzzIL design. While FuzzIL's main design goal will still be to facilitate mutations (and static analyses by the AbstractInterpreter), there is definitely room for changing things around

## Types
I think there are currently some issues around types. Generally, types appear in two places in a FuzzIL program:

1. In the [variable types data structure associated with every program](https://github.com/googleprojectzero/fuzzilli/blob/05bb3fbd49ef013b9fa32bc26be1cfc0bb8c33b2/Sources/Fuzzilli/FuzzIL/Program.swift#L39). This should contain the types of all variables and is used for splicing an mutating. It's pretty clear that the compiler should just leave this empty, and then there should probably be some logic during import that computes this information (either statically or through runtime type collection)

2. As part of some instructions. Currently, I think it's just the `BeginFunctionDefiniton`s that have a signature, but in the future there will likely be more (e.g. in `CreateObject`). I think here again the compiler shouldn't attempt to do too much, since Fuzzilli is clearly better equipped (it has the AbstractInterpreter and runtime type collection). Maybe one option here is to add a special `.placeholder` type and just use that exclusively throughout the compiler any time a type is required, then let Fuzzilli take care of replacing that type with something sensible, for example by analyzing the parameter types of functions to infer a sensible signature. WDYT? A very similar problem occurs in other places as well, e.g. when an existing function with a non-trivial signature is mutated, which potentially invalidates the signature, so that logic probably has to be implemented anyway. I'll add this to the TODO list for the upcoming refactoring.

## Variables
It seems the following program is currently handled incorrectly:

```
if (true) {
        var foo = 42;
}
print(foo);
```

This results in 
```
v0 <- LoadBoolean 'true'
BeginIf v0
    v1 <- LoadInteger '42'
BeginElse
EndIf
v2 <- LoadFromScope 'print'
v3 <- LoadFromScope 'v1'
v4 <- CallFunction v2, [v3]
```
Or (as JavaScript):
```
if (true) {
    const v1 = 42;
} else {
}
const v4 = print(v1);
```
Which isn't valid. One option would be to use `StoreToScope`, but the `*Scope` operations are currently mostly a workaround to support `with` statements and weren't designed for this use-case (although they could be adopted to support it). The alternative would probably be to do variable hoisting.

WilliamParks:
Here are some thoughts on the way forward. I've got some other commitments currently that are going to really limit how much time I have to spend on this. My preference is to focus on getting a simple compiler merged, and then additional features can be their own PRs later on, hopefully with community involvement!

Proposed goals for this PR:

- Add Dockerfile
- Add ProgramBuilder equivalent to simplify Protobuf construction, variable management (mentioned above)
- Continue improving documentation
- Figure out a sensible solution to variable hoisting (mentioned above)
- Remove features that cause significant compiler complexity (e.g. spread and classes). The bugs in the trophy case, at least the ones I can see, didn't use these features. They can be added later, potentially with modifications to FuzzIL.

Thoughts?

WilliamParks:
In the Readme changes I just made to the Compiler, I added commands to pull the compiled executable out of the container.

My workflow currently just used a pretty simple Makefile to compile the entirety of a corpus. The main benefit was that it made parallel compilation easy. I can add that if it makes sense (maybe with an option to do all the compilation using docker?)

WilliamParks:
Question on the hoisting issue: What's the purpose of the es5 vs es6 option, in having the first declare variables using `var` and the latter using `let`/`const`?

saelo:
Sure, feel free to include your Makefile. I don't think the Makefile itself needs to be aware of docker, probably it can just be made available inside the docker image?

Variables in FuzzIL are basically block scoped and not function scoped, so using `let`/`const` seems like the more natural choice. If the engine doesn't support these, then `es5` mode can be used to get `var` instead. Lifting has to be deterministic, so the lifter can't just randomly choose `var` or `let`, and so currently it's either always `var` or always `let`/`const`. Not sure if that answers your question?

WilliamParks:
I was thinking that switching to `var` for all declarations would fix the issue, but that fails for the following case:

``` javascript
function d() {
    e = f;
}
function c() {
    var f = 12;
}
```

The two options for hoisting seem to be doing an analysis pass of which vars need to be hoisted (and probably hoisting everything to the top of the Fuzzilli program), or to hoist based on how the variable is declared initially.

My preference is for the first one, as it seems the simplest and will keep from hoisting unnecessarily. Thoughts?

saelo:
Hmm right, so AFAICT, there are three types of variables that need to be handled:

### Block-local variables
```JavaScript
let b = 42;     // or const
```
This translates directly to a FuzzIL variable, and so doesn't need any special handling.

### Function-local variables:
```JavaScript
var f = 42;
```
As you suggested, these can be handled by hoisting them to an outer block and turning them into regular FuzzIL variables. Probably the best way to define these variables in FuzzIL would be through a `LoadUndefined`, since that's basically what a JavaScript engine will do:

```JavaScript
function foo() {
    print(x);
    if (1) {
        var x = 42;
    }
    print(x);
}
foo()
// undefined
// 42
```

There should be at least a few options for choosing the destination for hoisting (I think these basically correspond to your suggestions):
1. Always hoist to the beginning of the current function definition. This is basically what a JavaScript engine will do. On the other hand, this will probably result in somewhat unreadable code for many scripts that only use `var`
2. Hoist to the "highest" block that still contains all uses of the variable. This would probably result in nicely readable FuzzIL programs since scripts that only use `var` will simply be rewritten to use `let` instead.
3. A combination of the previous two: only hoist variables that are accessed outside of the block that they were defined in, but then hoist them to the start of the function (as a JavaScript engine would).

I think I'd prefer 2. or 3 (Probably 2. a bit more).

Alternatively, it would also be possible to add a new FuzzIL operation, for example
- a `StoreToFunctionScope`, which makes the variable visible until the end of the current function and is lifted to a var definition (although this still wouldn't work for the example above, since `x` is accessed before it is defined, which is valid for `var` but not in FuzzIL)
- a `DeclareFunctionVariable` operation which basically acts as a `LoadUndefined` and must come at the start of a function, and which must be combined with a `DefineFunctionVariable` operation which then assigns it a value and which is lifted to a var statement (the `DeclareFunctionVariable` wouldn't be lifted at all)
- Probably other options?

While this would allow Fuzzilli to also generate `var` variables during fuzzing (something that it maybe should be able to do?), my feeling is that it will probably be cleaner and easier to solve this issue in the compiler through hoisting, but let me know if you think otherwise!

### Global variables:
```JavaScript
g = 42;
// this.g (property .g of the global object) is now 42
```
These end up being installed as properties on the global object, and so turning them into "normal" variables would slightly change their semantics. If there is an easy way to detect these types of variable definitions (how do they show up in the AST?), it might be best to either add two new operations to FuzzIL: `StoreToGlobalObject` and `LoadFromGlobalObject` (or something similar, I can do this), or use `StoreToScope` and `LoadFromScope` (maybe better to just use these for now). 

WDYT?

WilliamParks:
For `var`, `let`, and `const`, the `VariableDeclaration` statement in Flow_ast has a field to differentiate between them, so no issues there.

Flow_ast does not give us anything to distinguish between global variable definitions, and standard assignment statements. I have basic functionality here https://github.com/googleprojectzero/fuzzilli/pull/169/files#diff-d199bbd676281f7afedf768400bf31afa081b7e6556b90e7fe2570f2f8a60816R471 (line 471 of translate.ml) that tracks if the variable has already been seen. If it has, a reassign happens, and if not, it makes a new variable. This can be repurposed 

I think handling this in the compiler, via option 2, is best for the standard hoisting issue. I'd prefer to note the global issue as a to-do for future PRs. How'd you like to track the various planned improvements (I've got classes, global vars, spread function, and template literals as issues currently)?

saelo:
Sounds great! Maybe just track todos/future improvements in some markdown file in the compiler directory? Ideally together with a short summary of roughly what would need to be implemented where (e.g. what you wrote above for the global variables).

WilliamParks:
The latest push includes a number of the smaller fixes, as well as converting the compiler to use a ProgramBuilder like interface, which made it much cleaner.

There are three tests in it that fail currently, due to variable scoping issues. I'm planning to roll fixes for those into the variable scoping update.

I've updated one test to use the ProgramBuilder interface (array_assign_sugared.ml). It got rid of the issue of having to manually assign vars, but seems to have come harder to read. I'd appreciate thoughts on it before I work through the rest of the test files.

saelo:
Right, I see what you mean. I guess there are a few options.

1. Turn the ProgramBuilder into a structure (or class?) with a mutable list of instructions, and let each constructor method add the instruction to the builder while only returning the output variable. Basically exactly how Fuzzilli's ProgramBuilder works. The tests would then look something like:

```
let b = init_builder in
let v0 = build_load_integer b 12 in
let v1 = build_create_array b [v0; v0; v0] in
...
finish_building b
```

Drawback: it's not really functional since the `build_X` functions now have side effects

2. Let the `build_X` functions return a new, updated ProgramBuilder instance, with the instruction added to it (basically the above, but without side-effects):
```
let b = init_builder in
let v0, b = build_load_integer b 12 in
let v1, b = build_create_array b [v0; v0; v0] in
...
finish_building b
```

This feels more functional, but needs a few more keystrokes

3. Fall back to something like the previous code structure, but make it clearer what the outputs and inputs are, e.g.
```
let correct =
    {
        outputs = [0];
        inputs = [];
        operation = Load_integer {value = 12L};
    };
    ...
```

4. Other options?

I think from these I'd prefer 2., but WDYT?

adrian-rt:
@WilliamParks  what is the current status of this?

I wanted to give this a try, but building the container fails with the following error:

Step 13/18 : RUN opam pin add -y flow_parser https://github.com/facebook/flow.git
 ---> Running in 5bb84823a7de
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0: git]
[flow_parser.0.80.0] synchronised from git+https://github.com/facebook/flow.git
flow_parser is now pinned to git+https://github.com/facebook/flow.git (version 0.145.0)

[NOTE] Pinning command successful, but your installed packages may be out of sync.
Sorry, no solution found: there seems to be a problem with your request.

WilliamParks:
@adrian-rt It's on my to do list, but I'm not going to be able to look at it for a bit.

adrian-rt:
after reading this again it seems that @saelo  has created the Dockerfile, however strangely enough it doesn't work for me when building the container. @saelo  does that still work for you?

WilliamParks:
@adrian-rt The latest push should fix your issue!

WilliamParks:
I'm working on the variable hoisting update. Variables seem to work fine (hoisting happens at the function level, only to variables that need to be hoisted), and I'm currently working on functions.

Are there any potential issues with the following you can think of?

```javascript
function a(){
    b();
}
function b(){}
```

becomes something like the following, with all of this happening before any other statements
```javascript
let b = undefined;
function a(){
     b();
}
function t(){} // Has the same body as the original b function
a=t;
```

saelo:
Cool! Ah yeah, function hoisting... Yeah I guess that should be the right approach to handle that issue, and should be what the engine does anyway, so there shouldn't really be any semantic differences between the two.

WilliamParks:
This latest update adds basic function hoisting.

The hoisting algorithm isn't algorithmically the best, as it gets called on each function, and walks everything inside of that function, including functions declared within it. The hoisting code then gets called again during the compilation of the first nested function, etc. (Compiling highly nested functions I think ends up being O(N^2) in the number of functions). I have a couple of tests that compile slowly, but it hasn't much of an issue in practice. 

Additionally, I suspect there are issues with how the hoisting algorithm handles variables that shadow one another, as it doesn't keep track of var/let, or removing them from scope at the appropriate times.

I think staying with the program builder interface is the best way to go, as it makes clear where each instruction ends up, at the cost of being more verbose.

From here, I think the big items are cleaning up the code (and doing another sanity check on correctness), and finishing converting the rest of the test cases.

Compiler/src/proto/operations_pb.ml
```diff
@@ -234,6 +234,54 @@ let default_eval_mutable () : eval_mutable = {
   code = "";
 }
 
+type begin_class_definition_mutable = {
```
saelo, line 237: Just to double check, all the code in the src/proto directory is automatically generated by the `build-pbs.sh` script from the protobuf definitions in Fuzzilli?

Compiler/README.md
```diff
@@ -65,5 +65,7 @@ To copy the executable out of the docker container, use:
         - Classes
         - Spread (update Fuzzilli)
     - Improve for/while loops. The current implementation makes all loops while loops, with a comparison against 0.
+    - Improve for-of/for-in loops. A large number of test cases have more complex left sides, while Fuzzilli only supports declaring/defining a new variable.
+        - Example from a regression: for (a[i++].x in [])
```
saelo, line 69: Oh interesting, I'm not even sure I was aware of that kind of JS syntax...

Compiler/package.json
```diff
@@ -14,7 +14,7 @@
     "@opam/uuidm": "*"
   },
   "resolutions": {
-    "flow_parser": "link:{REPLACE ME PATH HERE}/.opam/4.10.0/.opam-switch/sources/flow_parser/flow_parser.opam"
+    "flow_parser": "link:/home/b/.opam/4.10.0/.opam-switch/sources/flow_parser/flow_parser.opam"
```
saelo, line 17: Was this intended?

Compiler/src/ProgramBuilder.ml
```diff
@@ -0,0 +1,727 @@
...
+(* Keeps a mapping of ids to temps *)
+type temp_map_t = (string, var, Core.String.comparator_witness) Core.Map.t
+
+type inst = instruction
+
+(* Keeps a mapping of ids that are still visible in Javascipt, but that Fuzzilli considers out of scope.
+Used to determine if var should be loaded from scope, rather than accessed as a temp *)
+type var_scope_map_t = (string, string, Core.String.comparator_witness) Core.Map.t
+
+type tracker =
```
saelo, line 17: Nit: would it be nicer to call this `builder` instead of `tracker`? It seems like it is doing more than just tracking variables

Compiler/src/proto/operations_pb.ml
WilliamParks: Yes it is! `esy run-script build-pbs` generates all the files in `src/proto/`, using the protobuf files in the main Fuzzilli

Compiler/README.md
WilliamParks: Me neither, until I ran into samples that wouldn't compile! I'm not sure there's much value in compiling such samples. Thoughts? Maybe this should be done as a main thread discussion.

Compiler/package.json
WilliamParks: It was not. I've pushed a corrected version!

Compiler/src/ProgramBuilder.ml
WilliamParks: Yeah, I've been thinking the same. I'll switch it over.

WilliamParks:
I'd appreciate if we could get a list of remaining items on this project. I know there are some particulars of JS that we are missing (variable hoisting isn't 100%, and some JS syntax isn't supported).

I'm tracking that I need to switch over the test cases to use the builder, as well as general commenting/var name cleanups. Anything else?

Compiler/README.md
saelo: Yeah I agree. I think it's good the way it is now, i.e. just mentioning it in the README as TODO.

saelo:
So in general I'm happy to merge this whenever you consider it ready. It would be good if all the current limitations are tracked somewhere (I think most are tracked in the README's TODO section already?), but no need to deal with those now, the compiler is clearly very useful already. 

The only other question I can think of right now is how type information is currently handled, in particular what is stored in the [types field of Program protos](https://github.com/googleprojectzero/fuzzilli/blob/50589694353fad83b7763eca8c5860a119ca63ae/Sources/Fuzzilli/Protobuf/program.proto#L152). I think the right approach would probably be to leave this empty and let Fuzzilli deal with type information (same for e.g. function signature types, those should probably all just be `.anything` and return type `.unknown`). Is that what currently happens? Also this would be good to mention in the README if it isn't already.

WilliamParks:
@saelo The latest push updates all the tests. I'll include a section in the final readme about how type information is handled, as well as a full list of the current limitations.

WilliamParks:
Here's the current status of type data, which I'd like your thoughts on. Is there any easy way to get the numerical values from `.anything` or `.unknown`?

The types for the program are currently an empty list, set in [util.ml](https://github.com/googleprojectzero/fuzzilli/blob/7d4e7c76090e4258f5658f863980d5b46c4757bb/Compiler/src/util.ml#L115). Function types are set [here](https://github.com/googleprojectzero/fuzzilli/blob/7d4e7c76090e4258f5658f863980d5b46c4757bb/Compiler/src/ProgramBuilder.ml#L611) and [here](https://github.com/googleprojectzero/fuzzilli/blob/7d4e7c76090e4258f5658f863980d5b46c4757bb/Compiler/src/ProgramBuilder.ml#L636]) for function parameters and [here](https://github.com/googleprojectzero/fuzzilli/blob/7d4e7c76090e4258f5658f863980d5b46c4757bb/Compiler/src/ProgramBuilder.ml#L655) for return values. 

The generated OCaml protobuf has a concept of default_types, which are set to numerical zero, if that's helpful.

WilliamParks:
I think we can use [the base type values in TypeSystem.swift](https://github.com/googleprojectzero/fuzzilli/blob/0a83e0cad6a8750057f72cc13fb8e273d8c30505/Sources/Fuzzilli/FuzzIL/TypeSystem.swift#L782). Looks like `.anything` is `1 << 12 - 1` and `.unknown` is `1 << 8`

WilliamParks:
I've updated the Readme with a general overview (includes a brief overview of variable/function hoisting, how loops are handled, and types). 

I think it's ready to be merged, pending any further feedback.

Compiler/src/ProgramBuilder.ml
```diff
@@ -59,6 +59,9 @@ type unary_op = Not
     | PreDec
     | PostDec
 
+let unknown_type_int = Int32.sub (Int32.shift_left 1l 12) 1l
```
saelo, line 62: Please add a short comment that these numbers must be kept in sync with TypeSystem.swift.
Also it would probably be nice to have similar Type constructors as [those provided in TypeSystem.swift](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/FuzzIL/TypeSystem.swift#L112). On the other hand, it looks like you only have three places that create types currently in the compiler, so how about just adding a TODO comment here about creating proper type constructors eventually?

saelo:
Awesome, looks good! Left just one comment

WilliamParks:
Fixed!

saelo:
Great! So getting ready to merge this, two last questions:
1. Do you want to add something like "Originally written/developed/... by WilliamParks" into Compiler/README.md somewhere at the top? I think you should, but up to you of course :)
2. Do you have a Twitter account and want me to mention you when I "announce" this?

WilliamParks:
Authorship line added!

I keep my twitter private, but I appreciate the offer!

saelo:
🎉 🎉

PR #170, opened 2020-11-16
JS Environment improvements

amarekano:
- Date and Error builtin added
- Minor improvements to existing builtins 
- Added JSC specific builtins to JSCProfile


Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -23,7 +23,7 @@ public class JavaScriptEnvironment: ComponentBase, Environment {
                                       -1073741824, -536870912, -268435456,                              // -2**32 / {4, 8, 16}
                                       -65537, -65536, -65535,                                           // -2**16
                                       -4096, -1024, -256, -128,                                         // Other powers of two
-                                      -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 16, 64,                 // Numbers around 0
+                                      -2, -1, -0, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 16, 64,                 // Numbers around 0
```
WilliamParks, line 26: Does Swift differentiate -0 from +0 when using Int64?

dtlay: swift only understands it as 0

amarekano: Thanks Bill, I hadn't considered this aspect of Swift but you're right, it resolves to 0, which wasn't the effect I wanted it to have.

Recently there have been a couple of [v8 bugs](https://bugs.chromium.org/p/project-zero/issues/detail?id=1710) around the handling of -0, which I thought would be a good addition to our interesting numbers set. It appears though that this would now need to be handled by the lifter perhaps. What are your thoughts?

saelo: -0 is part of this set: https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Core/JavaScriptEnvironment.swift#L38
We could make these interesting numbers a WeightedList and give some of them (e.g. -0) a higher weight, but I'm not really sure it's worth the effort. Maybe a program template for the HybridEngine would be the right thing, I'll think about it a bit more (I'm currently planning a fairly big refactoring of the code generation engine in general), thanks for bringing it up!

WilliamParks: Between the large number of interesting integers, and how they're selected [here](https://github.com/googleprojectzero/fuzzilli/blob/35c2ecbc2dec88e4b831ebaf45ce20f56767b089/Sources/Fuzzilli/Core/ProgramBuilder.swift#L120), I'd expect that particular integer/float values that are needed to trigger a bug like the -0 ones @amarekano mentioned probably aren't selected by Fuzzilli too often. 

Looking through the V8 bug trophies, a number of them have POCs with integers/floats that appear to have come from the random range selection, although I'm not sure any of the bugs required those random values in particular.

@saelo What are your thoughts on implementing a more complex mutation scheduler, such as [MOpt](https://www.usenix.org/system/files/sec19-lyu.pdf) or similar? Since fuzzing JS engines is inherently much slower than standard AFL targets, my initial thought is that spending additional compute time doing more intelligent mutation selection could pay off.

saelo: Oh very interesting, I'll have a more thorough read of the paper you linked, but it definitely sounds worth exploring! I'd really love to get rid of these arbitrary weights here: https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/FuzzilliCli/main.swift#L284 :) Maybe the same could even be done to choose the CodeGenerator weights.

I'm not sure if it solves this specific problem though (about the -0). Some approaches that come to my mind:

1. A targeted mutator could work (one that replaces number typed values with -0), but I suspect that's too specific for a mutator (and with an algorithm like the above one, it would probably get close to zero weight anyway)
2. A targeted CodeGenerator that emits -0 literals.
3. A specific ProgramTemplate that for example generates or splices functions with floating point arguments, then makes sure to call them at least once with -0 as input

All of these are pretty specific. However, I generally assume that we will soon have many bug class specific ProgramTemplates for the HybridEngine, like this one: https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/FuzzilliCli/Profiles/V8Profile.swift#L28
The idea there is that the samples produced by the MutationEngine are likely to stay somewhat close to the samples that trigger new coverage, thus likely missing bugs that are "far away" from that distribution, for example these -0 issues (probably). That's where the HybridEngine comes into play, which generates programs from scratch and isn't biased by coverage. On the other hand, since the space of valid JavaScript Programs is extremely large, the ProgramTemplates should probably be somewhat specific and targeted towards specific bug classes to have at least a small chance of finding new bugs. With that, I think 3. would be the best here. WDYT?
I'm planning a fairly big refactoring currently to make this new HybridEngine work as described above, which will probably first involve some changes to the type system, as well as changes to the code generators. So I'd prefer to defer this discussion until after that refactoring is done (hopefully next couple of weeks).

WilliamParks: Yeah, I'm not 100% on that paper being the way to go, but I think Fuzzilli could benefit from a more targeted mutator schedule, and I think there are a handful of papers with different approaches applied to AFL. Something like a basic normalized weighing based on a combination of successful execution and new coverage I think could really help, making sure we hit the different mutators, code generators, and constants (ex: -0 probably causing more issues than, say, 64). I'll make a new issue to track this.

amarekano:
Hi @saelo was there something required of me on this PR? I can get rid of the -0 addition to the interesting numbers set and we can revisit it at a later stage if that helps.

Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -766,11 +791,81 @@ public extension ObjectGroup {
             "prototype" : .object()
         ],
         methods: [
-            "resolve"    : [.anything] => .jsPromise,
-            "reject"     : [.anything] => .jsPromise,
-            "all"        : [.object()] => .jsPromise,
-            "race"       : [.object()] => .jsPromise,
-            "allSettled" : [.object()] => .jsPromise,
+            "resolve"    : [.anything...] => .jsPromise,
```
saelo, line 794: Are these parameter types correct? From what I can tell, `resolve` and `reject` take one (arbitrary) argument (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) and `all` and similar methods expect an array of Promises as arguments: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

saelo (commented):
Yeah sorry, guess I just forgot about this one ...
Let's get rid of the `-0` (since it just turns into `0`, which is already there) plus the comments below and then let's merge it :)
Did you try running Fuzzilli in a debug build with this (`swift run -c debug FuzzilliCli ...`)? There are some consistency checks for the environment stuff, but I think they are only active in debug builds.

Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
amarekano: yes, this is incorrect. I've fixed them in the latest commit.

amarekano:
I've run this in debug mode and fixed a runtime crash due to the missing `Date` constructor. Interestingly, this runtime failure wasn't caught by a debug assertion in the code. I feel that the `ObjectGroup` struct should have a separate member to represent the constructor for an `ObjectGroup`. In this way we can detect if a builtin is missing a constructor definition in the same way we do for group name, properties and methods. Or perhaps there's an easier way to do that that I've not considered?

I've addressed your other review comments and also added some JSC specific builtins to JSCProfile. If you feel that the JSC Profile updates are unrelated to this PR then I shall separate that out into its own PR.

saelo:
Cool, thanks! Nice idea with the additional assertion also!

PR #179, opened 2021-01-05
Addes Hermes CVEs

WilliamParks:
Adds Hermes CVEs.

There's discussion regarding a Hermes patch and profile here: https://github.com/facebook/hermes/pull/441, which is a bit cleaner than my implementation.


saelo:
Thanks, great work!

PR #183, opened 2021-02-07
fix duplicate dfg tier-up counter

amarekano:


saelo:
Ah, good catch! Out of curiosity, have you looked into what the differences are between `--thresholdForOptimizeAfterWarmUp`, `--thresholdForOptimizeAfterLongWarmUp`, and `--thresholdForOptimizeSoon`? It seems by default they are all at 1000, so I guess it makes sense to keep them at the same value (like your PR does now).

amarekano:
The functions that set these values in the CodeBlock are defined in [CodeBlock.h](https://github.com/WebKit/WebKit/blob/2d77b53cd9604e46a7b7c5002e3420a0e883f83f/Source/JavaScriptCore/bytecode/CodeBlock.h#L753). If we have a CodeBlock running in the BaselineJIT that has never been optimised before, the execution counter to tier-up to the dfg is set by `OptimizeAfterWarmUp`. If an optimised CodeBlock gets jettisoned and we need to recompile, then the engine resets the execution counter using `OptimizeAfterLongWarmUp` to allow the block to run longer in the lower tier to gather better profiling data. Currently these two have the same value but I imagine the devs decided to have two separate thresholds so it can be tweaked independently for performance reasons.

`OptimizeSoon` is an an interesting one, I've not encountered this very often, but reading the dev comments it seems to be used as a delay to prevent tier-up at the next invocation if the engine determines that it's cheaper to continue running in the lower tier rather than OSR enter into the optimised code immediately. I also suspect that this may be used more often when `concurrentJIT` is enabled and the compilation thread notifies the main thread that optimisation has succeed and the main thread needs to decide on whether to OSR enter right away of delay entry.

saelo:
Cool, thanks!

Issue #185, opened 2021-02-11
Issues building and running the fuzzer


docfate111:
When I run swift build I get an error about run REPL not being defined.
Then when I run   swift run FuzzilliCli /mujs-fuzzilli/build/release/mujs
I get an error about the return statement?
https://github.com/docfate111/mujs-fuzzilli/blob/main/Dockerfile

docfate111:
```
/fuzzilli-0.9.1/Sources/Fuzzilli/Lifting/ScriptWriter.swift:45:40: warning: result of operator '&&' is unused       !c.isLetter && !c.isNumber && !ScriptWriter.specialCharacters.contains(c)             ~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ /fuzzilli-0.9.1/Sources/Fuzzilli/Lifting/ScriptWriter.swift:46:9: error: missing return in a function expected to return 'Bool'         }         ^
```

saelo:
Hm weird. It builds without issues with the latest Swift compiler on both macOS and linux, so maybe you are using an older swift compiler? You could try with this Dockerfile as basis: https://github.com/googleprojectzero/fuzzilli/blob/master/Cloud/Docker/FuzzilliBuilder/Dockerfile

That said, the compiler is complaining about a missing return statement, so maybe just try adding a `return` at the start of [this line](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Lifting/ScriptWriter.swift#L45). If that works, feel free to open a PR for it.

docfate111:
I tried adding the return before opening the issue and it didn't work? I ended up reverting back to version 0.9 since that worked.

drtychai:
I just pulled your [Dockerfile](https://github.com/docfate111/mujs-fuzzilli/blob/main/Dockerfile), up'd fuzzilli to 0.9.1, and it built without issue.

```bash
 ➜ curl https://raw.githubusercontent.com/docfate111/mujs-fuzzilli/main/Dockerfile | sed 's/0\.9/0\.9\.1/g' > Dockerfile && docker build -t mujsfuzzilli-0.9.1 . 
 
 [...]
 cc -Wl,-s -o build/release/mujs build/release/libmujs.o build/release/main.o -lreadline -lrt -lm
[...]

PR #187, opened 2021-02-12
Backwards compatibility for Swift <v5.2

drtychai:
After investigating #185 I was able to root-cause the error - Swift <v5.2 is not compatible, as is, for the following reasons:

1. Implicit returns just don't work in <v5.2 (no idea why)
1. `runREPRLTests` requires preprocesser flags to disable run
1. `dirPath` typo in FuzzILTool (unrelated bug)

So explicit `return`'s added in, and preprocesser flag set based on `>=5.2`, i.e.,
```swift
#if swift(>=5.2)
// Run a couple of tests now
runREPRLTests()
#endif
```

Testing Dockerfile (will fail to build, as is):
```dockerfile
FROM swift:5.0
# swift v5.0 breaks on current master branch of googleprojectzero/fuzzilli
#ENV FUZZILLI "--branch swift-v5.0 https://github.com/drtychai/fuzzilli"

#FROM swift:5.2
ENV FUZZILLI "https://github.com/googleprojectzero/fuzzilli"

ENV F_DIR /fuzzilli

RUN apt update && apt upgrade -y && apt install -y git 
RUN git clone --depth 1 ${FUZZILLI} ${F_DIR}

WORKDIR ${F_DIR}
RUN swift build
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,7 +32,7 @@ let package = Package(
         .target(name: "libreprl", dependencies: []),
         .target(name: "libcoverage", dependencies: [], linkerSettings: [.linkedLibrary("rt", .when(platforms: [.linux]))]),
         .target(name: "Fuzzilli", dependencies: ["SwiftProtobuf", "libsocket", "libreprl", "libcoverage", "JS"]),
-        //.target(name: "REPRLRun", dependencies: ["libreprl"]),
+        .target(name: "REPRLRun", dependencies: ["libreprl"]),
         .target(name: "FuzzilliCli", dependencies: ["Fuzzilli"]),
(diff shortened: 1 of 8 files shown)


saelo:
Ah great, thanks for doing this! Do you know why the call to `runREPRLTests` doesn't work on Swift <=5.2? Is it because the function is only defined later in the file or something like that? If you know, maybe just add a quick comment, otherwise don't bother, I'll just merge it :)

drtychai:
Huh - it _was_ that; `function runREPRLTests` moved to above it's callee. Build works for >=v5.0

saelo:
Great, thanks!

PR #195, opened 2021-05-13
Moves core file creation check before moving corpus directories

WilliamParks:
Previously, the fuzzer will move the corpus directory to `old_corpus` in preparation to resume a session, and then exit due to `kernel.core_pattern` not being set properly. This forces the user to have to move the files back in order to properly resume the session, which is not intuitive, and results in data loss if not done prior to restarting the fuzzer.

This moves the `kernel.core_pattern` check before moving the directory.


saelo:
Cool, thanks!

PR #196, opened 2021-05-20
Added support for finally blocks

amarekano:
This PR adds support for finally blocks in try-catch statements.

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -444,12 +444,18 @@ public let CodeGenerators: [CodeGenerator] = [
         b.beginTry() {
             b.generateRecursive()
         }
-        withProbability(0.5, do: {
+        // Generate either a catch or finally block
+        withProbability(0.66, do: {
             b.beginCatch() { _ in
                 b.generateRecursive()
             }
-        })
-        withProbability(0.5, do: {
+            // If we are generating a catch block then consider appending a finally block
+            withProbability(0.5, do: {
+                b.beginFinally {
+                    b.generateRecursive()
+                }
+            })
+        }, else: {
             b.beginFinally() {
                 b.generateRecursive()
             }
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -444,9 +444,16 @@ public let CodeGenerators: [CodeGenerator] = [
         b.beginTry() {
             b.generateRecursive()
         }
-        b.beginCatch() { _ in
-            b.generateRecursive()
-        }
+        withProbability(0.5, do: {
+            b.beginCatch() { _ in
+                b.generateRecursive()
+            }
+        })
+        withProbability(0.5, do: {
+            b.beginFinally() {
+                b.generateRecursive()
+            }
+        })
         b.endTryCatch()
     },
 
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -444,18 +444,21 @@ public let CodeGenerators: [CodeGenerator] = [
         b.beginTry() {
             b.generateRecursive()
         }
-        // Generate either a catch or finally block
-        withProbability(0.66, do: {
+        withEqualProbability({
+            // try-catch-finally
             b.beginCatch() { _ in
                 b.generateRecursive()
             }
-            // If we are generating a catch block then consider appending a finally block
-            withProbability(0.5, do: {
-                b.beginFinally {
-                    b.generateRecursive()
-                }
-            })
-        }, else: {
+            b.beginFinally() {
+                b.generateRecursive()
+            }
+        }, {
+            // try-catch
+            b.beginCatch() { _ in
+                b.generateRecursive()
+            }
+        }, {
+            // try-finally
             b.beginFinally() {
                 b.generateRecursive()
             }
--- a/Sources/Fuzzilli/FuzzIL/Semantics.swift
+++ b/Sources/Fuzzilli/FuzzIL/Semantics.swift
@@ -125,7 +125,7 @@ extension Operation {
         case is BeginForOf:
             return endOp is EndForOf
         case is BeginTry:
-            return endOp is BeginCatch || endOp is BeginFinally || endOp is EndTryCatch
+            return endOp is BeginCatch || endOp is BeginFinally
         case is BeginCatch:
             return endOp is BeginFinally || endOp is EndTryCatch
         case is BeginFinally:
--- a/Sources/Fuzzilli/FuzzIL/Semantics.swift
+++ b/Sources/Fuzzilli/FuzzIL/Semantics.swift
@@ -125,8 +125,10 @@ extension Operation {
         case is BeginForOf:
             return endOp is EndForOf
         case is BeginTry:
-            return endOp is BeginCatch
+            return endOp is BeginCatch || endOp is BeginFinally || endOp is EndTryCatch
         case is BeginCatch:
+            return endOp is BeginFinally || endOp is EndTryCatch
+        case is BeginFinally:
             return endOp is EndTryCatch
         case is BeginCodeString:
             return endOp is EndCodeString
--- a/Sources/Fuzzilli/Minimization/BlockReducer.swift
+++ b/Sources/Fuzzilli/Minimization/BlockReducer.swift
@@ -116,9 +116,9 @@ struct BlockReducer: Reducer {
     }
 
     private func reduceTryCatch(tryCatch: BlockGroup, in code: inout Code, with verifier: ReductionVerifier) {
-        // We first try to remove only the try-catch block instructions.
+        // We first try to remove only the try-catch-finally block instructions.
         // If that doesn't work, then we try to remove the try block including
-        // its last instruction but keepp the body of the catch block.
+        // its last instruction but keep the body of the catch and/or finally block.
         // If the body isn't required, it will be removed by the
         // other reducers. On the other hand, this successfully
         // reduces code like
@@ -128,19 +128,22 @@ struct BlockReducer: Reducer {
         //         throw 42;
         //     } catch {
         //         do_something_important2();
+        //     } finally {
+        //            do_something_important3();
         //     }
         //
         // to
         //
         //     do_something_important1();
         //     do_something_important2();
+        //     do_something_important3();
         //
 
         var candidates = [Int]()
 
-        candidates.append(tryCatch[0].index)
-        candidates.append(tryCatch[1].index)
-        candidates.append(tryCatch[2].index)
+        for i in 0...tryCatch.numBlocks {
+            candidates.append(tryCatch[i].index)
+        }
 
         if verifier.tryNopping(candidates, in: &code) {
             return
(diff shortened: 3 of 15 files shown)


Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -444,9 +444,16 @@ public let CodeGenerators: [CodeGenerator] = [
         b.beginTry() {
             b.generateRecursive()
         }
-        b.beginCatch() { _ in
-            b.generateRecursive()
-        }
+        withProbability(0.5, do: {
```
saelo, line 447: I think this can now end up generating just 
```
try {
   ...;
}
```
which then wouldn't be valid JavaScript ("At least one catch-block, or a finally-block, must be present", from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#description). So you'll have to ensure that at least one of the two is generated. I think ideally, you'd generate try-catch, try-finally, and try-catch-finally statements with equal probability. WDYT?

Sources/Fuzzilli/FuzzIL/Semantics.swift
```diff
@@ -125,8 +125,10 @@ extension Operation {
         case is BeginForOf:
             return endOp is EndForOf
         case is BeginTry:
-            return endOp is BeginCatch
+            return endOp is BeginCatch || endOp is BeginFinally || endOp is EndTryCatch
```
saelo, line 128: Remove `|| endOp is EndTryCatch` here

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -363,6 +363,50 @@ class LifterTests: XCTestCase {
...
+            b.endTryCatch()
+        }
+        b.callFunction(f, withArgs: [b.loadBool(true), b.loadInt(1)])
+
+        let program = b.finalize()
+
+        let lifted_program = fuzzer.lifter.lift(program)
+        let expected_program = """
+        function v0(v1,v2,v3) {
+            try {
```
saelo, line 393: Maybe also include a try-catch and a try-finally statement (within the same program or in another one)?

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -116,9 +116,9 @@ struct BlockReducer: Reducer {
     }
 
     private func reduceTryCatch(tryCatch: BlockGroup, in code: inout Code, with verifier: ReductionVerifier) {
```
saelo, line 118: Nit: rename to reduceTryCatchFinally?

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -128,19 +128,22 @@ struct BlockReducer: Reducer {
         //         throw 42;
         //     } catch {
         //         do_something_important2();
+        //     } finally {
+        //            do_something_important3();
```
saelo, line 132: Nit: indention

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -155,8 +158,7 @@ struct BlockReducer: Reducer {
                 break
             }
         }
-
-        if candidates.count == 4 && verifier.tryNopping(candidates, in: &code) {
+        if (candidates.count - tryCatch.numBlocks) == 0 && verifier.tryNopping(candidates, in: &code) {
```
saelo, line 161: I think this would need to be 1 instead of 0, but let's maybe just simplify this to `if removedLastTryBlockInstruction && verifier.tryNopping(candidates, in: &code) { ...`, then add `var removedLastTryBlockInstruction = false` above the preceding loop, and set it to true when appending the index. That should make it more obvious what's going on

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -168,9 +170,10 @@ struct BlockReducer: Reducer {
         //             const v17 = Math(v16,v16);
         //         }
         //      } catch {
+        //      } finally {
         //      }
         //
-        if candidates.count == 4 {
+        if (candidates.count - tryCatch.numBlocks) == 0 {
```
saelo, line 176: Then also turn this into just `if removedLastTryBlockInstruction {`

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -168,9 +170,10 @@ struct BlockReducer: Reducer {
...
             candidates.removeLast()
         }
```
saelo, line 179: I think the following comment isn't correct, so while you are at it, maybe change it to "Remove all instructions in the body of the try block"? Also (again not your code), I think line 182 can just become `if !(code[i].op is Nop) {` (remove `tryCatch.`)

saelo (commented):
Cool, thanks for implementing this! Left some comments, should be good to merge after addressing those :)

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: My bad! I've fixed that now with the latest commit. The math on the probability distribution may be a bit off, let me know if this needs tweaking :)

Tests/FuzzilliTests/LifterTest.swift
amarekano: testcases updated

amarekano:
I've addressed your review comments in the latest commit. 

Just a heads up, I have some concerns around the `tryCatchFinally` reducer and I'm seeing intermittent crashes when reducing try-catch-finally blocks. Consider the following program:

```
let v0 = 1337
try {
//... number of instructions
} catch {
let v0 = "foobar"
//... more catch block instructions
}
```
When the reducer tries to minimise this program, it attempts to Nop out the try block and the catch statement. So we end up with
```
let v0 = 1337
Noped BeginTry
Noped instructions
Noped BeginCatch
let v0 = "foobar"
//... catch block instructions
Noped EndTry
```
This generates invalid FuzzIL and fails static validation (due to v0 being re-defined) which triggers an assertion at runtime. I'm not sure if this is due to the bug in my code updates to the reducer (i.e. from the commit before addressing the review comments) or an issue elsewhere (e.g. should we Nop out the catch and finally blocks when reducing?). Thoughts?

I'm going to run the fuzzer with the latest changes to this PR and see if I can recreate these crashes and report back,

Sources/Fuzzilli/Core/CodeGenerators.swift
saelo: Yeah I think the math should be correct. Alternatively, I think you could also write it as
```
b.beginTry() { b.generateRecursive() }
withEqualProbability({
    // try-catch-finally
    b.beginCatch() { _ b.generateRecursive() }
    b.beginFinally() { b.generateRecursive() }
}, {
    // try-catch
    b.beginCatch() { _ b.generateRecursive() }
}, {
    // try-finally
    b.beginFinally() { b.generateRecursive() }
})
```
Which might make the final probabilities a bit more obvious/intuitive (at the slight cost of duplicating a few lines).

saelo:
Ok interesting. So I think 
```
let v0 = 1337
try {
//... number of instructions
} catch {
let v0 = "foobar"
//... more catch block instructions
}
```
should already be invalid FuzzIL since v0 is declared two time ([this](https://github.com/googleprojectzero/fuzzilli/blob/2bd387dc907cac95d518c359fd12bf74752b71e6/Sources/Fuzzilli/FuzzIL/Code.swift#L205) should catch it). So I'm not sure what might have happened there. I'll take another look at the reducer, but I agree with you, running it for a bit and monitoring the outputs is probably the best approach here.

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: good shout! definitely more intuitive and easier to maintain.

amarekano:
I've run the fuzzer built from commit ad7449d for a few hours with the following configuration and bumping up the weight on the `TryCatchGenerator`:

```
.build/debug/FuzzilliCli --profile=jsc --engine=multi --timeout=800 --logLevel=info --jobs=16 --storagePath=../fuzz_test --overwrite ./jsc
```
I haven't noticed a crash/debug assertion related to try catch generation or reduction but the debug build did crash with the following assertion:

```
Fuzzilli/ProgramBuilder.swift:508: Fatal error: Unexpectedly found nil while unwrapping an Optional value
Current stack trace:

0    libswiftCore.so                    0x00007ffff7dfd210 swift_reportError + 50
1    libswiftCore.so                    0x00007ffff7e71fc0 _swift_stdlib_reportFatalErrorInFile + 112
2    libswiftCore.so                    0x00007ffff7b5bf56 <unavailable> + 1425238
3    libswiftCore.so                    0x00007ffff7b5bb7f <unavailable> + 1424255
4    libswiftCore.so                    0x00007ffff7b5b91c <unavailable> + 1423644
5    libswiftCore.so                    0x00007ffff7b5b430 _assertionFailure(_:_:file:line:flags:) + 441
6    FuzzilliCli                        0x00005555557a5bbf <unavailable> + 2431935
7    FuzzilliCli                        0x00005555557a5c44 <unavailable> + 2432068
8    FuzzilliCli                        0x00005555557b6f04 <unavailable> + 2502404
9    libswiftCore.so                    0x00007ffff7b428d0 Collection.map<A>(_:) + 524
10   FuzzilliCli                        0x00005555557a1400 ProgramBuilder.generateVariable(ofType:) + 11788
11   FuzzilliCli                        0x00005555556fdd63 <unavailable> + 1744227
12   FuzzilliCli                        0x00005555556f6657 <unavailable> + 1713751
13   FuzzilliCli                        0x00005555556f6691 <unavailable> + 1713809
14   FuzzilliCli                        0x00005555556f6a40 CodeGenerator.run(in:with:) + 146
15   FuzzilliCli                        0x00005555557aaf40 ProgramBuilder.run(_:recursiveCodegenBudget:) + 2214
16   FuzzilliCli                        0x00005555557ac08f <unavailable> + 2457743
17   FuzzilliCli                        0x00005555557038bc <unavailable> + 1767612
18   FuzzilliCli                        0x00005555557b8871 <unavailable> + 2508913
19   FuzzilliCli                        0x0000555555a2df60 withEqualProbability<A>(_:) + 103
20   FuzzilliCli                        0x00005555557ab9b0 ProgramBuilder.generateInternal() + 878
21   FuzzilliCli                        0x00005555557ac0c0 ProgramBuilder.generate(n:) + 219
22   FuzzilliCli                        0x00005555557bdfdc <unavailable> + 2531292
23   FuzzilliCli                        0x00005555557b9a00 ProgramTemplate.generate(in:) + 281
24   FuzzilliCli                        0x0000555555717ee0 HybridEngine.generateTemplateProgram(baseTemplate:) + 849
25   FuzzilliCli                        0x00005555557183f0 HybridEngine.fuzzOne(_:) + 740
26   FuzzilliCli                        0x00005555557194f1 <unavailable> + 1856753
27   FuzzilliCli                        0x0000555555790210 MultiEngine.fuzzOne(_:) + 240
28   FuzzilliCli                        0x0000555555790b91 <unavailable> + 2345873
29   FuzzilliCli                        0x000055555587bf90 Fuzzer.fuzzOne() + 893
```
From the stack trace it would seem that this is an issue unrelated to the current PR but it's probably worth investigating and tracking as a separate issue.

I've rebased this PR with the latest changes from main so this should be good to merge.

saelo:
Cool! Yeah that crash indeed looks unrelated. It's unfortunately not really actionable without line numbers I guess, but if I get some time I'll try to reproduce it. Thanks!

amarekano:
quick update on the crashes. I've upgraded my setup from Swift 5.4 to 5.4.1 and I no longer see crashes.  I think this was a Swift quirk that was addressed with the update. Apologies, if I've set you on a wild goose chase. I don't think this requires any further investigation :)

PR #197, opened 2021-05-21
support generation of array with holes

amarekano:
The JavaScriptLifter is unable to generate arrays with holes since it lifts the *loadUndefined* operation to the string *undefined*. As a result, this replaces holes in an array with JS *undefined*. Semantically, this should not be a problem since holes in an array are considered as undefined values and testing array[index_at_hole] === undefined would return true.

However, I've come across this [JSC testcase](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/array-indexof-array-prototype-change.js) that demonstrates how arrays initialised with holes are affected by JIT optimisations. In this testcase, if the hole in the initialisation were to be replaced with *undefined*, it would case the testcase to  fail (presumably due to the JIT optimising array access to the hole differently).

I did consider creating a new operation (e.g. *loadHole*) for the purpose of adding holes to an array but that seemed redundant  when a minor change to the JSLifter would achieve the same effect.

### DIFF
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -187,7 +187,8 @@ public class JavaScriptLifter: Lifter {
                 output = ObjectLiteral.new("{" + properties.joined(separator: ",") + "}")
 
             case is CreateArray:
-                let elems = instr.inputs.map({ expr(for: $0).text == "undefined" ? "" : expr(for: $0).text }).joined(separator: ",")
+                // When creating arrays, treat undefined elements as holes. This also relies on literals always being inlined.
+                let elems = instr.inputs.map({ let text = expr(for: $0).text; return text == "undefined" ? "" : text }).joined(separator: ",")
                 output = ArrayLiteral.new("[" + elems + "]")
 
             case let op as CreateObjectWithSpread:
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -187,7 +187,7 @@ public class JavaScriptLifter: Lifter {
                 output = ObjectLiteral.new("{" + properties.joined(separator: ",") + "}")
 
             case is CreateArray:
-                let elems = instr.inputs.map({ expr(for: $0).text }).joined(separator: ",")
+                let elems = instr.inputs.map({ expr(for: $0).text == "undefined" ? "" : expr(for: $0).text }).joined(separator: ",")
                 output = ArrayLiteral.new("[" + elems + "]")
 
             case let op as CreateObjectWithSpread:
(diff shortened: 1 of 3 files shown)


Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -187,7 +187,7 @@ public class JavaScriptLifter: Lifter {
                 output = ObjectLiteral.new("{" + properties.joined(separator: ",") + "}")
 
             case is CreateArray:
-                let elems = instr.inputs.map({ expr(for: $0).text }).joined(separator: ",")
+                let elems = instr.inputs.map({ expr(for: $0).text == "undefined" ? "" : expr(for: $0).text }).joined(separator: ",")
```
saelo, line 190: Nit: can you please add a comment that (1) explains what this does (e.g. "treat undefined elements as holes") and (2) mention that this relies on literals always being inlined (otherwise, `expr(for: $0).text` would just give you something like `v3`)? Also maybe it would be slightly nicer to only evaluate `expr(for: $0).text` once, but then you'd need to assign the result to a temporary variable. Really up to you what you prefer :)

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -363,6 +363,31 @@ class LifterTests: XCTestCase {
...
+        let fuzzer = makeMockFuzzer()
+        let b = fuzzer.makeBuilder()
+
+        var initialValues = [Variable]()
+        initialValues.append(b.loadInt(1))
+        initialValues.append(b.loadInt(2))
+        initialValues.append(b.loadUndefined())
+        initialValues.append(b.loadInt(4))
+        initialValues.append(b.loadUndefined())
+        initialValues.append(b.loadInt(5))
```
saelo, line 377: Nit: should this be 6 instead of 5? Doesn't really matter though

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -363,6 +363,31 @@ class LifterTests: XCTestCase {
...
+        b.createArray(with: initialValues)
```
saelo, line 378: How about adding the following code here as well:
```
let v = b.loadString("foobar")
b.reassign(v, b.loadUndefined())
initialValues.append(v)
```
and then checking that the expected program is (I think)
```
let v6 = 42;
v6 = undefined;
const v8 = [1,2,,4,,6,v6];
```
?

Just to be sure that the lifter really only treats the `undefined` literal as hole and not other variables that happen to contain the undefined value.

saelo (commented):
Nice! Yeah I agree, the alternative would probably be to add a `CreateArrayWithHoles` operation (`LoadHole` wouldn't really work I think, unless you treat it the same as `LoadUndefined` for all other uses), but your solution is much simpler and should work fine in practice, so let's go with that! :)

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: done, comments and syntax cleanup added ;)

Tests/FuzzilliTests/LifterTest.swift
amarekano: good spot! I've updated the testcase to include the inlining test

amarekano:
I've addressed your review comments and we should good to go.

saelo:
Nice, thanks!

PR #200, opened 2021-06-03
Add CallComputedMethod operation

amarekano:
This PR adds support to generating computed method calls to fuzzilli. This would now allow generating JS of the form:
 ```
let v = callee[method](arguments)
```
An example of this usecase would be the use of `Symbol.iterator` demonstrated in this [JSC testcase](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/arguments-iterator.js#L36)

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -254,8 +254,9 @@ public let CodeGenerators: [CodeGenerator] = [
             guard b.mode != .conservative else { return }
             methodName = b.genMethodName()
         }
+        let method = b.loadString(methodName!)
         guard let arguments = b.randCallArguments(forMethod: methodName!, on: obj) else { return }
-        b.callMethod(methodName!, on: obj, withArgs: arguments)
+        b.callComputedMethod(method, on: obj, withArgs: arguments)
     },
 
     CodeGenerator("FunctionCallGenerator", input: .function()) { b, f in
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -248,6 +248,16 @@ public let CodeGenerators: [CodeGenerator] = [
         b.callMethod(methodName!, on: obj, withArgs: arguments)
     },
 
+    CodeGenerator("ComputedMethodCallGenerator", input: .object()) { b, obj in
+        var methodName = b.type(of: obj).randomMethod()
+        if methodName == nil {
+            guard b.mode != .conservative else { return }
+            methodName = b.genMethodName()
+        }
+        guard let arguments = b.randCallArguments(forMethod: methodName!, on: obj) else { return }
+        b.callMethod(methodName!, on: obj, withArgs: arguments)
+    },
+
     CodeGenerator("FunctionCallGenerator", input: .function()) { b, f in
         guard let arguments = b.randCallArguments(for: f) else { return }
         b.callFunction(f, withArgs: arguments)
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -351,8 +351,8 @@ public struct AbstractInterpreter {
         case let op as CallMethod:
             set(instr.output, inferMethodSignature(of: op.methodName, on: instr.input(0)).outputType)
 
-        case let op as CallComputedMethod:
-            set(instr.output, inferMethodSignature(of: op.methodName, on: instr.input(0)).outputType)
+        case is CallComputedMethod:
+            set(instr.output, inferMethodSignature(of: instr.input(1).description, on: instr.input(0)).outputType)
 
         case is Construct:
             set(instr.output, inferConstructedType(of: instr.input(0)))
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -351,6 +351,9 @@ public struct AbstractInterpreter {
         case let op as CallMethod:
             set(instr.output, inferMethodSignature(of: op.methodName, on: instr.input(0)).outputType)
 
+        case let op as CallComputedMethod:
+            set(instr.output, inferMethodSignature(of: op.methodName, on: instr.input(0)).outputType)
+
         case is Construct:
             set(instr.output, inferConstructedType(of: instr.input(0)))
 
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -340,8 +340,11 @@ public struct AbstractInterpreter {
         case let op as LoadProperty:
             set(instr.output, inferPropertyType(of: op.propertyName, on: instr.input(0)))
 
+        // TODO: An additional analyzer is required to determine the runtime value of the output variable generated from the following operations
+        // For now we treat this as .unknown
         case is LoadElement,
-             is LoadComputedProperty:
+             is LoadComputedProperty,
+             is CallComputedMethod:
             set(instr.output, .unknown)
 
         case is CallFunction,
@@ -350,9 +353,6 @@ public struct AbstractInterpreter {
 
         case let op as CallMethod:
             set(instr.output, inferMethodSignature(of: op.methodName, on: instr.input(0)).outputType)
-
-        case is CallComputedMethod:
-            set(instr.output, inferMethodSignature(of: instr.input(1).description, on: instr.input(0)).outputType)
 
         case is Construct:
             set(instr.output, inferConstructedType(of: instr.input(0)))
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

File: Sources/Fuzzilli/Core/CodeGenerators.swift [hunk 1]
```diff
@@ -248,6 +248,16 @@ public let CodeGenerators: [CodeGenerator] = [
...
     },
 
+    CodeGenerator("ComputedMethodCallGenerator", input: .object()) { b, obj in
+        var methodName = b.type(of: obj).randomMethod()
+        if methodName == nil {
+            guard b.mode != .conservative else { return }
+            methodName = b.genMethodName()
+        }
+        guard let arguments = b.randCallArguments(forMethod: methodName!, on: obj) else { return }
+        b.callMethod(methodName!, on: obj, withArgs: arguments)
```
saelo, line 258: shouldn't this be `callComputedMethod`?

saelo (commented):
Nice, thanks for implementing this!
So in general, I think the `CallComputedMethod` should be similar to e.g. `LoadComputedProperty` in that it uses a variable as method name and not a string constant. Then, the ProgramBuilder interface would be `public func callComputedMethod(_ name: Variable, on object: Variable, withArgs arguments: [Variable]) -> Variable`.
For the CodeGenerator, I think what you could do is basically this:
```
CodeGenerator("ComputedMethodCallGenerator", input: .object()) { b, obj in
    var methodName = b.type(of: obj).randomMethod()
    if methodName == nil { ... }
    let method = b.loadString(methodName)
    guard let arguments = b.randCallArguments(forMethod: methodName!, on: obj) else { return }
    b.callComputedMethod(method, on: obj, withArgs: arguments)
}
```
What do you think?

amarekano:
Thanks for the feedback, switching from `String` to `Variable` sounds good to me. I've addressed your review comments and updated this PR.

As a side note: this approach has also given me ideas on how to address global var declarations using `StoreToScope`. This is wrt to the [discussion](https://github.com/googleprojectzero/fuzzilli/pull/169#issuecomment-752158233) on the FuzzIL compiler. I'll have a separate PR to run these changes by you.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -415,6 +415,17 @@ class CallMethod: Operation {
...
 }
 
+class CallComputedMethod: Operation {
+    var numArguments: Int {
+        return numInputs - 2
+    }
+
+    init(numArguments: Int) {
+        // reference object is the first input and method name is the second input
+        super.init(numInputs: numArguments + 2, numOutputs: 1, attributes: [.isParametric, .isVarargs, .isCall])
```
saelo, line 425: I think this is now no longer `.isParametric`

Sources/Fuzzilli/Lifting/FuzzILLifter.swift
```diff
@@ -150,6 +150,10 @@ public class FuzzILLifter: Lifter {
             let arguments = instr.inputs.dropFirst().map({ $0.identifier })
             w.emit("\(instr.output) <- CallMethod \(input(0)), '\(op.methodName)', [\(arguments.joined(separator: ", "))]")
 
+        case is CallComputedMethod:
+            let arguments = instr.inputs.dropFirst(2).map({ $0.identifier })
+            w.emit("\(instr.output) <- CallComputedMethod \(input(0)), '\(input(1))', [\(arguments.joined(separator: ", "))]")
```
saelo, line 155: Nit: please remove the `'` around `input(1)` (similar to `LoadComputedProperty`)

Sources/Fuzzilli/Minimization/CallArgumentReducer.swift
```diff
@@ -41,6 +41,15 @@ struct CallArgumentReducer: Reducer {
...
                 let newInstr = Instruction(newOp, output: instr.output, inputs: Array(instr.inputs.dropLast()))
                 verifier.tryReplacing(instructionAt: instr.index, with: newInstr, in: &code)
+
+            case let op as CallComputedMethod:
+                guard op.numArguments > minArgCount else {
+                    break
+                }
+
+                let newOp = CallComputedMethod(numArguments: op.numArguments - 2)
+                let newInstr = Instruction(newOp, output: instr.output, inputs: Array(instr.inputs.dropLast(2)))
```
saelo, line 51: Why are you dropping two arguments at the same time and not just one like the above cases?

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -351,6 +351,9 @@ public struct AbstractInterpreter {
         case let op as CallMethod:
             set(instr.output, inferMethodSignature(of: op.methodName, on: instr.input(0)).outputType)
 
+        case is CallComputedMethod:
+            set(instr.output, inferMethodSignature(of: instr.input(1).description, on: instr.input(0)).outputType)
```
saelo, line 355: This doesn't quite work. Without an additional analyzer that can determine the runtime value of a variable this won't be possible, so I think you can just set the result to `.unknown` directly, similar to how e.g. [`LoadComputedProperty` is handled](https://github.com/googleprojectzero/fuzzilli/blob/40a119bdba15f764b96330c19a3d9de55516459d/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift#L344). Feel free to make a TODO comment about improving this eventually if you want though

Sources/Fuzzilli/FuzzIL/Instruction.swift
```diff
@@ -590,6 +592,9 @@ extension Instruction: ProtobufConvertible {
             op = Await()
         case .callMethod(let p):
             op = CallMethod(methodName: p.methodName, numArguments: inouts.count - 2)
+        case .callComputedMethod(_):
+            // We subtract 3 from the inouts count since the first two elements are the callee and method and the last element is the instruction index
```
saelo, line 596: I think this comment isn't correct. The first two are the reference object and the method, and the last element is the output variable (not the instruction index)

Sources/Fuzzilli/Minimization/CallArgumentReducer.swift
amarekano: That was a mistake on my part. I've corrected it now

Sources/Fuzzilli/Lifting/FuzzILLifter.swift
amarekano: done

Sources/Fuzzilli/FuzzIL/Operations.swift
saelo: The rough rule of thumbs is that every Operation class that has members other than those already in the `Operation` class is parametric. For example integer values (`LoadInteger`), string values (`LoadProperty` and `CallMethod`), or Arrays (`CallFunctionWithSpread`). Maybe a better definition is that an operation should be parametric if it can/should be mutated by the [OperationMutator](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Mutators/OperationMutator.swift). The `CallComputedMethod` class now no longer has any members and there is no obvious way in which the OperationMutator could meaningfully mutate it.

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -79,6 +79,8 @@ public class OperationMutator: BaseInstructionMutator {
             newOp = DeleteElement(index: b.genIndex())
         case let op as CallMethod:
             newOp = CallMethod(methodName: b.genMethodName(), numArguments: op.numArguments)
+        case let op as CallComputedMethod:
```
saelo, line 82: This no longer mutates anything, so let's remove this (together with the `.isParametric` flag of the operation)

Sources/Fuzzilli/FuzzIL/Operations.swift
amarekano: great, thanks for that explanation that's very helpful. Shall I add this explanation to the comments under Instruction.swift just so that we have this knowledge is captured in the source?

saelo: Yep nice, thanks!

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -512,6 +512,29 @@ class LifterTests: XCTestCase {
...
         XCTAssertEqual(lifted_program,expected_program)
     }
+
+    func testComputedMethodLifting(){
+        let fuzzer = makeMockFuzzer()
+        let b = fuzzer.makeBuilder()
+
+        let v0 = b.loadString("Hello World")
+        let v2 = b.loadBuiltin("Symbol")
+        let v3 = b.loadProperty("iterator", of: v2)
```
saelo, line 522: Nit: the variable numbers here don't match the ones in `expected_program` below

Sources/Fuzzilli/FuzzIL/Operations.swift
saelo: Now you just need to remove the `.isParametric` here and I'll merge it ;D

amarekano: woops! all done now :)

saelo:
Perfect, thanks!

amarekano:
and thank you for your patience with these reviews :D

PR #201, opened 2021-06-05
SpiderMonkey now supports fuzzilli by default. 

sploitem:
So there's no need of patches. Removed patches and added flag in to build script.


sploitem:
@googlebot I signed it!

Targets/Spidermonkey/fuzzbuild.sh
```diff
@@ -22,6 +22,6 @@ export CXX=clang++-10
 
 mkdir fuzzbuild_OPT.OBJ
 cd fuzzbuild_OPT.OBJ
-/bin/sh ../configure.in --enable-debug --enable-optimize --disable-shared-js
+/bin/sh ../configure.in --enable-debug --enable-optimize --disable-shared-js --enable-js-fuzzilli
 
-make -j
+make -j8
```
saelo, line 27: Is this change intentional?

saelo (commented):
Nice, thanks! Just one question

Targets/Spidermonkey/fuzzbuild.sh
sploitem: Yeah, i've noticed it forces my virtual machine to reset. Xubuntu 20.04 on Virtual Box.

saelo: Ok that's non ideal, limiting it to 8 will probably make building on the 72 core machine I'm using awfully slow :D How about querying the number of available CPUs, e.g. via `getconf _NPROCESSORS_ONLN` (seems to work on Linux and macOS) and maybe falling back to 8 if that returns an empty string?

sploitem: I get it :)
Sounds good.

sploitem:
Done (getconf _NPROCESSORS_ONLN)

saelo:
Thanks!

PR #203, opened 2021-06-10
Some fixes

sploitem:
Deleted REVISION file.
Removed step 2 in README.


saelo:
Thanks for cleaning things up. I think this change will break the Docker builder though: https://github.com/googleprojectzero/fuzzilli/blob/master/Cloud/Docker/SpidermonkeyBuilder/Dockerfile (I think it's probably already broken due to the removed Patches/ directory, my bad...). I think the easiest way to solve this is to just copy the [structure of the V8 directory](https://github.com/googleprojectzero/fuzzilli/tree/master/Targets/V8):

* Put the REVISION file back in (feel free to update to a newer commit though)
* Add the Patches/ directory back with just an empty .gitkeep
* Keep the changes to the README like they are now in this PR (i.e. just the three steps)

Alternatively, we can update the build.sh and the Dockerfile for the SpidermonkeyBuilder, whatever you prefer. Although I think keeping a Patches/ directory makes sense in any case, since that makes it easy to test e.g. custom engine instrumentation.

sploitem:
> Thanks for cleaning things up. I think this change will break the Docker builder though: https://github.com/googleprojectzero/fuzzilli/blob/master/Cloud/Docker/SpidermonkeyBuilder/Dockerfile (I think it's probably already broken due to the removed Patches/ directory, my bad...). I think the easiest way to solve this is to just copy the [structure of the V8 directory](https://github.com/googleprojectzero/fuzzilli/tree/master/Targets/V8):
> 
> * Put the REVISION file back in (feel free to update to a newer commit though)
> * Add the Patches/ directory back with just an empty .gitkeep
> * Keep the changes to the README like they are now in this PR (i.e. just the three steps)
> 
> Alternatively, we can update the build.sh and the Dockerfile for the SpidermonkeyBuilder, whatever you prefer. Although I think keeping a Patches/ directory makes sense in any case, since that makes it easy to test e.g. custom engine instrumentation.

Ok. I've restored patches and revision.

PR #204, opened 2021-06-11
Add assignment operators

amarekano:
This PR adds support for [assignment operations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#assignment_operators).

### DIFF
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -430,6 +430,28 @@ public struct AbstractInterpreter {
                 set(instr.output, .boolean)
             }
 
+        case let op as AssignmentOperation:
+            switch op.op {
+            case .Add:
+                set(instr.input(0), .primitive)
+            case .Sub,
+                 .Mul,
+                 .Exp,
+                 .Div,
+                 .Mod:
+                set(instr.input(0), .number | .bigint)
+            case .BitAnd,
+                 .BitOr,
+                 .Xor,
+                 .LShift,
+                 .RShift,
+                 .UnRShift:
+                set(instr.input(0), .integer | .bigint)
+            case .LogicAnd,
+                 .LogicOr:
+                set(instr.input(0), .boolean)
+            }
+
         case is TypeOf:
             set(instr.output, .string)
 
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -430,23 +430,23 @@ public struct AbstractInterpreter {
                 set(instr.output, .boolean)
             }
 
-        case let op as AssignmentOperation:
+        case let op as BinaryOperationAndReassign:
             switch op.op {
             case .Add:
-                set(instr.input(0), .primitive)
+                set(instr.input(0), maybeBigIntOr(.primitive))
             case .Sub,
                  .Mul,
                  .Exp,
                  .Div,
                  .Mod:
-                set(instr.input(0), .number | .bigint)
+                set(instr.input(0), maybeBigIntOr(.number))
             case .BitAnd,
                  .BitOr,
                  .Xor,
                  .LShift,
                  .RShift,
                  .UnRShift:
-                set(instr.input(0), .integer | .bigint)
+                set(instr.input(0), maybeBigIntOr(.integer))
             case .LogicAnd,
                  .LogicOr:
                 set(instr.input(0), .boolean)
--- a/Sources/Fuzzilli/FuzzIL/Operations.swift
+++ b/Sources/Fuzzilli/FuzzIL/Operations.swift
@@ -546,7 +546,7 @@ class BinaryOperation: Operation {
 }
 
 /// Assigns a value to its left operand based on the value of its right operand.
-class AssignmentOperation: Operation {
+class BinaryOperationAndReassign: Operation {
     let op: BinaryOperator
 
     init(_ op: BinaryOperator) {
--- a/Sources/Fuzzilli/FuzzIL/Operations.swift
+++ b/Sources/Fuzzilli/FuzzIL/Operations.swift
@@ -545,6 +545,16 @@ class BinaryOperation: Operation {
     }
 }
 
+/// Assigns a value to its left operand based on the value of its right operand.
+class AssignmentOperation: Operation {
+    let op: BinaryOperator
+
+    init(_ op: BinaryOperator) {
+        self.op = op
+        super.init(numInputs: 2, numOutputs: 0)
+    }
+}
+
 /// Duplicates a variable, essentially doing `output = input;`
 class Dup: Operation {
     init() {
--- a/Tests/FuzzilliTests/LifterTest.swift
+++ b/Tests/FuzzilliTests/LifterTest.swift
@@ -546,6 +546,35 @@ class LifterTests: XCTestCase {
         let v4 = b.compare(v2, v3, with: .greaterThan)
         let _ = b.conditional(v4, v2, v3)
 
+        let lifted_program = fuzzer.lifter.lift(program)
+        let expected_program = """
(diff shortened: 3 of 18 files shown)


Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -399,6 +399,28 @@ public struct AbstractInterpreter {
                 set(instr.output, .boolean)
             }
 
+        case let op as AssignmentOperation:
+            switch op.op {
+            case .Add:
+                set(instr.input(0), .primitive)
```
saelo, line 405: I made some changes to the AbstractInterpreter's handling of bigints in https://github.com/googleprojectzero/fuzzilli/commit/7dbaa342b2fe00da8f33c8d96c7b5eb460e24fdc please copy the handling of these from e.g. the BinaryOperation again :)

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -531,6 +531,16 @@ class BinaryOperation: Operation {
     }
 }
 
+/// Assigns a value to its left operand based on the value of its right operand.
+class AssignmentOperation: Operation {
```
saelo, line 535: I'm not 100% happy about this name since at first sight I might assume this is just a `x = y`, but that's exactly what this *isn't*. How about we change it to `BinaryOperationAndReassignment` or something like that? That seems a bit more intuitive, as this class uses a `BinaryOperator` internally but also performs what `Reassign` does. WDYT?

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -535,6 +535,41 @@ class LifterTests: XCTestCase {
 
         XCTAssertEqual(lifted_program,expected_program)
     }
+
+    func testAssignmentOperationLifting(){
```
saelo, line 539: Nice, thanks for adding tests! Can you please also extend [this](https://github.com/googleprojectzero/fuzzilli/blob/7dbaa342b2fe00da8f33c8d96c7b5eb460e24fdc/Tests/FuzzilliTests/InterpreterTest.swift#L606) test? (Just add another loop and pretty much copy everything else from the BinaryOperation case)

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -535,6 +535,41 @@ class LifterTests: XCTestCase {
...
+        let fuzzer = makeMockFuzzer()
+        let b = fuzzer.makeBuilder()
+
+        let v0 = b.loadInt(1337)
+        let v1 = b.loadFloat(13.37)
+        b.reassign(v0, to: v1)
```
saelo, line 545: Maybe drop this `reassign` after fixing the `VariableAnalyzer` (see main comment)

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -535,6 +535,41 @@ class LifterTests: XCTestCase {
...
+        b.assign(v0, to: v1, with: .Add)
+        b.assign(v0, to: v1, with: .Mul)
+        b.assign(v0, to: v1, with: .LShift)
+
+        let v2 = b.loadString("")
+        let v3 = b.loadString("hello")
+        b.reassign(v2, to: v3)
```
saelo, line 552: Same here

saelo (commented):
Thanks! Apart from the comments, I think you also need to add this new operation in [Semantics.swift](https://github.com/googleprojectzero/fuzzilli/blob/7dbaa342b2fe00da8f33c8d96c7b5eb460e24fdc/Sources/Fuzzilli/FuzzIL/Semantics.swift#L41), and maybe a few other places (maybe quickly check where `Reassign` is referenced. This new instruction needs to be handled similarly). This is e.g. necessary for [this](https://github.com/googleprojectzero/fuzzilli/blob/7dbaa342b2fe00da8f33c8d96c7b5eb460e24fdc/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L37) analyzer to work correctly, which in turn is necessary for the JavaScriptLifter to emit a `let` instead of a `const`. I think currently, you could end up with
```
const v0 = 42;
v0 += 43;
```
We should probably have a test for this, but unsure how best to test it. Maybe just a `VariableAnalyzerTest` in LifterTest.swift that verifies that all variables that are reassigned through either

* a .PostInc and .PreInc UnaryOperation,
* a `Reassign` instruction, or
* a `BinaryOperationAndReassign` instruction

are correctly [marked to have > 1 assignment](https://github.com/googleprojectzero/fuzzilli/blob/7dbaa342b2fe00da8f33c8d96c7b5eb460e24fdc/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L92) by the analyzer and are then lifted to `let` instead of `const`? WDYT?

Sources/Fuzzilli/FuzzIL/Operations.swift
amarekano: Tbf this was the hardest part of the PR: finding a name for the operation that caused least confusion :D
I like your idea, could we perhaps shorten it to `BinaryOperationAndReassign`?

saelo: Sounds good, let's go with that!

amarekano:
I've addressed you review comments and added testcases as requested. I've noticed that that some of the testcases you've added like [`testBigintTypeTracking`](https://github.com/googleprojectzero/fuzzilli/blob/87ee22bad2ac3bafbc8db13d5bfb7ea3c25fc39c/Tests/FuzzilliTests/InterpreterTest.swift#L606) and [`testTypeInstantiation`](https://github.com/googleprojectzero/fuzzilli/blob/87ee22bad2ac3bafbc8db13d5bfb7ea3c25fc39c/Tests/FuzzilliTests/ProgramBuilderTest.swift#L144) are missing from [XCTestManifests](https://github.com/googleprojectzero/fuzzilli/blob/87ee22bad2ac3bafbc8db13d5bfb7ea3c25fc39c/Tests/FuzzilliTests/XCTestManifests.swift#L26). I've added them manually to the manifest but I'm unsure why they aren't being auto generated. Perhaps it isn't an issue on macOS since it runs all tests regardless of whether they have been added to the manifest but on *nix the testcases must be in the manifest for them to be executed.

saelo:
Ah yeah great, thank you! Yeah I most likely just forgot to run `swift test --generate-linuxmain`, although for me that now prints `warning: '--generate-linuxmain' option is deprecated; tests are automatically discovered on all platforms` so it seems like we no longer need the `XCTestManifests.swift` and `LinuxMain.swift` at all?

amarekano:
> now prints warning: '--generate-linuxmain' option is deprecated; tests are automatically discovered on all platforms so it seems like we no longer need the XCTestManifests.swift and LinuxMain.swift at all?

Yes, I get the same warning when I build on Ubuntu 20.04. I've been manually adding my testcases to the manifest to keep the test suite in sync. I think we still need the manifest or people (like me) building on *nix won't be able to run the entire test suite. I'm not sure if there's a way to refactor this to get rid of the manifest and LinuxMain.

saelo:
So I just tried removing those two files and building the [Docker container](https://github.com/googleprojectzero/fuzzilli/blob/master/Cloud/Docker/FuzzilliBuilder/Dockerfile) which [seems to be ubuntu 18.04 and swift 5.4.1](https://github.com/apple/swift-docker/blob/29bb1fc9226ef199e979bbad55d89d088615d513/5.4/ubuntu/18.04/Dockerfile) and there it appears to run the tests just fine without those files: `Executed 91 tests, with 0 failures (0 unexpected) in 6.276 (6.276) seconds`. Does that work for you as well?

amarekano:
Nice! Yeah that works at my end to (Ubuntu 20.04 and swift 5.4.1) and it executes 93 test cases successfully. Presumably your build is a commit behind main.

saelo:
Great, I'll just remove those files then, thanks for checking! And yes, my build was just behind a bit :)

PR #207, opened 2021-06-15
Add support for switch statements

amarekano:
This PR adds support for JS switch statements.

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -413,22 +413,18 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
-        b.beginSwitch(cond){
-            b.generateRecursive()
-        }
-
-        var caseStmts = [Variable]()
-        // Generate 0 or more switch cases
-        for _ in 0..<Int.random(in: 0...5) {
-            let caseStmt = b.randVar(ofType: b.type(of: cond)) ?? b.randVar()
-            if !caseStmts.contains(caseStmt) {
-                caseStmts.append(caseStmt)
-                b.beginSwitchCase(caseStmt) {
+        b.beginSwitch(on: cond) { cases in
+            cases.addDefault {
+                b.generateRecursive()
+            }
+            for _ in 0..<Int.random(in: 0...5) {
+                var fallsThrough = false
+                withProbability(0.1, do: {fallsThrough = true})
+                cases.add(b.randVar(), fallsThrough: fallsThrough) {
                     b.generateRecursive()
                 }
             }
         }
-        b.endSwitch()
     },
 
     CodeGenerator("WhileLoopGenerator") { b in
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -412,6 +412,26 @@ public let CodeGenerators: [CodeGenerator] = [
         b.endIf()
     },
 
+    CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
+        b.beginSwitch(cond)
+
+        var caseStmts = [Variable]()
+        // Generate 0 or more switch cases
+        for _ in 0..<Int.random(in: 0...5) {
+            let caseStmt = b.randVar(ofType: b.type(of: cond)) ?? b.randVar()
+            if !caseStmts.contains(caseStmt) {
+                caseStmts.append(caseStmt)
+                b.beginSwitchCase(caseStmt) {
+                    b.generateRecursive()
+                }
+            }
+        }
+        // End switch will always generate a default case
+        b.endSwitch {
+            b.generateRecursive()
+        }
+    },
+
     CodeGenerator("WhileLoopGenerator") { b in
         let loopVar = b.reuseOrLoadInt(0)
         let end = b.reuseOrLoadInt(Int64.random(in: 0...10))
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -413,14 +413,12 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
-        b.beginSwitch(on: cond) { cases in
+        b.doSwitch(on: cond) { cases in
             cases.addDefault {
                 b.generateRecursive()
             }
             for _ in 0..<Int.random(in: 0...5) {
-                var fallsThrough = false
-                withProbability(0.1, do: {fallsThrough = true})
-                cases.add(b.randVar(), fallsThrough: fallsThrough) {
+                cases.add(b.randVar(), fallsThrough: probability(0.1)) {
                     b.generateRecursive()
                 }
             }
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -413,7 +413,9 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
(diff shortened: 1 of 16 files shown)


saelo (commented):
Looks great, thanks! Two thoughts:

1) Should we allow fallthrough between cases? I think the way to do that would be to add a parameter `let fallthrough: Bool` to the `BeginSwitchCase` operation, which, if true, causes the *preceding* case to fall through to it (and so no "break;" is emitted by the Lifter). Then the CodeGenerator can just set a few cases (how about, say, 10% or so?) to fall through.

2) I think the ProgramBuilder API can be a bit more intuitive. How about something like this, inspired by the [ClassBuilder](https://github.com/googleprojectzero/fuzzilli/blob/87ee22bad2ac3bafbc8db13d5bfb7ea3c25fc39c/Sources/Fuzzilli/Core/ProgramBuilder.swift#L1210):

```swift
b.switch(on: v3) { cases in
    cases.addDefault {
        b.generateRecursive()
    }
    for _ in 0..<Int.random(in: 0...5) {
        cases.add(b.randVar(), previousCaseFallsThrough: p(0.1)) {
            b.generateRecursive()
        }
    }
}
```

In that case, the ProgramBuilder side would probably look something like this:
```swift
struct SwitchBuilder {
    public typealias SwitchCaseGenerator = () -> ()
    fileprivate var defaultCaseGenerator: SwitchCaseGenerator? = nil
    fileprivate var caseGenerators: [(value: Variable, fallsthrough: Bool, body: SwitchCaseGenerator)] = []

    func addDefault(_ generator: SwitchCaseGenerator) { defaultCaseGenerator = body; }
    func add(_ v: Variable, previousCaseFallsThrough: Bool, body: SwitchCaseGenerator) {
        ...
    }
}
public func switch(on v: Variable, body: inout (SwitchBuilder) -> () {
    var builder = SwitchBuilder()
    body(&builder)

    perform(BeginSwitchCase(), withInputs: [v])
    builder.defaultCaseGenerator?()
    for (val, fallsthrough, bodyGenerator) in builder.cases {
        ...;
    }
    perform(EndSwitch())
}
```

WDYT?

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -754,6 +754,30 @@ class EndIf: ControlFlowOperation {
...
+    }
+}
+
+class BeginSwitchCase: ControlFlowOperation {
+    init() {
+        super.init(numInputs: 1, attributes: [.isBlockBegin, .isBlockEnd])
+    }
+}
+
+class EndSwitchCase: ControlFlowOperation {
```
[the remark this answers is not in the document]
amarekano: my bad, that was leftover code from a previous implementation that I was trying out. I'll remove this.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -402,6 +402,25 @@ public let CodeGenerators: [CodeGenerator] = [
...
 
+    CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
+        b.beginSwitch(cond){
+            b.generateRecursive()
+        }
+
+        var caseStmts = [Variable]()
+        // Generate 0 or more switch cases
+        for _ in 0..<Int.random(in: 0...5) {
+            let caseStmt = b.randVar(ofType: b.type(of: cond)) ?? b.randVar()
```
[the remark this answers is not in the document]
amarekano: An alternative that I can think of is:
1. Add a check to prevent `cond` from being added to the case `caseStmts`. So that should prevent code like the one above from being generated.
2. If `caseStmts` is empty and we haven't generated any case statments, then add `cond` to the `caseStmts` and generate atleast one case statement.
WDYT?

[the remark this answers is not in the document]
saelo: Right ok, sounds good. Yeah you're right, you'll have to deal with the (rare) scenario of `cond` being the only variable available in the program. Maybe you can do something like:
```
var caseValue: Variable
if let v = b.randVar(ofType: b.type(of: cond)), v != cond, !cases.contains(v) {
    caseValue = v
} else {
    caseValue = b.randVar()
}
cases.insert(caseValue)
```
That should achieve more or less the same I think. WDYT?
I still think it'd be ok to just `b.randVar()` for all cases, but I'm definitely fine either way!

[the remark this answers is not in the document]
amarekano: Let's keep it simple for now and just go with randVar we can always revisit it later if we need to introduce special handling for the rare case :)

amarekano:
The SwitchBuilder approach was a great idea and I definitely appreciate the cleaner code :)
I've addressed all your review comments but do take another look and let me know if there's anything else you'd like me to change.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -412,6 +412,21 @@ public let CodeGenerators: [CodeGenerator] = [
         b.endIf()
     },
 
+    CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
+        b.beginSwitch(on: cond) { cases in
+            cases.addDefault {
+                b.generateRecursive()
+            }
+            for _ in 0..<Int.random(in: 0...5) {
+                var fallsThrough = false
```
saelo, line 421: Can this just be `let fallsThrough = probability(0.1)`?

saelo: Or I guess even just
`cases.add(b.randVar(), fallsThrough: probability(0.1)) {`
below

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1301,6 +1301,30 @@ public class ProgramBuilder {
         perform(EndIf())
     }
 
+    public struct SwitchBuilder {
+        public typealias SwitchCaseGenerator = () -> ()
+        fileprivate var defaultCaseGenerator: SwitchCaseGenerator? = nil
+        fileprivate var caseGenerators: [(value: Variable, fallsthrough: Bool, body: SwitchCaseGenerator)] = []
+
+        public mutating func addDefault(_ generator: @escaping SwitchCaseGenerator) { defaultCaseGenerator = generator }
```
saelo, line 1309: Let's add an `assert(defaultCaseGenerator == nil)` here

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1301,6 +1301,30 @@ public class ProgramBuilder {
...
+        public mutating func add(_ v: Variable, fallsThrough: Bool = false, body: @escaping SwitchCaseGenerator) {
+            caseGenerators.append((v, fallsThrough, body))
```
saelo, line 1311: Let's add a `assert(defaultCaseGenerator != nil)` here with a comment like "Default case must be generated first due to the way FuzzIL represents switch statements".

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1301,6 +1301,30 @@ public class ProgramBuilder {
...
+        }
+    }
+
+    public func beginSwitch(on v: Variable, body: (inout SwitchBuilder) -> ()) {
```
saelo, line 1315: Nit: maybe just call this `switch` since there is now no longer and `endSwitch` or similar?

amarekano: the compiler complains that `switch` is a keyword and if must use it then I need to wrap it in backticks. Thought it's just cleaner to use `beginSwitch` instead but I'm happy to change it :)

saelo: Ah, I see, how about `doSwitch`? That'd be consistent with other methods, e.g. [doBreak](https://github.com/googleprojectzero/fuzzilli/blob/cb05c8f8205e9f175956005907e856360e5bcc2e/Sources/Fuzzilli/Core/ProgramBuilder.swift#L1334)

amarekano:
all done! I missed adding a generator weight from #204 which I've included in this PR

saelo:
Nice, thanks!

PR #214, opened 2021-06-23
Refactored Fuzzer instance synchronization

saelo:


saelo:
@WilliamParks Please take a look :)

WilliamParks:
Great! It's going to be a couple days before I can check it out. Do you know if it fixes the issue with the log error you were getting?

saelo:
Cool yeah no worries!
I don't think this will fix the `Failed to find edge in map` warning, at least I didn't look into that any closer. If your suspicion is correct and it's some kind of race between the minimizer and `regenProgramList()`, then probably it's fine and we can just make a comment about that. But if you have some idea about how best to debug this (i.e. add some checks/assert somewhere) then I'm happy to do another run with it.

saelo:
Hm so I still have the root instance from the weekend fuzzing session running with ~24k total interesting samples found and around 26% coverage on JSC. When it now regenerates the list, it prints `[MarkovCorpus] Failed to find edge in map` easily a couple hundred times before printing `[MarkovCorpus] Markov Corpus selected 1756 new programs`. So I don't think it would be such a race condition :/

WilliamParks:
Hmm I think we're leaking an edge somewhere, with a new edge non-deterministically being found and not properly reset. Maybe during minimization or an early exit during the determinism checks?

WilliamParks:
Looking through the code, I don't think the issue is with the minimizer, as it uses the compare coverage function of libcoverage, rather than generating new ProgramAspects.

I'll look for other places that generate aspects.

WilliamParks:
Looks good to me!

As a thought, would it make sense to compare the performance of instances that restart periodically, with ones that are consistently running? The performance improvement comes from being able to target infrequently hit edges, and longer running instances develop a better sense of which edges are hit more/less frequently. A Program in the corpus could have its unique edges also triggered by a commonly used mutation (such as calling a builtin), and a long running instance would thus prioritize that sample less.

saelo:
Nice, thanks for taking a look! I'll let it run with this new code for a bit and see if there are any issues.

It could make sense to run such an experiment, but I think it's probably not too important. It's should be fairly clear that performance will suffer somewhat, and that's also the case for the BasicCorpus (mostly because the `age` vector isn't transferred there). Maybe the effect isn't even that large, since the edge counts are now already computed when importing the corpus, so the worker instance should have somewhat decent counts right away. On the other hand, preemtible cloud instances are significantly cheaper, so it's a trade-off.

WilliamParks:
I just triggered `Corpus must not be empty for fuzzing and we failed to import the seed program. Is the evaluator working correctly?` when using `--importCorpusNewCov`. I'm thinking any coverage from  `makeSeedProgram` is covered by programs in the corpus which have been run, but not yet minimized.

What do you think of moving the call to `ensureCorpusIsPopulated` to prior to importing the corpus?

saelo:
Ah yeah, that's because the import hasn't finished and so the corpus is still empty, but we also have evaluated all these samples and so the seed program doesn't trigger any new coverage. I think we should just wait until the corpus import has actually finished. Can you check if https://github.com/googleprojectzero/fuzzilli/commit/bc057d1eb2d47878b700022e59409e974bd01597 fixes the issue? It seems to work fine for me now

WilliamParks:
That works for me as well. Thanks!

PR #216, opened 2021-06-25
simplify switch statement generation

amarekano:
This PR attempts to simplify switch statement generation by getting rid of the `fallsThrough` flag and replaces the same functionality with `doBreak` calls in the CodeGenerator.


saelo:
So I think we should keep it the way it is now. Here's why:

Assume you have a sample with a Switch in it:
```javascript
switch (v7) {
    default:
        // do stuff
        break;
    case X:
        // do stuff
        break;
    case Y:
        // do stuff
        break;
    case Z:
        // do stuff
        break;
}
```
Now if you minimize that sample, and the Switch statement and the code in case Y is relevant for that sample, the sample will probably become
```javascript
switch (v7) {
    default:
    case X:
    case Y:
        // do stuff
    case Z:
}
````
Because the minimizer will remove everything in the other cases, then determine that all those `Break` instructions are also not needed and so remove those as well. But at that point, the entire Switch statement is pretty useless since basically any newly inserted code is always executed, and there's essentially no way to "recover" it again. Basically, the rule of thumb is that the minimizer should only be able to remove code structures that a CodeGenerator can somewhat easily add back later on, but that's not the case here (it's highly unlikely for the [BreakGenerator](https://github.com/googleprojectzero/fuzzilli/blob/f31ba8ee5f1d7d5f338a0d047a55d21b13494f7a/Sources/Fuzzilli/Core/CodeGenerators.swift#L481) to insert Break operations into the right places).

So long-story short, I think it's better to encode whether a case should fall through in the Switch operations, as is currently the case. Does that make sense?

**Comment by amarekano** [comment]
Yeah that makes sense and something that I hadn't considered. However, there's another challenge that I've run into due to the design of this op that makes parsing switch statements problematic and this is mainly due to two issues described below. Consider [this testcase](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/bytecode-cache-cached-string-impl.js) that we're trying to compile to FuzzIL:

```
 switch ('foo') {
        case "foo":
            break;
        case "bar":
        case "baz":
        default:
            throw new Error('Should have jumped to `foo`');
            break;
    }
```

In our current design we would first need construct a `BeginSwitch` and that would required to loop over the list of cases to locate the `default` case and hoist it to the top of the switch block. Here lies the first issue, hoisting `default` in this way changes the semantics of the switch statement. We can get away with this in some situations if two conditions are met:

1. the case that precedes `default` ends with a `break` statement.
2. the `default` case itself ends with a `break`. 

The second issue is parsing a case block. In the example above, if we did a simple linear sweep of the statements and generate FuzzIL using the current op implementation, we would end up with double `Fuzzilli.Break` statements:

```
BeginSwitch foo
    DefaultCase
        v6 <- Construct v5, [v4]
        ThrowException v6
        Break	// Generated by compiling js break; to Fuzzilli.Break
        Break	// Emitted by the lifter from setting the fallsThrough flag to false in the next BeginSwitchCase.
    BeginSwitchCase foo
        Break	 
        Break
    BeginSwitchCase bar
    BeginSwitchCase baz
    EndSwitch
```

Due to this, a simple linear sweep doesn't work and we now need to do more to track the state of the of the previous block that's been parsed. More specifically, we need to record the fuzzilli op the last js statement in the previous case block generated. If the last op generated at the end of a case block is a `Fuzzilli.Break` statement we would need to disable the `fallsThrough` flag in the next `BeginSwitchCase` op AND remove the `Fuzzilli.Break` operation that's been generated.

This state tracking is unnecessarily complex and this PR was meant to solve the second problem and make parsing easier. If we drop the `fallsThrough` flag then linear sweep parsing becomes simple and there's no state dependency between case blocks. The issue you've raised on minimisation is valid and I think we can solve this with a custom block reducer similar to how we do it for try-catch-finally blocks (I'll add commits for a custom reducer if we decide to go down this path.)

Having said that the PR does not solve the first issue (`default` hoisting). I'm still trying to think of an implementation that plays nice with our AI and I'd appreciate your thoughts on how best to resolve these two issues.

**Comment by saelo** [comment]
Nice, thanks for the detailed explanation, definitely makes sense!

So for issue (1) currently the best option I can think of would be to add a 2nd, optional input to `BeginSwitch` and make the inputs to `BeginSwitchCase` optional as well. Then, the 2nd input to `BeginSwitch` is the value for the first `case`, and the case that is missing an input is the default case (there can then be a `BeginSwitchCase.isDefaultCase` and `BeginSwitch.firstCaseIsDefaultCase` getter which simply returns `numInputs == 0` and `numInputs == 1` respectively to make this clear). That way, we would avoid introducing additional complexities from e.g. having to ensure the first instruction following a `BeginSwitch` is a `BeginSwitchCase` or something like that or having to teach the AbstractInterpreter about that, and we also don't require any additional Operations. We might want to verify that there is always at most one default branch though, probably in [`Code.check()`](https://github.com/googleprojectzero/fuzzilli/blob/f31ba8ee5f1d7d5f338a0d047a55d21b13494f7a/Sources/Fuzzilli/FuzzIL/Code.swift#L162). This should also work nicely with the current `SwitchBuilder`: the `addDefault` method now also just appends to the `casesGenerators` array, but without an input variable.

For issue (2), I still think it'd be better to keep the fallsThrough flag in the `BeginSwitchCase` operations, since otherwise you'll need to add similar state tracking to the [GenericInstructionReducer](https://github.com/googleprojectzero/fuzzilli/blob/main/Sources/Fuzzilli/Minimization/GenericInstructionReducer.swift) which would otherwise remove the `Break` instructions if they aren't relevant to the discovered coverage. I'd like for that reducer to stay as simple as possible. Can we instead do the following:
* When the compiler finds a JS `break` statement, it emits a Fuzzilli.Break operation
* When the compiler finds a JS `case` begin, and the last emitted instruction was a Break, then it sets the fallsThrough flag to false

Now this will, as you described, leave us with two consecutive break statements when lifting that FuzzIL code back to JavaScript. I think that's Ok though, since it's syntactically valid, semantically the same as a single Break, and will be "fixed" as soon as the minimizer runs as it will remove the unnecessary Break instruction. If we really care about not having those double-Break code, we could still add a simple post processing step to the compiler which runs after compilation is finished and detects these double-Break patterns and resolves them. Such a post processing pass might be useful for other things as well.

As a side node, maybe we should then also teach the [`BreakGenerator`](https://github.com/googleprojectzero/fuzzilli/blob/f31ba8ee5f1d7d5f338a0d047a55d21b13494f7a/Sources/Fuzzilli/Core/CodeGenerators.swift#L481) to generate Break instructions inside Switch cases by adding a new context type?

WDYT?

**Comment by amarekano** [comment]
Thanks for suggestions, I've incorporated the optional parameters approach to solving issue (1). I took some liberties with the code but I think there may be room to make it more concise. Some considerations:
> We might want to verify that there is always at most one default branch though, probably in Code.check()

I've attempted to do this in the CodeGenerator itself using the `hasDefault` flag. I don't think we should include this check in `Code.check` since this will cause imported samples that don't have default cases in their switch blocks to be rejected.

> This should also work nicely with the current SwitchBuilder: the addDefault method now also just appends to the casesGenerators array, but without an input variable.

I've gone off script here and renamed `addDefault` to `addFirstCase` since the first case need not be the default case.

Wrt issue (2), I've reverted the changes back to using `fallsThrough`. You're right about the double break statements not altering the semantics of the program and I agree that if we do want to clean this up, then we should offload that to a post-compilation function.

One thing to note however with the double `break;` statements and using the flag is that when we end up with two `break;` statements in the lifted code there will be just one Fuzzilli.Break operation (generated from parsing a sample). The second `break` statement which is emitted by the lifter due to the flag state won't be reduced away and we would have to toggle the flag state to achieve the same effect. I'm not sure if that's the intention here.

> As a side node, maybe we should then also teach the BreakGenerator to generate Break instructions inside Switch cases by adding a new context type?

I've added a `switchCase` context to support this, let me know what you think.

File: Sources/Fuzzilli/Core/ProgramBuilder.swift [hunk 1]
```diff
@@ -1317,28 +1317,28 @@ public class ProgramBuilder {
...
         public typealias SwitchCaseGenerator = () -> ()
-        fileprivate var defaultCaseGenerator: SwitchCaseGenerator? = nil
-        fileprivate var caseGenerators: [(value: Variable, fallsthrough: Bool, body: SwitchCaseGenerator)] = []
+        fileprivate var firstCaseGenerator: SwitchCaseGenerator? = nil
+        fileprivate var caseGenerators: [(value: [Variable], fallsthrough: Bool, body: SwitchCaseGenerator)] = []
 
-        public mutating func addDefault(_ generator: @escaping SwitchCaseGenerator) {
-            assert(defaultCaseGenerator == nil)
-            defaultCaseGenerator = generator
+        public mutating func addFirstCase(_ generator: @escaping SwitchCaseGenerator) {
```
saelo, line 1323: Do we still need this at all? It feels a bit weird, and I guess even an empty Switch statement (`switch (x) { }`) is valid JavaScript, so I don't think we need to handle the first case differently from the others. How about we do it like this:
```swift
public mutating func addDefault(fallsThroughToFollowingCase fallsThrough: Bool = false, body: @escaping SwitchCaseGenerator) {
    caseGenerators.append((nil, fallsThrough, body))
}

public mutating func add(_ v: Variable, fallsThroughToFollowingCase fallsThrough: Bool = false, body: @escaping SwitchCaseGenerator) {
    caseGenerators.append((v, fallsThrough, body))
}
```

That should make for a fairly good API.

Then it's also obvious that nothing can fallthrough to the first case, and whether the last case fallsthrough or not doesn't matter. Now, the `doSwitch` method just needs to do a bit more work by tracking whether the _previous_ case falls through when emitting the current case, but I think overall it will be simpler and more intuitive.

Sources/Fuzzilli/FuzzIL/Analyzer.swift
```diff
@@ -187,6 +190,11 @@ struct ContextAnalyzer: Analyzer {
         } else if instr.op is BeginClassDefinition {
             // We are no longer in the previous context
             contextStack.append([.classDefinition, .function])
+        } else if instr.op is BeginSwitch {
+            contextStack.append([context, .switchCase])
+        } else if instr.op is BeginSwitchCase {
```
saelo, line 195: Is this case even necessary? At this point, the top context must be `[prevContext, .switchCase]`, and so this doesn't really do anything. I think it can just be removed

Sources/Fuzzilli/Lifting/FuzzILLifter.swift
```diff
@@ -233,17 +233,25 @@ public class FuzzILLifter: Lifter {
             w.decreaseIndentionLevel()
             w.emit("EndIf")
 
-        case is BeginSwitch:
+        case let op as BeginSwitch:
```
saelo, line 236: Thinking about this, I think I'd prefer if this was just something like
```swift
case is BeginSwitch:
    w.emit("BeginSwitch \(input(0))\(op.firstCaseIsDefaultCase ? "" : input(1))")
    w.increaseIndentionLevel()
```

Similar to how e.g. `BeginClassDefinition` is lifted. That way, the textual representation of FuzzIL stays as close to the "real" representation as possible. Otherwise, it might seem like there are multiple instructions, when in reality, it's just one.

Sources/Fuzzilli/Lifting/FuzzILLifter.swift
```diff
@@ -233,17 +233,25 @@ public class FuzzILLifter: Lifter {
...
+            }
             w.increaseIndentionLevel()
 
         case let op as BeginSwitchCase:
             if !op.fallsThrough {
                 w.emit ("Break")
             }
             w.decreaseIndentionLevel()
-            w.emit("BeginSwitchCase \(input(0))")
+            if op.isDefaultCase {
```
saelo, line 250: Same here, I think this should just be one line with zero or one inputs and maybe a note if the previous case falls through or not. E.g. `BeginSwitchCase v4 previousCaseFallsThrough` for a "normal" case with fallthrough or `BeginSwitchCase` for a default case with no fallthrough

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -802,20 +802,31 @@ class EndIf: ControlFlowOperation {
...
+    }
+
+    init(numArguments: Int) {
+        super.init(numInputs: numArguments, attributes: [.isBlockBegin])
     }
 }
 
 class BeginSwitchCase: ControlFlowOperation {
     /// If true, causes the preceding case to fall through to it (and so no "break;" is emitted by the Lifter)
     let fallsThrough: Bool
```
saelo, line 821: Should we call this `previousCaseFallsThrough` to make it more clear what this means?

Sources/Fuzzilli/FuzzIL/Analyzer.swift
amarekano: I think you mean the `BeginSwitchCase` entry. Yeah that seems right, I'll remove it

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: I've made these changes to `SwitchBuilder` and updated `doSwitch` to handle the first case. Not entirely happy with the way it's currently written but I can't think of a simpler alternative. Could you give it a look and let me know your thoughts?

Tests/FuzzilliTests/InterpreterTest.swift
```diff
@@ -670,7 +670,7 @@ class AbstractInterpreterTests: XCTestCase {
         let v4 = b.loadString("42")
 
         b.doSwitch(on: v2){ cases in
-            cases.addDefault {
+            cases.addDefault() {
```
saelo, line 673: Nit: are the `()` required here and below?

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -427,12 +427,24 @@ public let CodeGenerators: [CodeGenerator] = [
...
+                        b.generateRecursive()
+                    }
+                    hasDefault = true
+                } else {
+                    cases.add(b.randVar(), previousCaseFallsThrough: probability(0.1)) {
+                        b.generateRecursive()
+                    }
+                }
+            }
+            // If we still don't have a default case, then add one
```
saelo, line 445: Do we  really need this? I think it could also make sense to sometimes generate switch statements without a default case?

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -478,8 +490,8 @@ public let CodeGenerators: [CodeGenerator] = [
         }
     },
 
-    CodeGenerator("BreakGenerator", inContext: .loop) { b in
-        assert(b.context.contains(.loop))
+    CodeGenerator("BreakGenerator", inContext: ProgramContext([.loop, .switchCase])) { b in
```
saelo, line 493: I think this is incorrect, because it'll be interpreted as "must be in the .loop **and** the .switchCase" context (while we'd want "or"): https://github.com/googleprojectzero/fuzzilli/blob/10d5db4700b2d02481b916302269025cee9a1f69/Sources/Fuzzilli/Core/ProgramBuilder.swift#L834

I think the easiest way to "fix" this for now is just to duplicate this code generator and call one of them `LoopBreakGenerator` and the other `SwitchCaseBreakGenerator` maybe? Otherwise, it'd probably need a couple more internal changes...

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1335,10 +1332,17 @@ public class ProgramBuilder {
         var builder = SwitchBuilder()
         body(&builder)
 
-        perform(BeginSwitch(), withInputs: [v])
-        builder.defaultCaseGenerator?()
+        var handleBeginSwitch: Bool = true
+        
         for (val, fallsThrough, bodyGenerator) in builder.caseGenerators {
-            perform(BeginSwitchCase(fallsThrough: fallsThrough), withInputs: [val])
+            if handleBeginSwitch {
```
saelo, line 1338: So IIUC, right now if `builder.caseGenerators.isEmpty`, this would never run and so this function would only emit a `EndSwitch` instruction, resulting in invalid code. That's probably not great.
I guess there are broadly two options: (1) force the caller to add at least one case and add a `precondition` here that that's the case, or (2) if no cases were provided, create an empty `default` case. I'm fine both ways, maybe slightly preferring (1).

Also, nit, I think it'd might be be a bit nicer if the `BeginSwitch` was handled separately, outside the loop, and so the loop only did `for (val, fallsThrough, bodyGenerator) in builder.caseGenerators.dropFirst() {` (you should be able to use tuple destructuring syntax to get the pieces: `let (val, _, generator) = builder.caseGenerators.first!`)

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Sounds good! I'll add a comment so that we can look into improving this at a later stage

saelo:
Related issue: https://github.com/googleprojectzero/fuzzilli/issues/233 let me know if you have any thoughts on that! :)

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Yeah that works for me, I think from a previous discussion we always wanted a default case to be generated but I can remove this.

amarekano:
Hey @saelo when you get a chance do let me know if there's any additional changes you'd like to see with this PR. I think the changes to the `InputMutator` are the only pending items that need review and we should be good to merge.

saelo:
Yep, sorry! So two questions:
- with https://github.com/googleprojectzero/fuzzilli/pull/256 would there be a benefit in modeling switch-case similar to the Object creation as proposed in https://github.com/googleprojectzero/fuzzilli/issues/253? i.e. with a `BeginSwitch` and `EndSwitch` in between which we only allow `BeginSwitchCase` and `BeginDefaultCase`?
- Do you mind splitting this PR into two, one which changes the visibility of variables at Block ends (the `InputMutator` changes) and one which does the switch-case stuff? I think that'd be a bit easier

Thanks!

amarekano:
> with Splicing instructions is now context aware #256 would there be a benefit in modeling switch-case similar to the Object creation as proposed in Enhance Object creation #253? i.e. with a BeginSwitch and EndSwitch in between which we only allow BeginSwitchCase and BeginDefaultCase?

I think there are a couple of benefits to doing this: `BeginSwitchCase` and `BeginDefaultCase` would now declare a required context and the `InputMutator` can use this information to determine which valid scopes to retrieve compatible variables from. The other reason were re-modelling may be beneficial is improving code readability and consistency (e.g. we can do away with the logic to determine the number of input vars). I'd probably favour refactoring, do you have a preference?  

> Do you mind splitting this PR into two, one which changes the visibility of variables at Block ends (the InputMutator changes) and one which does the switch-case stuff? I think that'd be a bit easier 

Can do, I'll create one for the InputMutator changes in a bit

saelo:
Cool, yeah I think now that we'll anyway have the necessary logic to support it, I'm also happy to go with this refactored version, since it is probably a bit more intuitive. We might need some special handling in the `BlockMinimizer` though, so that it doesn't remove single cases from the Switch blocks, only the entire SwitchCase (because we probably won't have a CodeGenerator that adds cases to existing SwitchCase blocks).

amarekano:
I've tried to model Switch statements similar to #253 but it turns out that AI tracking becomes a bit more complex. Consider the following representation:
```
BeginSwitch
BeginSwtichCase
.... instructions
BeginDefault
... instructions
EndSwitch
```

When the AI encounters `BeginSwitch` it creates a new state but we don't insert instructions between `BeginSwitch` and `BeginSwitchCase` so that leaves us with this empty state that we need to handle. My initial thoughts on this would be to ignore `BeginSwitch` and `EndSwitch` ops in the AI and just parse blocks that begin with `BeginSwitchCase` and `BeginDefaultCase`. Not an approach I'm very keen on :(

I'm not sure if there's a simple way to handle this in the AI. I now think that it's probably easier to live with the existing modelling (similar to how we model class definitions). WDYT?

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -462,13 +462,24 @@ public let CodeGenerators: [CodeGenerator] = [
...
+        b.doSwitch(on: b.randVar()) { cases in
+            var caseVariables: [Variable] = []
+            for _ in 0..<Int.random(in: 1...5) {
+                if !hasDefault && probability(0.5) {
+                    cases.addDefault(previousCaseFallsThrough: probability(0.1)) {
+                        b.generateRecursive()
+                    }
+                    hasDefault = true
+                } else {
+                    let v = b.randVarInternal(excludeInnermostScope: true) ?? b.generateVariable(ofType: .object())
```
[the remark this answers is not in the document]
amarekano: @saelo any thoughts on this one? I think this is probably the last piece that needs to be resolved before this PR is good to go, unless you have other changes that you'd like me to make.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -912,20 +912,29 @@ class EndIf: ControlFlowOperation {
...
 
 class BeginSwitchCase: ControlFlowOperation {
     /// If true, causes the preceding case to fall through to it (and so no "break;" is emitted by the Lifter)
-    let fallsThrough: Bool
+    let previousCaseFallsThrough: Bool
 
-    init(fallsThrough: Bool) {
-        self.fallsThrough = fallsThrough
-        super.init(numInputs: 1, attributes: [.isBlockBegin, .isBlockEnd])
+    var isDefaultCase: Bool {
```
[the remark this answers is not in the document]
amarekano: If we add this check then importing switch blocks that don't have a default case will fail. I can update the generator to ensure that we always emit a default case. Would that work?

Sources/Fuzzilli/Core/CodeGenerators.swift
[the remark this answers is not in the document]
amarekano: Yeah agreed, that's a much better approach

Sources/Fuzzilli/FuzzIL/Operations.swift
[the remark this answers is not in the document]
saelo: No I think we should support switch statements without a default case (both in `Code.check` and in the generator). I'd just like to check that we don't have more than one default case :)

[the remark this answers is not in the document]
amarekano: Ah i see what you mean now, this makes sense to me. I'll add the check in there.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -527,13 +527,24 @@ public let CodeGenerators: [CodeGenerator] = [
...
-            cases.addDefault {
-                b.generateRecursive()
-            }
-            for _ in 0..<Int.random(in: 0...5) {
-                cases.add(b.randVar(), fallsThrough: probability(0.1)) {
-                    b.generateRecursive()
+        var hasDefault: Bool = false
+        
+        b.doSwitch(on: b.randVar()) { cases in
+            var caseVariables: [Variable] = []
```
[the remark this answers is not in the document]
amarekano: Yeah I think that might be a good idea to leave the duplicates in. If it gets too noisy (i.e. high proportion of cases that use the same variable) then we can revisit and tweak this.

amarekano:
review comments have been addressed with the latest commit. please review and let me know if I've missed something out.

saelo (commented):
Cool! Mostly just a couple more nits

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -316,9 +316,11 @@ public class ProgramBuilder {
 
         // Prefer inner scopes
         withProbability(0.75) {
-            candidates = chooseBiased(from: scopes, factor: 1.25)
-            if let f = filter {
-                candidates = candidates.filter(f)
+            if scopes.count > 0 {
```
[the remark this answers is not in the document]
saelo: Right but I think that's ok. The caller should never use `randVarInternal(excludeInnermostScope: true)` in that case (in the outermost scope), so better to crash early then return weird results I'd say

Sources/Fuzzilli/FuzzIL/Code.swift
```diff
@@ -201,6 +202,18 @@ public struct Code: Collection {
                 }
             }
 
+            // Ensure that we have at most one default case in a switch block
+            if instr.op is BeginSwitchCase && instr.inputs.count == 0{
+                let stackTop = defaultSwitchCaseStack.popLast()
+
+                // Check if the current block already has a default case
+                guard stackTop != nil && stackTop == false else {
+                    throw FuzzilliError.codeVerificationError("more than one default swtich case defined")
```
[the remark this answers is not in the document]
amarekano: so much for that NY resolution :D

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: makes sense, also now that the generator no longer relies on `randVarInternal` directly this is no longer needed.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -527,18 +527,34 @@ public let CodeGenerators: [CodeGenerator] = [
...
+        // Generate a minimum of two variables for the switch block
+        for _ in 0..<Int.random(in: 2...7) {
+            candidates.append(b.randVar())
+        }
+
+        guard candidates.count > 2 else { return }
+
+        b.doSwitch(on: candidates[0]) { cases in
+            for v in candidates.dropFirst() {
+                if !cases.hasDefault && probability(0.5) {
```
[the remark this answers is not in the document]
amarekano: So adding this change (i.e. the placeholder var number) causes the following assertion in `internalAppend` to fail:
```
assert(!instr.inouts.contains(where: { $0.number >= numVariables }))
```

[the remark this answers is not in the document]
amarekano: I think it might be easier to use a bitmap where the number of bits is one greater than the number of candidates. We then set one bit at random and that's the placeholder for the default case.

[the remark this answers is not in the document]
amarekano: I've added a simpler implementation to the latest commit. I think that gets us the random distribution that we are after.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -527,18 +527,36 @@ public let CodeGenerators: [CodeGenerator] = [
...
 
     CodeGenerator("SwitchCaseGenerator", input: .anything) { b, cond in
-        b.doSwitch(on: cond) { cases in
-            cases.addDefault {
-                b.generateRecursive()
-            }
-            for _ in 0..<Int.random(in: 0...5) {
-                cases.add(b.randVar(), fallsThrough: probability(0.1)) {
-                    b.generateRecursive()
+        var candidates: [Variable] = []
```
saelo, line 530: Okay cool! Yeah I agree that this approach is better. I'd still like to be able to generate switch statements without a default case though, and there are a few other nits. How about we turn this function into

```swift
var candidates: [Variable] = []

// Generate a minimum of three cases (including a potential default case)
for _ in 0..<Int.random(in: 3...8) {
    candidates.append(b.randVar())
}

// If this is set, the selected candidate becomes the default case
var defaultCasePosition = -1
if probability(0.8) {
  defaultCasePosition = Int.random(in: 0..<candidates.count)
}

b.doSwitch(on: cond) { cases in
    for (idx, val) in candiates.enumerated() {
        if idx == defaultCasePosition {
            cases.addDefault(previousCaseFallsThrough: probability(0.1)) {
                b.generateRecursive()
            }
        } else {
            cases.add(val, previousCaseFallsThrough: probability(0.1)) {
                b.generateRecursive()
            }
        }
    }
}
```
I think that should then cover everything :D

amarekano: Nice one! I'll add these changes

saelo:
Wohoo! :)

Issue #223, opened 2021-07-05
Warning "Failed to find edge in map" when using MarkovCorpus


saelo:
Transferring the discussion from https://github.com/googleprojectzero/fuzzilli/pull/171#issuecomment-864851774 to a new issue.

saelo:
So @WilliamParks from the ideas you listed in the original pull request, currently I think the idea that crashes set edges that are then "leaked" sounds most likely to me. There is [a mechanism](https://github.com/googleprojectzero/fuzzilli/blob/22d9a3708450027dd8dc4982263d0c84991f4cd7/Sources/Fuzzilli/Fuzzer.swift#L643) that should prevent a new fuzzing iterations before the previous iteration finished (and that includes minimization etc.), so I don't think that there is some kind of race condition there. On the other hand, a crashing sample will, as you said, cause `ProgramCoverageEvaluator.evaluateCrash` -> `cov_evaluate_crash` -> `internal_evaluate` to be called, and that will update the edge counts. I'll quickly hack something together that stops it from updating the edge counts when evaluating crashes and see if that fixes the issue.

saelo:
Update: I'm still getting a few `Failed to find edge in map` warnings (seems to be less though) even with that change and even on instances that have zero crashes, so probably that wasn't the only reason.

WilliamParks:
Hmm, interesting.

I haven't worked out the exact path that would cause the problem, but could there be an issue with `ThreadSync` and `NetworkSync` causing executions without entering the fuzzGroup DispatchGroup?

For example, `handleMessageInternal` in `NetworkSync` then calls `fuzzer.importProgram`, `importProgram` and finally `processInteresting' without ever entering `fuzzGroup`

WilliamParks:
Thinking about this more, I'm pretty sure `[evaluateAndIntersect](https://github.com/googleprojectzero/fuzzilli/blob/bc057d1eb2d47878b700022e59409e974bd01597/Sources/Fuzzilli/Evaluation/ProgramCoverageEvaluator.swift#L213)` leaks edges if said edge is over the resetLimit, as that edge will not be returned, but also will not be reset.

WilliamParks:
I've got a draft fix for the resetLimit issue, which doesn't seem to fully fix the problem. I'm in the process of adding instrumentation to track the source of each edge, so we can see where the unfound ones are coming from.

WilliamParks:
[...]

PR #237, opened 2021-07-29
Make deterministicCorpus the default

saelo:
Instead of --deterministicCorpus, there is now a --noDeterministicCorpus flag
to achieve the opposite.

Due to deterministicCorpus, we can now also run the target JS engines with
background threads enabled (a source of non-determinism), which more closely
resembles their usual mode of execution.


saelo:
@WilliamParks WDYT?

WilliamParks:
Looks good to me!

I found some issues with improper handling of nondeterministic edges while investigating #223, which should help as well

saelo:
Thanks!

PR #239, opened 2021-07-31
Update CI setup

amarekano:
Updates branch name to fix the pipeline


saelo:
Aha, so that's why that stopped working, thanks! :D

PR #240, opened 2021-07-31
Delete operations have one boolean output

amarekano:
[Several](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/delete-property-check-structure-transition.js) [jsc](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/multi-del-by-offset-doesnt-always-def.js) [testcases](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/array-toString-non-callable-join.js) use the output from delete operations and this PR enables this capability in FuzzIL


saelo:
Thanks!

Issue #241, opened 2021-07-31
Adding a FuzzIL op for strict directive


amarekano:
I've come across [a](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/tail-call-in-inline-cache.js) [few](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/arguments-exit-strict-mode.js) [testcases](https://github.com/WebKit/WebKit/blob/main/JSTests/stress/arith-tan-on-various-types.js) where the `use strict` directive is applied to the entire js program rather than a function definition. I think it might be worth adding a FuzzIL op to generate this directive. This would also allow us to refactor the various Begin*Function and End*Function operations to use the strict directive operation rather than having to maintain multiple ops for strict and non-strict functions. The op would be similar to `Break`, `Continue` or `Nop` ops something as follows:

```
class StrictDirective: Operation {
    init(numOutputs: Int = 0) {
        super.init(numInputs: 0, numOutputs: numOutputs)
    }
}
```

I don't believe this new op would require changes to the reducers or the mutators. Let me know what you think and I'll put together a PR for these changes.

amarekano:
One potential issue with PR #242 is with the reduction and mutation of strict arrow functions. The `use strict` instruction needs to be generated before the function declaration and as a result this won't be captured as part of the function block. So this may require handling by the reducer and splice mutators. I'm not sure if it's worth adding special handling for the strict arrow functions since the outer scope, within which the arrow function is defined, would run in strict mode anyway and the existing reducers/mutators would handle the outer scope like any other block reduction/mutation. WDYT?

PR #242, opened 2021-08-02
Implements a strict-mode toggle for Function operations

amarekano:
This is a WIP PR for #241


saelo:
Cool, thanks! So while I like the simplicity of this approach, I'm a bit worried about two things (specific to the MutationEngine):
1. The chance of a "use strict" being placed at the right point by a mutation is very low I think (it may even be impossible to place it at the start of a script due to the [prefix](https://github.com/googleprojectzero/fuzzilli/blob/9b7d61aeb6a1df06afba2835935a47956cd1e1d1/Sources/Fuzzilli/Core/MutationEngine.swift#L25)), and so
2. A lot of "use strict" will be removed by the minimizer, and then probably never come back

But I also can't really think of a better way. Maybe we should keep the `StrictFunctionDefinition`, but then somehow allow the entire script to become strict mode through some other way, for example by adding an attribute or comment to the `Program` class? I don't really think we'd need to create *new* strict mode programs (i.e. don't need a mutation that turns an existing program into a strict-mode program), since we should cover the same functionality just through strict-mode functions. Alternatively, and maybe even simpler, when compiling a strict-mode JS program, maybe we could just wrap the entire code into a strict-mode function? WDYT?

amarekano:
> Alternatively, and maybe even simpler, when compiling a strict-mode JS program, maybe we could just wrap the entire code into a strict-mode function? WDYT?

Yeah that's a good shout! Saves us the bother of having to track program attributes.

> Maybe we should keep the StrictFunctionDefinition

Can we instead use a boolean flag to track whether a function should be lifted to strict mode? The reason being we would then have to define _strict_ variants for all function definition ops. Iirc we had a boolean flag to track this in an earlier implementation of `BeginFunctionDefinition` before spliting them into `BeginPlain*` and  `BeginStrict*` version.

amarekano:
As we discussed, the strict directive op has now been replaced with a strict-mode flag for each Function operation.

amarekano:
@saelo any thoughts on this one?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -482,7 +482,7 @@ public class ProgramBuilder {
         }
         if type.Is(.function()) {
             let signature = type.signature ?? FunctionSignature(withParameterCount: Int.random(in: 2...5), hasRestParam: probability(0.1)) 
-            return definePlainFunction(withSignature: signature) { _ in
+            return definePlainFunction(withSignature: signature, isStrict: probability(0.2)) { _ in
```
saelo, line 485: Is there a reason why the probability here is 0.2 but above it's 0.1? Maybe just use 0.1 everywhere for consistency?

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -358,14 +358,16 @@ class In: Operation {
...
-    
+    let isStrict: Bool
+
     /// Whether the last parameter is a rest parameter.
     var hasRestParam: Bool {
         return signature.hasVarargsParameter()
     }
     
-    init(signature: FunctionSignature) {
+    init(signature: FunctionSignature, isStrict: Bool = false) {
```
saelo, line 368: Nit: is the default parameter value necessary? Just because the other Operations don't have default parameters either.

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -283,12 +286,11 @@ public class JavaScriptLifter: Lifter {
             case let op as BeginPlainFunctionDefinition:
                 liftFunctionDefinitionBegin(op, "function")
 
-            case let op as BeginStrictFunctionDefinition:
-                liftFunctionDefinitionBegin(op, "function")
-                w.emit("'use strict';")
-
             case let op as BeginArrowFunctionDefinition:
                 let params = liftFunctionDefinitionParameters(op)
+                if op.isStrict {
```
saelo, line 291: Isn't this before the ArrowFunction now? Shouldn't it be after line 295?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -300,6 +302,9 @@ public class JavaScriptLifter: Lifter {
 
             case let op as BeginAsyncArrowFunctionDefinition:
                 let params = liftFunctionDefinitionParameters(op)
+                if op.isStrict {
```
saelo, line 305: Same as above

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: The few examples that I've come across had the strict directive before the arrow function definition. Let me look this up again

amarekano: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#arrow_functions_used_as_methods

It appears that strict and non-strict variants of arrow functions are semantically the same. I suppose we could emit the directive in the block rather than before it.

amarekano:
It appears that the two most recent comments to `main` are causing build failures :(

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
saelo: So I think it has to be inside the block:
```javascript
> let nonstrict = () => { no_such_var = 42; }
undefined
> let strict = () => { 'use strict'; no_such_var2 = 42; }
undefined
> nonstrict()
undefined
> strict()
Uncaught ReferenceError: no_such_var2 is not defined
    at strict (REPL2:1:49)
> 
```
It's also how I'd interpret "Function code is strict mode code if the associated FunctionDeclaration, FunctionExpression, GeneratorDeclaration, GeneratorExpression, AsyncFunctionDeclaration, AsyncFunctionExpression, AsyncGeneratorDeclaration, AsyncGeneratorExpression, MethodDefinition, ArrowFunction, or AsyncArrowFunction is contained in strict mode code or if the code that produces the value of the function's [[ECMAScriptCode]] internal slot begins with a Directive Prologue that contains a Use Strict Directive." (from https://tc39.es/ecma262/#sec-strict-mode-code)

amarekano:
changes incorporated and PR rebased

saelo:
Thanks!

Issue #244, opened 2021-08-05
Lifting IR Operation createArray


oicu0619:
```
v0 <- LoadInteger '1'
v1 <- LoadUndefined
v2 <- CreateArray [v0, v1]
```
will be lifted to 
```
const v2 = [1,];
```
Which means [1].  The comma will be interpreted as trailing comma by JS engine, which will be ignored. Then if i want a array which end with hole, i have to put two ```LoadUndefined``` there.
Is this intends? Thanks!

saelo:
Ah yes, good catch and thanks for the PR, let's continue the discussion there!

PR #245, opened 2021-08-06
Lifting IR Operation createArray

oicu0619:
PR for #244 .

check for if the last element of array is HOLE, if so, add another trailing comma.


Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -188,7 +188,10 @@ public class JavaScriptLifter: Lifter {
 
             case is CreateArray:
                 // When creating arrays, treat undefined elements as holes. This also relies on literals always being inlined.
-                let elems = instr.inputs.map({ let text = expr(for: $0).text; return text == "undefined" ? "" : text }).joined(separator: ",")
+                var elems = instr.inputs.map({ let text = expr(for: $0).text; return text == "undefined" ? "" : text }).joined(separator: ",")
+                if instr.inputs.last != nil && expressions[instr.inputs.last!] != nil && expressions[instr.inputs.last!]!.text=="undefined" {
```
saelo, line 192: I think this can be simplified a bit, for example to something like
```swift
if elems.last == "," {
    // If the last element is supposed to be a hole, we need two commas
    elems += ","
}
```

saelo (commented):
Great, thanks!

Apart from the comment, could you please also change this test
https://github.com/googleprojectzero/fuzzilli/blob/73ff3047fdab8394582e753996eab215f45ba29c/Tests/FuzzilliTests/LifterTest.swift#L367
and add a hole to the end of some array there to test this new behaviour?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
oicu0619: i changed it to the simplified version of code and added the test~

saelo:
Thanks!

PR #247, opened 2021-08-09
add holey support for createarraywithspread

oicu0619:
add holey support for createarraywithspread, as in #197  #245


saelo:
Thanks!

Issue #249, opened 2021-08-10
about LoadRegExp IR inlined while lifting


oicu0619:
```
let a=/$/;
Object.is(a, a);
```
will be compiled to 
```
v0 <- LoadRegExp '$' ''
v1 <- LoadBuiltin 'globalThis'
v2 <- LoadProperty v1, 'Object'
v3 <- CallMethod v2, 'is', [v0, v0]
```
Then will be lifted to
```
const v2 = globalThis.Object;
const v3 = v2.is(/$/,/$/);
```

The original JS is true, but the compiled JS is false. Because Regexp is object, and two same regexp is not the same object.
Maybe LoadRegExp IR should not be inlined while lifting?  What do you think
Thanks!

PR #250, opened 2021-08-10
RegExp should not be inlined while lifting

oicu0619:
related to #249

### DIFF
--- a/Sources/Fuzzilli/Lifting/JSExpressions.swift
+++ b/Sources/Fuzzilli/Lifting/JSExpressions.swift
@@ -15,6 +15,15 @@
 // JavaScript expressions. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence
 public let Identifier           = ExpressionType(precedence: 20,                        inline: .always)
 public let Literal              = ExpressionType(precedence: 20,                        inline: .always)
+// RegExp are objects, and so for example for the FuzzIL program 
+//     v1 <- CreateRegExp
+//     Compare v1, v1
+// there is a difference between 
+//    let a = /a/;
+//    a === a;  // (true)
+// and 
+//     /a/ === /a/;  // (false)
+// the former being the correct JavaScript equivalent.
 public let RegExpLiteral        = ExpressionType(precedence: 20,                        inline: .singleUseOnly)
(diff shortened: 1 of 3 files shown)


Sources/Fuzzilli/Lifting/JSExpressions.swift
```diff
@@ -15,6 +15,7 @@
 // JavaScript expressions. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence
 public let Identifier           = ExpressionType(precedence: 20,                        inline: .always)
 public let Literal              = ExpressionType(precedence: 20,                        inline: .always)
+public let RegExpLiteral        = ExpressionType(precedence: 20,                        inline: .singleUseOnly)
```
saelo, line 18: Maybe add a comment here or in the JavaScriptLifter.swift to explain why we handle them differently? I guess something like 
```swift 
// RegExp are objects, and so for example for the FuzzIL program 
//     v1 <- CreateRegExp
//     Compare v1, v1
// there is a difference between 
//    let a = /a/;
//    a === a;  // (true)
// and 
//     /a/ === /a/;  // (false)
// the former being the correct JavaScript equivalent.
```
or so?

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -736,6 +736,30 @@ class LifterTests: XCTestCase {
 
         XCTAssertEqual(lifted_program,expected_program)
     }
+
+    func testRegExpInline(){
+        let fuzzer = makeMockFuzzer()
+        let b = fuzzer.makeBuilder()
+
+        let v0 = b.loadRegExp("a", RegExpFlags(rawValue:0))
```
saelo, line 744: Nit: here and below, I think you can just use `b.loadRegExp("a", RegExpFlags())`

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -736,6 +736,30 @@ class LifterTests: XCTestCase {
...
+        b.compare(v0,v0, with: Comparator.equal);
```
saelo, line 745: Nit: I think you can drop `Comparator` here and below: `b.compare(v0,v0, with: .equal);`

oicu0619:
Fixed that~

saelo:
Thanks!

PR #252, opened 2021-08-18
Added missing EndClassDefinition to ContextAnalyzer

amarekano:


Sources/Fuzzilli/FuzzIL/Analyzer.swift
```diff
@@ -169,7 +169,8 @@ struct ContextAnalyzer: Analyzer {
     mutating func analyze(_ instr: Instruction) {
```
saelo, line 169: Given this bug, we should probably have tests for this. Unless you really feel like adding such a test, maybe just add a `// TODO add a test for this analyzer when modifying this function in the future` here?

saelo (commented):
Oh good catch, thanks!

Sources/Fuzzilli/FuzzIL/Analyzer.swift
amarekano: I've added a test for the Context Analyzer

amarekano:
Would we also need additional handling for BeginMethod and EndMethod ops?

Sources/Fuzzilli/FuzzIL/Analyzer.swift
saelo: thanks!

saelo:
I think we don't, currently. The handler for `BeginClassDefinition` adds the `.function` context, which will stay active throughout the function definition. This works because the content of a `ClassDefinition` only consists of method definitions.

amarekano:
Cool, I guess we're good to merge in that case!

Issue #235, opened 2021-07-28
Proposal for improvements to `generateInternal`


Zon8Research:
Proposal for improvements to `generateInternal`.

So currently code generation works like this (from what I can see):

1. Pick any random codegen
2. If we meet the required context and inputs, generate code.
3. Otherwise, do nothing and return.

I propose we take the following approach:

1. Find all codegens that meet our context/inputs.
2. Pick a random codegen from this newly generated list.

The reasons for doing this are as follows:
* It can be common for the random choice to pick something that is out of context or don't have inputs for, therefore returning without generating code. The new approach will mean we always generate a valid codegen.
* It would also improve the chances of hitting lower probability code generators if other out of scope codegens are removed.
* This helps reduce a larger issue, and that is that the test cases generated by fuzzilli have an over proportionate amount of empty functions, loops and if statements. eg

```
for (const v375 in v374) {
}
function v376(v377) {
}
function v378(v379,v380) {
    for (let v384 = 0; v384 < 100; v384 = v384 || -1024) {
    }
    const v385 = (v386,v387,v388,v389,v390) => {
    };
}
```

Example code for this change would be to replace https://github.com/googleprojectzero/fuzzilli/blob/10d5db4700b2d02481b916302269025cee9a1f69/Sources/Fuzzilli/Core/ProgramBuilder.swift#L832 with as follows:

```swift
// We can't run code generators if we don't have any visible variables.
var inScopeCodeGens: [CodeGenerator] = []
guard self.scopeAnalyzer.visibleVariables.count > 0 else { return }
for codegen in self.fuzzer.codeGenerators {
    if codegen.requiredContext.isSubset(of: self.context) {
        inScopeCodeGens.append(codegen)
    }
}
let generator = inScopeCodeGens.randomElement()
if generator == nil {
    return
}

self.run(generator!)
```

One argument against this might be potential performance impact of finding all codegens that are suitable.

I've timed the generateInternal function with the new changes and this is the speed in debug mode:

```
0.00012600421905517578 seconds elapsed
```

Compared to the original speed which was:

```
1.0013580322265625e-05 seconds elapsed
```

However, you could also argue that this performance impact is inconsequential compared to how long is spent executing the payload. Also, this isn't even factoring in the time wasted currently by attempting to generate a codegen, but then returning with nothing. So you may actually end up saving time by having less wasted cycles.

Looking forward to hearing your thoughts.

saelo:
[...]

PR #255, opened 2021-08-25
Codegen improvements

amarekano:


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -829,10 +829,19 @@ public class ProgramBuilder {
                 self.splice(from: program)
             }, {
                 // We can't run code generators if we don't have any visible variables.
-                guard self.scopeAnalyzer.visibleVariables.count > 0 else { return }
-                let generator = self.fuzzer.codeGenerators.randomElement()
-                if generator.requiredContext.isSubset(of: self.context) {
-                    self.run(generator)
+                if self.scopeAnalyzer.visibleVariables.count == 0 {
+                    // Generate some variables
+                    self.run(self.fuzzer.programPrefixGenerators.randomElement()!)
```
saelo, line 834: Nit: I think you can use `chooseUniform(from: self.fuzzer.programPrefixGenerators)` to avoid the `!`

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -829,10 +829,19 @@ public class ProgramBuilder {
...
+                }
+                
+                // Select a generator at random and run it
+                while true {
```
saelo, line 838: I'm a bit paranoid about `while true` since there've been various bugs in the past that have led to infinite loops. Can we do one of these instead:

- Turn this into a for loop that runs, say, 100 times. Turn the `break` in line 843 into a `return`. Log an error after the loop, stating that we couldn't find a suitable CodeGenerator to run in the active context
- Follow the approach suggested in https://github.com/googleprojectzero/fuzzilli/issues/235 and first generate a list of all suitable CodeGenerators, then pick a random one from that list

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -93,6 +93,21 @@ public class Fuzzer {
...
+            CodeGenerators.get("StringGenerator"),
+            CodeGenerators.get("BuiltinGenerator"),
+            CodeGenerators.get("RegExpGenerator"),
+            CodeGenerators.get("BigIntGenerator"),
+            CodeGenerators.get("FloatGenerator"),
+            CodeGenerators.get("FloatArrayGenerator"),
+            CodeGenerators.get("IntArrayGenerator"),
+            CodeGenerators.get("TypedArrayGenerator"),
+            CodeGenerators.get("ObjectArrayGenerator"),
+            CodeGenerators.get("ObjectArrayGenerator"),
```
saelo, line 108: Duplicate entry

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: I'm in favour of the first options (i.e. running a loop over fixed iterations). I did consider the proposal in #235 but then we'd be looping over all generators (multiplied by their weights) at each codegen call which is probably slower than just running fewer iterations to pick a random element.

Sources/Fuzzilli/Fuzzer.swift
amarekano: Oh Interesting, so the [existing code](https://github.com/googleprojectzero/fuzzilli/blob/4219480e371bd86a4eb5c2810e3ca37adc1a0a69/Sources/Fuzzilli/Core/FuzzEngine.swift#L73) had two entries (for a different generator) and I assumed this was intentional perhaps as to increase probability of objects being created?

saelo: Hmm right, yeah it was probably meant for that. I think with this PR, the `fuzzer.programPrefixGenerators` now has two somewhat distinct purposes: (1) a list of CodeGenerators from which to build program prefixes, and (2) a list of CodeGenerators that can be run when no other variables are currently visible. The two sets necessarily identical (e.g. for a prefix, we could also run a CodeGenerator that requires inputs).

Maybe we should just rename this to something like 'trivialCodeGenerators` with a comment that states that these are CodeGenerators that don't require inputs and generate simple objects/values that can subsequently be used. Then we just use that list to also generate the program prefix, and maybe do that by randomly picking 5 (which should be a constant `programPrefixSize` in FuzzEngine probably) CodeGenerators from the list. WDYT?

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Yep, sounds good!

Sources/Fuzzilli/Fuzzer.swift
amarekano: yeah that works for me, so essentially we still have one list then for both the ProgramBuilder and FuzzEngine to use. Do you also want to include generators that require inputs to the FuzzEngine in the way the exsiting code does it?
If we decide to go with randomly picking and running generators (rather than executing them sequentially like the existing code) when generating a prefix, then we should avoid adding generators that require inputs because running them without any available vars would cause `randVar` to crash at runtime.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -829,10 +829,19 @@ public class ProgramBuilder {
                 self.splice(from: program)
             }, {
                 // We can't run code generators if we don't have any visible variables.
-                guard self.scopeAnalyzer.visibleVariables.count > 0 else { return }
-                let generator = self.fuzzer.codeGenerators.randomElement()
-                if generator.requiredContext.isSubset(of: self.context) {
-                    self.run(generator)
+                if self.scopeAnalyzer.visibleVariables.count == 0 {
```
saelo, line 832: Nit: can this be `if self.scopeAnalyzer.visibleVariables.isEmpty {`?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -829,10 +829,19 @@ public class ProgramBuilder {
...
+                    // Generate some variables
+                    self.run(chooseUniform(from: self.fuzzer.trivialCodeGenerators))
```
saelo, line 834: Nit: maybe add an `assert(!self.scopeAnalyzer.visibleVariables.isEmpty)` after this line

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: On 2nd, I'm not so sure anymore. Let's say we have
```
BeginSwitch
// [ X ]
BeginSwitchCase
...
```
Here, if we try CodeGeneration at point `X`, we would try 50 times, then fail, since we don't have a `SwitchCaseGenerator` (even if we did, the probability of randomly finding it would be pretty low). The more restricted contexts we have, the bigger this issue becomes. So maybe the 2nd approach would actually be better here? We can then just add a comment that we could build a cache of available code generators indexed by context if performance ever becomes an issue.

Sources/Fuzzilli/Fuzzer.swift
saelo: I think it's good the way it is now, i.e. without CodeGenerators that take inputs, to keep it `trivial` :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Agreed, a cache would make sense here. For now I'll change this to the second approach.

saelo: [Don't know why my latest comment disappeared, but I guess you saw it :)]

amarekano: woah thats odd! I saw the comment posted in two places and that was strange but now they both seem to have been removed :D

amarekano:
done

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -829,11 +829,23 @@ public class ProgramBuilder {
...
+                // Enumerate generators that have the required context
+                var availableGenerators: [CodeGenerator] = []
+                for generator in self.fuzzer.codeGenerators {
+                    if generator.requiredContext.isSubset(of: self.context) {
+                        availableGenerators.append(generator)
+                    }
+                }
+
+                // Select a generator at random and run it
+                let generator = chooseUniform(from: availableGenerators)
```
saelo, line 847: I think you need to handle the case that `availableGenerators` isEmpty here (and in that case, just return)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -829,11 +829,23 @@ public class ProgramBuilder {
...
-                let generator = self.fuzzer.codeGenerators.randomElement()
-                if generator.requiredContext.isSubset(of: self.context) {
-                    self.run(generator)
+                if self.scopeAnalyzer.visibleVariables.isEmpty {
+                    // Generate some variables
+                    self.run(chooseUniform(from: self.fuzzer.trivialCodeGenerators))
+                    assert(!self.scopeAnalyzer.visibleVariables.isEmpty)
                 }
+                
+                // Enumerate generators that have the required context
```
saelo, line 838: Nit: and maybe add the comment here about possibly caching these results

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -93,6 +93,20 @@ public class Fuzzer {
     /// Fuzzer instances can be looked up from a dispatch queue through this key. See below.
     private static let dispatchQueueKey = DispatchSpecificKey<Fuzzer>()
 
+    /// List CodeGenerators that don't require inputs and generate simple objects/values that can subsequently be used.
```
saelo, line 96: Nit: "List of"?

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Fair, should we also log a warning or perhaps even a assert to ensure we have generators for each context?

saelo: So I think that this is going to be an expected situation if for example a Mutator tries to generate code right after a `BeginSwitch` but before the `BeginSwitchCase` (unless we have a CodeGenerator for just `BeginSwitchCase`). We should probably handle that case though, I guess by just aborting code generation entirely? Alternatively, I think you can also just put an assert/error in here for now, and we'll deal with this when we add restrictive contexts.

Also, while you are at it, I think you can now remove the `assert(performSplicingDuringCodeGeneration || hasVisibleVariables)` in line 825

amarekano: hmm, for now lets go with aborting code generation. We can add in assertions and testcases when we introduce restrictive contexts.

Issue #253, opened 2021-08-19
Enhance Object creation


amarekano:
Currently Object creation ops (i.e. `CreateObject` and `CreateObjectWithSpread`) only allows generating objects with data properties and spread elements. I would like to enhance these object creation that will allow generation of the following js object features:

* getters
* setters
* computed properties
* computed getters/setters

I've been thinking of a way to remodel these ops to support the features above and what I've come up with is a bit more radical and I'd like peoples thoughts on this. I'd like to introduce 9 new ops to support the existing and the new features of object creation. An example of which is listed below:

```
let obj <- BeginObjectCreation // 0 inputs and 1 output; isBeginBlock
	CreateProperty (name, variable) // generates simple data property; 1 inputs, 0 outputs
	
	CreateComputedProperty (variable, variable) // generates a computed data property; 2 inputs, 0 outputs	
	
	CreateSpreadProperty (name, variable) // spreads variable; 1 inputs, 0 outputs
	
	BeginGetter (name) // 0 inputs, 0 outputs
		...FuzzIL instructions
	
	BeginComputedGetter (variable) // 1 input, 0 outputs
		....FuzzIL instructions
	
	BeginSetter (name) // will have a default function signature with one param; 0 inputs, 0 outputs, 1 innerOutput
		....FuzzIL instructions
	
	BeginComputedSetter (name, variable) // will have a default function signature with one param; 1 input, 0 outputs, 1 innerOutput
		...FuzzIL instructions
EndObjectCreation
```

This model is very similar to the way `ClassDefinitions` have been modeled with `Begin*` ops being the start of a new block and `EndObjectCreation` closing the last opened block. However, `BeginObjectCreation` differs from `BeginClassDefinition` in two significant ways:

1. It does not store attributes such as property names or take any inputs. Generation of properties is done by adding additional operations such `CreateProperty`, `CreateComputedProperty` and `CreateSpreadProperty`.

2. The instructions that follow a `BeginObjectCreation` can only be of the type `CreateProperty`, `CreateComputedProperty` and `CreateSpreadProperty`. Unlike `BeginClassDefinition` where the instructions that follow the block start describe the class constructor.

# Design Considerations:
## Context and Scope changes
* We would need to add an additional context to the Context Analyzer to track object creation context.

## Semantic validation and code checks
* Similar to class semantic verification; we need to ensure that getters/setters are closed before starting new functions
* Similar block and scope management for `Begin*` operations
[...]

Issue #8, opened 2019-03-21
Improve loops in FuzzIL


saelo:
FuzzIL's representation of loops is oversimplified and cannot express the fact that more or less arbitrary computations can be performed in the loop header. This might, however, be interesting for things like JIT compiler fuzzing. As such it might make sense to split loops into constructs such as:

```
BeginLoopHeader
...
BeginLoopBody
...
EndLoop
```

However, the loop header has special requirements as only expressions are valid in there but not additional statements (so e.g. you cannot put more control flow operations into it). This would have to be expressed somehow (e.g. as a new [context](https://github.com/googleprojectzero/fuzzilli/blob/f6edda203f1c6bc3b38af49812f75b8f7102f9f7/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L191)).

Additionally, it might make sense to define something like a "loop variable", which could be something as simple as a Phi that is emitted by the LoopHead. This would simplify the current code generators for loops.

saelo:
Basically, JITing is just based on a simple counter: every unit of code (function or possibly loop body) has a counter counting how often it has been executed. Once the counter reaches a certain value, the code is JIT compiled. JSC and Spidermonkey work that way afaik. V8 does it slightly differently: there the JS execution stack is regularly inspected and the counter of every function on the stack is incremented (this approach probably favours long-running functions). In both cases, it's basically enough to just execute the function multiple times. This is easiest achieved through some form of loop, but the concrete form of that loop (a for loop, a while loop, etc.) shouldn't matter.

That said, it might still be interesting to have a `JITCall` instruction which behaves like a function call but executes the given function a specific number of time to force compilation. It's not clear if that would be superior to the [current approach](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/FuzzilliCli/Profiles/V8Profile.swift#L17) but could be something to experiment with.

oicu0619:
Do you have any further thought about the problem? it is not easy to compile JS with expressions in loop header to Fuzzil OP. Thanks!

saelo:
So once https://github.com/googleprojectzero/fuzzilli/pull/256 lands, we should be able to properly implement loops, in particular loop headers.

For that, we first need to split what is currently the `.script` context into two new contexts, e.g. `.statement` and `.expression`, and use the two everywhere where we currently use `.script`.

[...]

PR #256, opened 2021-08-26
Splicing instructions is now context aware

amarekano:
When splicing instructions into a program we want to ensure that the instructions being spliced fit the context of the program at the point of insertion. To implement this there are two fundamental changes that are required:
1. Operations now need to declare the context that is required for them to exist
2. Programs need to expose the contexts that are currently active at the point of insertion.

This also allows us to implement operations that can be scoped to a certain context (see #253) and be confident in the knowledge that the ops won't be spliced into a context that they aren't intended for.

### DIFF
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -686,7 +686,7 @@ public class ProgramBuilder {
     }
 
     /// Append a splice from another program.
-    public func splice(from program: Program, at index: Int) {
+    public func splice(from program: Program, at index: Int, activeContexts: [ProgramContext]) {
         trace("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
         beginAdoption(from: program)
 
@@ -759,7 +759,7 @@ public class ProgramBuilder {
         trace("Splicing done")
     }
 
-    func splice(from program: Program) {
+    func splice(from program: Program, activeContexts: [ProgramContext]) {
         // Pick a starting instruction from the selected program.
         // For that, prefer dataflow "sinks" whose outputs are not used for anything else,
         // as these are probably the most interesting instructions.
@@ -771,7 +771,7 @@ public class ProgramBuilder {
             // Some instructions are less suited to be the start of a splice. Skip them.
         } while counter < 25 && (program.code[idx].isJump || program.code[idx].isBlockEnd || !program.code[idx].hasInputs)
 
-        splice(from: program, at: idx)
+        splice(from: program, at: idx, activeContexts: activeContexts)
     }
 
     private var openFunctions = [Variable]()
@@ -817,7 +817,7 @@ public class ProgramBuilder {
             withEqualProbability({
                 guard self.performSplicingDuringCodeGeneration else { return }
                 let program = self.fuzzer.corpus.randomElementForSplicing()
-                self.splice(from: program)
+                self.splice(from: program, activeContexts: self.contextAnalyzer.activeContexts)
             }, {
                 // We can't run code generators if we don't have any visible variables.
                 if self.scopeAnalyzer.visibleVariables.isEmpty {
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -687,20 +687,8 @@ public class ProgramBuilder {
 
     /// Append a splice from another program.
     public func splice(from program: Program, at index: Int) {
-        print("\n ================ HOST PROG ============")
-        for instr in code {
-            print(instr)
-        }
-        print("================= END HOST PROG =================\n\n")
-
-        print("\n ============ TARGET PROG =================")
-        for instr in program.code {
-            print(instr)
-        }
-        print("================= END TARGET PROG ================\n")
-
         trace("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
-        print("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
+        
         beginAdoption(from: program)
         let source = program.code
 
@@ -714,7 +702,7 @@ public class ProgramBuilder {
         //   Otherwise copy the whole block including its content
         var requiredInputs = VariableSet()
         var remainingInputs = VariableSet()
-        var requiredContextStack = [Context.script]
+        var requiredContextStack = [Context.empty]
 
         // Helper function to add an instruction, or possibly multiple instruction in the case of blocks, to the slice.
         func add(_ instr: Instruction, includeBlockContent: Bool = false) {
@@ -725,7 +713,6 @@ public class ProgramBuilder {
                     if instr.hasOutputs {
                         for output in instr.outputs {
                             if remainingInputs.contains(output) {
-                                //print("removing \(output) from remainingInputs")
                                 remainingInputs.remove(output)
                             }
                         }
@@ -735,22 +722,23 @@ public class ProgramBuilder {
                 if !remainingInputs.isEmpty && !instr.innerOutputs.isEmpty {
                     for innerOutput in instr.innerOutputs {
                         if remainingInputs.contains(innerOutput){
-                                //print("removing \(innerOutput) from remainingInputs")
-                                remainingInputs.remove(innerOutput)
-                            }
+                            remainingInputs.remove(innerOutput)
                         }
                     }
                 }
-                //print("Appending \(neededInputs) to remainingInputs")
-                requiredContextStack.last.formUnion(instr.op.requiredContext)
+                
                 requiredInputs.formUnion(instr.inputs)
                 remainingInputs.formUnion(instr.inputs)
+                var currentContext = requiredContextStack.popLast()!
+                currentContext.formUnion(instr.op.requiredContext)
+                requiredContextStack.append(currentContext)
                 slice.insert(instr.index)
             }
 
             if instr.isBlock {
                 let group = BlockGroup(around: instr, in: source)
-                for instr in includeBlockContent ? group.includingContent() : group.excludingContent() {
+                let instructions = includeBlockContent ? group.includingContent() : group.excludingContent()
+                for instr in instructions.reversed() {
                     internalAdd(instr)
                 }
             } else {
@@ -763,13 +751,16 @@ public class ProgramBuilder {
 
         // First, add the selected instruction.
         add(source[idx], includeBlockContent: true)
-        print("requiredContexts: \(requiredContextStack)")
         // Then add all instructions that the slice has data dependencies on.
         while idx > 0 {
             
-            //print("requiredInputs.isEmpty: \(requiredInputs.isEmpty) Checking: requiredContexts: \(requiredContexts) and current context: \(self.context)")
-            if remainingInputs.isEmpty && requiredContextStack.last.isSubset(of: self.context) {
-                break
+            if remainingInputs.isEmpty && requiredContextStack.count == 1 {
+                let requiredContext = requiredContextStack.popLast()!
+                if requiredContext.isSubset(of: self.context) {
+                    requiredContextStack.append(requiredContext)
+                    break
+                }
+                requiredContextStack.append(requiredContext)
             }
 
             idx -= 1
@@ -789,27 +780,39 @@ public class ProgramBuilder {
                     add(instr)
                 }
             }
+
+            if instr.isBlockBegin {
+                var requiredContext = requiredContextStack.popLast()!
+                if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext != .empty && requiredContext != self.context {
+                    requiredContextStack.append(requiredContext)
+                    add(instr)
+                    requiredContext = requiredContextStack.popLast()!
+                } 
+                requiredContext = requiredContext.subtracting(instr.op.contextOpened)
+                if requiredContextStack.count < 1 {
+                    requiredContextStack.append(requiredContext)
+                }
+            }
+            if instr.isBlockEnd {
+                requiredContextStack.append([])
+            }
         }
         
         // If, after the loop, the current context does not contain the required context (e.g. because we are just after a BeginSwitch), abort the splicing
-        print("Checking: requiredContexts: \(requiredContexts) and current context: \(self.context)")
-        guard self.context.contains(requiredContexts) else {
-            print("bailing out of splice, requiredContexts: \(requiredContexts) and current context: \(self.context)")
+        let stillRequired = requiredContextStack.popLast()!
+        guard self.context.contains(stillRequired) else {
             endAdoption()
             return
         }
 
         // Finally, insert the slice into the current program.
-        print("\n============ ADOPTING into context: \(self.context) ===============")
         for instr in source {
             if slice.contains(instr.index) {
-                print(instr)
                 adopt(instr, keepTypes: true)
             }
         }
-        print("============END================\n")
         endAdoption()
-        trace("Splicing done")
+        trace("Splicing done")    
     }
 
     func splice(from program: Program) {
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -709,12 +709,10 @@ public class ProgramBuilder {
             guard !slice.contains(instr.index) else { return }
 
             func internalAdd(_ instr: Instruction) {
-                if !remainingInputs.isEmpty {
-                    if instr.hasOutputs {
-                        for output in instr.outputs {
-                            if remainingInputs.contains(output) {
-                                remainingInputs.remove(output)
-                            }
+                if !remainingInputs.isEmpty && instr.hasOutputs {
+                    for output in instr.outputs {
+                        if remainingInputs.contains(output) {
+                            remainingInputs.remove(output)
                         }
                     }
                 }
@@ -812,7 +810,7 @@ public class ProgramBuilder {
             }
         }
         endAdoption()
-        trace("Splicing done")    
+        trace("Splicing done")
     }
 
     func splice(from program: Program) {
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -778,7 +778,7 @@ public class ProgramBuilder {
 
             // When we encounter a block begin:
             // 1. We ensure that the context being opened removes at least one required context
-            // 2. The default context (.script) isn't the only context being opened
+            // 2. The default context (.script) isn't the only context being removed
             // 3. The required context is not empty
             // 4. The required context is not equal to self.context (we don't want to capture blocks that don't open atleast one requried context)
             if instr.isBlockBegin {
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -708,6 +708,13 @@ public class ProgramBuilder {
         // A stack of contexts that are required by the instruction in the slice
         var requiredContextStack = [Context.empty]
 
+        // Helper function to add a context to the context stack
+        func addContextRequired(requiredContext: Context) {
+            var currentContext = requiredContextStack.removeLast()
+            currentContext.formUnion(requiredContext)
+            requiredContextStack.append(currentContext)
+        }
+
         // Helper function to add an instruction, or possibly multiple instruction in the case of blocks, to the slice.
         func add(_ instr: Instruction, includeBlockContent: Bool = false) {
             guard !slice.contains(instr.index) else { return }
@@ -715,13 +722,9 @@ public class ProgramBuilder {
             func internalAdd(_ instr: Instruction) {
                 remainingInputs.subtract(instr.allOutputs)
 
-                remainingInputs.subtract(instr.innerOutputs)
-                
                 requiredInputs.formUnion(instr.inputs)
                 remainingInputs.formUnion(instr.inputs)
-                var currentContext = requiredContextStack.removeLast()
-                currentContext.formUnion(instr.op.requiredContext)
-                requiredContextStack.append(currentContext)
+                addContextRequired(requiredContext: instr.op.requiredContext)
                 slice.insert(instr.index)
             }
 
@@ -797,7 +800,7 @@ public class ProgramBuilder {
         }
         
         // If, after the loop, the current context does not contain the required context (e.g. because we are just after a BeginSwitch), abort the splicing
-        let stillRequired = requiredContextStack.popLast()!
+        let stillRequired = requiredContextStack.removeLast()
         guard stillRequired.isSubset(of: self.context) else {
             endAdoption()
             return
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -780,10 +780,9 @@ public class ProgramBuilder {
             // 1. We ensure that the context being opened removes at least one required context
             // 2. The default context (.script) isn't the only context being removed
             // 3. The required context is not empty
-            // 4. The required context is not equal to self.context (we don't want to capture blocks that don't open atleast one requried context)
             if instr.isBlockBegin {
                 var requiredContext = requiredContextStack.removeLast()
-                if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext.intersection(instr.op.contextOpened) != .script && requiredContext != .empty && requiredContext != self.context {
+                if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext.intersection(instr.op.contextOpened) != .script && requiredContext != .empty {
                     requiredContextStack.append(requiredContext)
                     add(instr)
                     requiredContext = requiredContextStack.removeLast()
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -753,12 +753,10 @@ public class ProgramBuilder {
         while idx > 0 {
             
             if remainingInputs.isEmpty && requiredContextStack.count == 1 {
-                let requiredContext = requiredContextStack.popLast()!
+                let requiredContext = requiredContextStack.last!
                 if requiredContext.isSubset(of: self.context) {
-                    requiredContextStack.append(requiredContext)
                     break
                 }
-                requiredContextStack.append(requiredContext)
             }
 
             idx -= 1
@@ -798,7 +796,7 @@ public class ProgramBuilder {
         
         // If, after the loop, the current context does not contain the required context (e.g. because we are just after a BeginSwitch), abort the splicing
         let stillRequired = requiredContextStack.popLast()!
-        guard self.context.contains(stillRequired) else {
+        guard stillRequired.isSubset(of: self.context) else {
             endAdoption()
             return
         }
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -49,6 +49,10 @@ public class ProgramBuilder {
         return contextAnalyzer.context
     }
 
+    public var activeContexts: [ProgramContext] {
+        return contextAnalyzer.activeContexts
+    }
+
     /// Counter to quickly determine the next free variable.
     private var numVariables = 0
 
@@ -759,7 +763,7 @@ public class ProgramBuilder {
         trace("Splicing done")
     }
 
-    func splice(from program: Program, activeContexts: [ProgramContext]) {
+    func splice(from program: Program) {
         // Pick a starting instruction from the selected program.
         // For that, prefer dataflow "sinks" whose outputs are not used for anything else,
         // as these are probably the most interesting instructions.
@@ -771,7 +775,7 @@ public class ProgramBuilder {
             // Some instructions are less suited to be the start of a splice. Skip them.
         } while counter < 25 && (program.code[idx].isJump || program.code[idx].isBlockEnd || !program.code[idx].hasInputs)
 
-        splice(from: program, at: idx, activeContexts: activeContexts)
+        splice(from: program, at: idx, activeContexts: self.activeContexts)
     }
 
     private var openFunctions = [Variable]()
@@ -817,7 +821,7 @@ public class ProgramBuilder {
             withEqualProbability({
                 guard self.performSplicingDuringCodeGeneration else { return }
                 let program = self.fuzzer.corpus.randomElementForSplicing()
-                self.splice(from: program, activeContexts: self.contextAnalyzer.activeContexts)
+                self.splice(from: program)
             }, {
                 // We can't run code generators if we don't have any visible variables.
                 if self.scopeAnalyzer.visibleVariables.isEmpty {
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -709,7 +709,7 @@ public class ProgramBuilder {
         var requiredContextStack = [Context.empty]
 
         // Helper function to handle context updates when handling block instructions
-        func handleBlockInstruction(instr: Instruction){
+        func handleBlockInstruction(instruction instr: Instruction, shouldAdd: Bool = false){
             // When we encounter a block begin:
             // 1. We ensure that the context being opened removes at least one required context
             // 2. The default context (.script) isn't the only context being removed
@@ -718,6 +718,9 @@ public class ProgramBuilder {
                 var requiredContext = requiredContextStack.removeLast()
                 if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext.intersection(instr.op.contextOpened) != .script && requiredContext != .empty {
                     requiredContextStack.append(requiredContext)
+                    if shouldAdd {
+                        add(instr)
+                    }
                     requiredContext = requiredContextStack.removeLast()
                 } 
                 requiredContext = requiredContext.subtracting(instr.op.contextOpened)
@@ -757,7 +760,7 @@ public class ProgramBuilder {
                 requiredInputs.formUnion(instr.inputs)
                 remainingInputs.formUnion(instr.inputs)
                 addContextRequired(requiredContext: instr.op.requiredContext)
-                handleBlockInstruction(instr: instr)
+                handleBlockInstruction(instruction: instr)
                 slice.insert(instr.index)
             }
 
@@ -809,35 +812,7 @@ public class ProgramBuilder {
                 }
             }
 
-            // When we encounter a block begin:
-            // 1. We ensure that the context being opened removes at least one required context
-            // 2. The default context (.script) isn't the only context being removed
-            // 3. The required context is not empty
-            if instr.isBlockBegin {
-                var requiredContext = requiredContextStack.removeLast()
-                if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext.intersection(instr.op.contextOpened) != .script && requiredContext != .empty {
-                    requiredContextStack.append(requiredContext)
-                    add(instr)
-                    requiredContext = requiredContextStack.removeLast()
-                } 
-                requiredContext = requiredContext.subtracting(instr.op.contextOpened)
-
-                // If the required context is not a subset of the current stack top, then we have contexts that should be propagated to the current stack top
-                // We must have at least one context on the stack
-                if requiredContextStack.count >= 1 {
-                    var currentTop = requiredContextStack.removeLast()
-                    requiredContext = requiredContext.subtracting(currentTop)
-                    if requiredContext != .empty {
-                        currentTop.formUnion(requiredContext)
-                    }
-                    requiredContextStack.append(currentTop)
-                } else {
-                    requiredContextStack.append(requiredContext)
-                }
-            }
-            if instr.isBlockEnd {
-                requiredContextStack.append([])
-            }
+            handleBlockInstruction(instruction: instr, shouldAdd: true)
         }
         
         // If, after the loop, the current context does not contain the required context (e.g. because we are just after a BeginSwitch), abort the splicing
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -740,6 +740,7 @@ public class ProgramBuilder {
             if instr.isBlock {
                 let group = BlockGroup(around: instr, in: source)
                 let instructions = includeBlockContent ? group.includingContent() : group.excludingContent()
+                // Instructions within blocks are evaluated in reverse order so that the evaluation is consistent with the caller loop
                 for instr in instructions.reversed() {
                     internalAdd(instr)
                 }
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -787,7 +787,7 @@ public class ProgramBuilder {
             // When we encounter a block begin:
             // 1. We ensure that the context being opened removes at least one required context
             // 2. The required context is not empty
-            // 3. The required context is not equal to self.context (we don't want to capture blocks don't open atleast one requried context)
+            // 3. The required context is not equal to self.context (we don't want to capture blocks that don't open atleast one requried context)
             if instr.isBlockBegin {
                 var requiredContext = requiredContextStack.removeLast()
                 if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext != .empty && requiredContext != self.context {
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -45,7 +45,7 @@ public class ProgramBuilder {
     /// Whether to perform splicing as part of the code generation.
     public var performSplicingDuringCodeGeneration = true
 
-    public var context: ProgramContext {
+    public var context: Context {
         return contextAnalyzer.context
     }
 
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -686,10 +686,22 @@ public class ProgramBuilder {
     }
 
     /// Append a splice from another program.
-    public func splice(from program: Program, at index: Int, activeContext: ProgramContext) {
-        trace("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
-        beginAdoption(from: program)
+    public func splice(from program: Program, at index: Int) {
+        print("\n ================ HOST PROG ============")
+        for instr in code {
+            print(instr)
+        }
+        print("================= END HOST PROG =================\n\n")
 
+        print("\n ============ TARGET PROG =================")
+        for instr in program.code {
+            print(instr)
+        }
+        print("================= END TARGET PROG ================\n")
+
+        trace("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
+        print("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
+        beginAdoption(from: program)
         let source = program.code
 
         // The slice of the given program that will be inserted into the current program.
@@ -701,13 +713,38 @@ public class ProgramBuilder {
         //   If we need an inner output of a block instruction then only copy the block instructions, not the content
         //   Otherwise copy the whole block including its content
         var requiredInputs = VariableSet()
+        var remainingInputs = VariableSet()
+        var requiredContextStack = [Context.script]
 
         // Helper function to add an instruction, or possibly multiple instruction in the case of blocks, to the slice.
         func add(_ instr: Instruction, includeBlockContent: Bool = false) {
             guard !slice.contains(instr.index) else { return }
 
             func internalAdd(_ instr: Instruction) {
+                if !remainingInputs.isEmpty {
+                    if instr.hasOutputs {
+                        for output in instr.outputs {
+                            if remainingInputs.contains(output) {
+                                //print("removing \(output) from remainingInputs")
+                                remainingInputs.remove(output)
+                            }
+                        }
+                    }
+                }
+
+                if !remainingInputs.isEmpty && !instr.innerOutputs.isEmpty {
+                    for innerOutput in instr.innerOutputs {
+                        if remainingInputs.contains(innerOutput){
+                                //print("removing \(innerOutput) from remainingInputs")
+                                remainingInputs.remove(innerOutput)
+                            }
+                        }
+                    }
+                }
+                //print("Appending \(neededInputs) to remainingInputs")
+                requiredContextStack.last.formUnion(instr.op.requiredContext)
                 requiredInputs.formUnion(instr.inputs)
+                remainingInputs.formUnion(instr.inputs)
                 slice.insert(instr.index)
             }
 
@@ -726,9 +763,15 @@ public class ProgramBuilder {
 
         // First, add the selected instruction.
         add(source[idx], includeBlockContent: true)
-
+        print("requiredContexts: \(requiredContextStack)")
         // Then add all instructions that the slice has data dependencies on.
         while idx > 0 {
+            
+            //print("requiredInputs.isEmpty: \(requiredInputs.isEmpty) Checking: requiredContexts: \(requiredContexts) and current context: \(self.context)")
+            if remainingInputs.isEmpty && requiredContextStack.last.isSubset(of: self.context) {
+                break
+            }
+
             idx -= 1
             let instr = source[idx]
 
@@ -747,14 +790,24 @@ public class ProgramBuilder {
                 }
             }
         }
+        
+        // If, after the loop, the current context does not contain the required context (e.g. because we are just after a BeginSwitch), abort the splicing
+        print("Checking: requiredContexts: \(requiredContexts) and current context: \(self.context)")
+        guard self.context.contains(requiredContexts) else {
+            print("bailing out of splice, requiredContexts: \(requiredContexts) and current context: \(self.context)")
+            endAdoption()
+            return
+        }
 
         // Finally, insert the slice into the current program.
+        print("\n============ ADOPTING into context: \(self.context) ===============")
         for instr in source {
             if slice.contains(instr.index) {
+                print(instr)
                 adopt(instr, keepTypes: true)
             }
         }
-
+        print("============END================\n")
         endAdoption()
         trace("Splicing done")
     }
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -688,8 +688,8 @@ public class ProgramBuilder {
     /// Append a splice from another program.
     public func splice(from program: Program, at index: Int) {
         trace("Splicing instruction \(index) (\(program.code[index].op.name)) from \(program.id)")
-        
         beginAdoption(from: program)
+
         let source = program.code
 
         // The slice of the given program that will be inserted into the current program.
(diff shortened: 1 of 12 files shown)


saelo:
Nice! So I'm wondering exactly how the context stuff should work for https://github.com/googleprojectzero/fuzzilli/issues/253

I think there are broadly two models that should work, discussed below. In any case, the CodeGenerators also need to state their context requirements. Maybe those could instead just state which Operations they generate, then those lists could be computed from that?

## 1. Permissive and restrictive contexts

In this model, two separate things can be expressed:
- Certain operations must only be present if a certain context is "active", i.e. in the top entry of the context stack. For example, `Break` can only occur inside loops, and the suggested `CreateComputedProperty` can only occur inside an object definition. I think it makes sense to attach these requirements to the Operation itself, as you are doing here with the `requiredContext` attribute
- Certain operations must not occur if a certain context is active. For example, with https://github.com/googleprojectzero/fuzzilli/issues/253 this is the case for most operations inside `BeginObjectCreation`. For https://github.com/googleprojectzero/fuzzilli/issues/8 most operations except for certain expressions would be disallowed inside loop headers. This could probably either be attached to the context, or it could be a separate attribute of the Operation, e.g. `permittedContext`.

In this model:
- `CreateComputedProperty` would have `permittedContext == requiredContext == .objectDefinition`
- `Break` would have `permittedContext == .empty` and `requiredContext == .loop`
- (with proper loop header support) `SomeExpression` would have 'permittedContext == .loopHeader`, `requiredContext == .empty`

## 2. Only permissive contexts, with default context

In this model, every Operation would just have a `requiredContext`, but we have a generic `defaultContext` or `statementContext` or so. The `requiredContext` must then be present in the active context for the operation to be allowed. Then:
- Most operations (at least those that are lifted to statements) would have `requiredContext == .defaultContext`
- `CreateComputedProperty` would have `requiredContext == .objectDefinition`
- `Break` would have `requiredContext == .loop`
- (with proper loop header support) `SomeExpression` would have `requiredContext == [.defaultContext, .loopHeader]` which means that it requires `.defaultContext` *or* `.loopHeader`

This model is probably slightly less flexible, but I don't think that matters. It maybe also puts a bit more burden on the `ContextAnalyzer`, as it e.g. has to remove all other contexts inside `ObjectDefinitions`. But I think all the logic for this is already there and should work correctly (we should probably add more tests though). Apart from that, this model is probably simpler, and also easier to statically verify in `Code.check`, so that's a big plus.

WDYT?

amarekano:
Thanks for the feedback on this and I agree with the points you've made! I'm currently leaning towards the second approach for the reasons you've already listed and this PR implements most of what's been discussed :)
The bit that is missing is the checks that would need to be added to `Code.check`. I'll add another commit for this.
 
What did you mean by removing contexts inside `ObjectDefinitions`? I believe managing the context stack for object creation should be similar to how we handle class creation, unless there's something I'm missing here.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -49,6 +49,10 @@ public class ProgramBuilder {
         return contextAnalyzer.context
     }
 
+    public var activeContexts: [ProgramContext] {
```
saelo, line 52: Is it enough to just expose the "active" context, i.e. the topmost one on the stack? The "inactive" contexts (below the topmost one) shouldn't really be required from what I can tell.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -20,6 +20,9 @@ public class Operation {
     /// The attributes of this operation.
     let attributes: Attributes
     
+    /// The context in which the operation can exist
+    let contextRequired: ProgramContext
```
saelo, line 24: Nit: maybe rename to `requiredContext`?

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -409,27 +413,27 @@ class EndAsyncGeneratorFunctionDefinition: EndAnyFunctionDefinition {}
...
-        super.init(numInputs: 1, numOutputs: 0, attributes: [.isJump])
+        super.init(numInputs: 1, numOutputs: 0, attributes: [.isJump], contextRequired: ProgramContext([.function, .generatorFunction, .asyncFunction]))
     }
 }
 
 // A yield expression in JavaScript
 class Yield: Operation {
     init() {
-        super.init(numInputs: 1, numOutputs: 1, attributes: [])
+        super.init(numInputs: 1, numOutputs: 1, attributes: [], contextRequired: ProgramContext.generatorFunction)
```
saelo, line 423: I think here and elsewhere in this file you can just write `requiredContext: .generatorFunction` instead of `requiredContext: ProgramContext.generatorFunction`

saelo (commented):
Cool, yeah so let's try that!

Yes, I think `ObjectDefinition` should basically just behave like the `ClassDefinition`, like you said, i.e. not propagate the current context as is e.g. [the case for loops](https://github.com/googleprojectzero/fuzzilli/blob/4219480e371bd86a4eb5c2810e3ca37adc1a0a69/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L176) but instead start a fresh one.

So I think what we'll need for this to work:
- A default context: `public static let default          = ProgramContext(rawValue: 1 << 0)` which replaces the current `script` context
- `requiredContext` for operations, like you already have, with a default value of `.default`
- Additional checks in `Code.check` which just verify that every operations only occurs inside its stated `requiredContext`
- Some adjustments to Mutators like the [`CombineMutator`](https://github.com/googleprojectzero/fuzzilli/blob/main/Sources/Fuzzilli/Mutators/CombineMutator.swift) so that they only insert new code when allowed (I guess when inside a default context)
- (optional) Since the contexts are now more important, I think it'd be nice to move [the ProgramContext struct](https://github.com/googleprojectzero/fuzzilli/blob/4219480e371bd86a4eb5c2810e3ca37adc1a0a69/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L135) into a new File in the same directory (I think the Analyzer can stay where it is though). Also maybe it should just be called `Context` now?
- (optional) additional tests for the ContextAnalyzer, since that is now even more important.

I think the CodeGenerators should already work properly, except that [this line](https://github.com/googleprojectzero/fuzzilli/blob/4219480e371bd86a4eb5c2810e3ca37adc1a0a69/Sources/Fuzzilli/Core/CodeGenerator.swift#L58) probably needs to be changed to `context: ProgramContext = .default` (and maybe we can then also remove the `.any` context).

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: I've been running some tests and we would need to walk the stack and check all contexts. We would need this for nested blocks. Here's an example:
```
async function x(){
  for(;;) {
   //.. instructions in for loop
  }
 }
```
In the for loop above it should be valid to splice a `Await` instruction since it's contained within the async function so the splicer would need to figure out is there's an 'active' asyncFunction context.

I've got some changes to fix this that I'll push in a bit.

saelo: But in this example, wouldn't you have the following context stack:
```
[.function, .asyncFunction, .loop]   // TOP
[.function, .asyncFunction]
[.default]
```
Because [the ContextAnalyzer will copy the existing context into the loop context](https://github.com/googleprojectzero/fuzzilli/blob/4219480e371bd86a4eb5c2810e3ca37adc1a0a69/Sources/Fuzzilli/FuzzIL/Analyzer.swift#L176)?

saelo: I think with this approach to contexts, we must only check the topmost ("active") context, since otherwise, the following example would break:

```
...
BeginFunctionDefinition
    BeginObjectDefinition
        X
...
```

In this case, at `X` you'd have the following context stack:
```
[.objectDefinition]
[.default, .function]
[.default]
```
And so it'd now be invalid to insert operations that require the `.default` context, even though that context is on the stack.

amarekano: Ah yes you're right

amarekano:
> A default context: public static let default = ProgramContext(rawValue: 1 << 0) which replaces the current script context
Unfortunately, `default` is a keyword so we can't use that. Should we just continue with `script`?

> (optional) additional tests for the ContextAnalyzer, since that is now even more important.
I've added a couple of tests for class and function splicing. Are there any other scenarios you'd like me to add?

Sources/Fuzzilli/FuzzIL/Context.swift
```diff
@@ -0,0 +1,39 @@
...
+/// Current context in the program
+public struct Context: OptionSet {
+    public let rawValue: Int
+    
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
+    }
+    
+    // Default script context
+    public static let script            = Context([])
```
saelo, line 24: This now needs to have a value (i.e. `Context(rawValue: 1 << 0)`), because otherwise the subset tests won't work correctly (the `script` context would be a subset of every possible context since it's the empty set)

Sources/Fuzzilli/Minimization/Verifier.swift
```diff
@@ -47,7 +47,7 @@ class ReductionVerifier {
         }
 
         // At this point, the code must be statically valid though.
-        assert(code.isStaticallyValid())
+        guard code.isStaticallyValid() else { return false }
```
saelo, line 50: Why this change?

Tests/FuzzilliTests/AnalyzerTest.swift
```diff
@@ -30,8 +30,8 @@ class AnalyzerTests: XCTestCase {
             for instr in program.code {
                 contextAnalyzer.analyze(instr)
             }
-
-            XCTAssertEqual(contextAnalyzer.context, ProgramContext.script)
+            print(contextAnalyzer.context)
```
saelo, line 33: Nit :)

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -184,6 +184,119 @@ class ProgramBuilderTests: XCTestCase {
...
+                b.await(value: v3)
+                let v8 = b.loadInt(1337)
+                b.yield(value: v8)
+            }
+            b.doReturn(value: v4)
+        }
+
+        let original = b.finalize()
+
+        b.splice(from: original, at: original.code.lastInstruction.index - 5, activeContext: [.script])
```
saelo, line 282: Nit: would be nice to include a small comment what instruction something like `original.code.lastInstruction.index - 5` corresponds to

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -184,6 +184,119 @@ class ProgramBuilderTests: XCTestCase {
...
+        var actualSplice = b.finalize()
+
+        XCTAssertEqual(actualSplice, Program.init())
```
saelo, line 285: I think you can just write `Program()`

Tests/FuzzilliTests/AnalyzerTest.swift
```diff
@@ -30,8 +30,8 @@ class AnalyzerTests: XCTestCase {
...
+            XCTAssertEqual(contextAnalyzer.context, .script)
```
saelo, line 34: Can you extend this test case a bit or add a second one? Something like 
- Initial context is script context
- Start a function, ensure context is `.script, .function`
- Start a loop, ensure context is `.script, .function, .loop`
- Start another function inside the loop, ensure context is now `.script, .function` (.loop is gone)
- close the inner function, ensure context is again `.script, .function, .loop`
- maybe also start a class definition

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -184,6 +184,119 @@ class ProgramBuilderTests: XCTestCase {
         let float3 = b.reuseOrLoadFloat(4.2)
         XCTAssertNotEqual(floatOutOfScope!, float3)     // Variable went out of scope
     }
+
+    func testClassSplicing() {
```
saelo, line 188: Great, thanks for adding these tests!

Sources/Fuzzilli/FuzzIL/Context.swift
saelo: And yeah, I think we can just keep calling it `.script` then :)

Sources/Fuzzilli/Minimization/BlockReducer.swift
```diff
@@ -84,32 +84,34 @@ struct BlockReducer: Reducer {
         verifier.tryNopping(candidates, in: &code)
     }
 
-    private func reduceGenericBlockGroup(_ group: BlockGroup, in code: inout Code, with verifier: ReductionVerifier) {
-        var candidates = group.excludingContent().map({ $0.index })
-        if verifier.tryNopping(candidates, in: &code) {
-            // Success!
-            return
+    private func reduceGenericBlockGroup(_ group: BlockGroup, in code: inout Code, with verifier: ReductionVerifier, excludingContent excludeContent: Bool = true) {
```
saelo, line 87: What's the idea behind this change (the `excludingContent` parameter)?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -757,6 +756,14 @@ public class ProgramBuilder {
             }
         }
 
+        // Ensure that all required instructions are being adopted into a valid scope
+        for idx in requiredInstructions {
```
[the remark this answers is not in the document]
amarekano: That's true but you can also end up in a situation where the instruction being spliced has an input that is the result of an operation that has a non-default context. For example:
```
BeginAsyncFunction
...
v1 = Await v0
CallFunction(x, v1)
...
EndAsyncFunction
```
If we were to splice `CallFunction` the input re-wiring logic would also include the `Await` instruction as that is one of the inputs to `CallFunction`.

There were two approaches that I considered:
1. Drop the splice and return
2. Replace instructions that require a non-default context with LoadNull or LoadUndefined instructions.

I think the former seemed simpler and straightforward to implement. The latter whilst doable I wasn't fully convinced that just null-ing out inputs would be any better than just dropping the splice.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -665,9 +665,9 @@ public class ProgramBuilder {
     }
 
     /// Append a splice from another program.
-    public func splice(from program: Program, at index: Int) {
+    public func splice(from program: Program, at index: Int, activeContext: Context) {
```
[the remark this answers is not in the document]
amarekano: Initially I'd left this in there so that it's easier to specify arbitrary context when writing tests for this function. But now that I think about it we can get rid of it since tests for non-default contexts requires mocking up a host program and then calling splice within the right context. Similar to how it's being done for `testClassSplicing`. So yeah this param is redundant

Sources/Fuzzilli/Minimization/Verifier.swift
amarekano: Block reductions can lead to statically invalid code so rather that triggering asserts and failing silently on release builds I thought it makes sense to convert the assert to guard. For example consider the following async function that's being reduced:
```
BeginAsyncFunc
....
Await v0
...
EndAsyncFunc
```
When this block is reduced the reducer first tries to Nop out the Being and End instructions and then test the remaining instructions for static validity. This worked before we introduced context analysis to `Code.check` but now the check would fail due to `Await` being introduced to a non-default context.

So it made sense to me to have the assertion converted to a guard since if we don't have a statically valid program at that point of execution then we bail out of the reduction.

Sources/Fuzzilli/Minimization/BlockReducer.swift
amarekano: I had to introduce this for blocks that beign with `BeginClassDef`, `BeginWith` and `BeginAsync/Generator*` operations. For these specific blocks, Noping out the Begin and End operations more often leads to statically invalid Programs since they provision special contexts for some operations to exist in (eg. `BeginMethodDef` under a `BeginClassDef`). So it made sense to me to skip this initial reduction step and jump right to reduction that takes into consideration the block content.

amarekano:
Review comments addressed and testcases added.

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: Right, ok, correct. That's unfortunate. 

So I think the proper way to handle this would be to include surrounding Blocks in the slice if they are required to create the necessary context. The logic for that is probably not exactly trivial though :( It'd probably involve keeping track of context requirements in the loop that computes `requiredInstructions`, and including the next block that creates a necessary context if that context is not currently active.

I think for now we can also go with your approach and make a comment about how this should probably work in the future though.

Sources/Fuzzilli/Minimization/Verifier.swift
saelo: I see, thanks for explaining! I'm a bit worried that this will potentially hide bugs in reducers. Let's say a reducer generates statically invalid code due to a bug, we'd then never find out because it will just return here instead of failing.
On the other hand, it is a nice and simple solution. 
So how about just adding a comment here: 
```
TODO returning silently here might hide bugs in Reducers. However, certain Reducers are currently
allowed to generate statically invalid programs, for example the BlockReducer. To fix that, we should
change the return type of this function to an enum and encode the reason for failure. That way, reducers
that are expected to always generate statically valid programs can simply assert that that's the case.
```

When we do this, do we still need the changes in BlockReducer.swift?

amarekano: > When we do this, do we still need the changes in BlockReducer.swift?

The changes in BlockReducer were mainly to optimise this reduction logic to save us a few cpu cycles. I can revert those changes.

amarekano:
Another issue that I've just run into, that may be relevant to the discussion on #261, is the order in which spliced instructions are inserted into the host program.

For example consider that splice operation identifies the following `requiredInstructions` where the current context at the point of splicing in the host program is `[.script, .with]`:
```
StoreToScope name, p1
BeginAsyncFunctionDef p0, p1
EndAsyncFunctionDef
```
The `requiredInstructions` when evaluated independently are all valid under the current context. However due to dataflow dependencies, the instructions get inserted in the following order:
```
BeginAsyncFunctionDef p0, p1
StoreToScope name, p1
EndAsyncFunctionDef
```
This results in a statically invalid program since `BeginAsyncFunctionDef` creates a new context and the previous context (i.e. `[.script, .with]`) is no longer active. So what we'd need here is a mechanism for _context lookhead_ that can track predicted changes to context in the host program before the `requiredInstructions` are inserted into the host. One approach that I'm going to trail is cloning the existing contextAnalyzer and then analyze the requried instruction with this cloned analyzer. If an instruction exists in an invalid context then we bail on the splicing without modifying the host program or the original contextAnalyzer.

saelo:
Ah yes, good point! Just to make sure I understand correctly, in your example, the code from the target program is
```
...
BeginAsyncFunctionDef p0, p1
...
BeginWith
StoreToScope name, p1
EndWith
...
EndAsyncFunctionDef
...
```
And the splicing starts at the `StoreToScope` operation, correct?

I was also wondering if it'd be possible to add the information about which new context a block creates into the `BlockOperation`. That might make this kind of emulation and/or finding the necessary opening block instruction easier. So, for example if `BeginWith` included `openedContext = [.with]`, then we could figure out during splicing that this operation is required if we, for example, have the `StoreToScope` in the slice.
It wouldn't quite work like that though, we'd also need to encode if the existing context is closed. For example, `BeginForLoop` only adds `.loop` context but otherwise keeps the existing context. On the other hand, `BeginSwitch`, `BeginObjectCreation`, and the function definitions would open their respective context(s) but close all existing contexts. Maybe it's enough to encode this as a boolean flag `let propagatesContext: Bool` though. Doing this would also simplify the ContextAnalyzer a bit.

amarekano:
Sorry, I should have included the target program! yes you're right about the target program and the splicing does start at `StoreToScope`

Even if we encode the context opening/propagation, we would still need to perform context emulation when splicing to keep track of context changes, right? I agree on the point that the encoding would simplify the ContextAnalyzer.

saelo:
Would the following approach work:
- In [this](https://github.com/googleprojectzero/fuzzilli/blob/60b68c51af8a881d0504681d8bb2e29b6aba4265/Sources/Fuzzilli/Core/ProgramBuilder.swift#L765) loop, keep track of the current context requirement, simply by ORing the required contexts of each instruction kept (in a `requiredContexts` variable, similar to `requiredInputs`)
- When finding a BlockStart instruction, check if it's `openedContext` includes one or more of the currently `requiredContexts`, if so, `keep` the BlockStart and remove `openedContext` from `requiredContexts`.
- Stop the loop traversal as soon as there are no more missing inputs _and_ the currently active context contains the required context
- If, after the loop, the current context does not contain the required context (e.g. because we are just after a `BeginSwitch`), abort the splicing

I think this would be the ideal outcome, as it should be able to correctly handle most scenarios. It also maybe isn't too complex to build ...

saelo:
Nice, yeah good point and great testing! I think it should still work if we use a stack for tracking the `requiredContext`. At least so far, stacks have usually worked for tracking stuff across blocks. In your example if we start at `LoadArguments`:
- Initially: `requiredContext = [[.script]]`
- We `add` the `LoadArguments`, now `requiredContext = [[.script, .function]]`
- We find the `BeginFor`, which we ignore since it's not required. We remove `.loop` from the currently active context, but we don't pop an entry from the stack since it only has one element, meaning this must be a block surrounding our current slice. `requiredContext = [[.script, .function]]`
- We find the `EndPlainFunc` and push a new context. Now `requiredContext = [[.script, .function], []]`
- We find the `BeginPlainFunc`, and pop the latest context. (Really, what we need to do here is something `stillRequired = requiredContext.pop().subtracting(BeginPlainFunc.openedContexts)`, then add `stillRequired` to the now active `requiredContext`. See also next example). `requiredContext = [[.script, .function]]`
- We find the `BeginAsyncFunc`, remove `.function` from the current context, and so are now finished.

Now, if we instead start at the `BeginFor`:
- We find the block end, and so really start at the `EndFor`. So we push a new context. Now `requiredContext = [[.script], []]`
- We find the `LoadArguments`, now `requiredContext = [[.script], [.script, .function]]`
- We find `BeginFor`. So now we subtract `.loop` from the "active" context (which doesn't have `.loop` in it), then add the remainder to the next entry in the stack. So `requiredContext = [[.script, .function]]`
- Same as above, we find `EndPlainFunc`, push a new entry, then find `BeginPlainFunc`, and pop the entry again. Still `requiredContext = [[.script, .function]]`
- We find the `BeginAsyncFunc`, remove `.function` from the current context, and so are now finished.

I think this should work. But yeah, it's not exactly the most trivial algorithm... On the other hand, it should be quite generic, and able to handle all kinds of different scenarios correctly.

One remaining question is whether we always need to push `[.script]` when we find a closing block operation or if an empty context is enough. I think an empty context is enough, since I guess we'd add `.script` basically any time we add another operation?

amarekano:
Is there a reason we can't just extend the ContextAnalyzer like I've done [here](https://github.com/googleprojectzero/fuzzilli/pull/256/commits/bb135af2296767049780338157600f386a53d4f1). I feel like we're recreating functionality that the ContextAnalyzer already does for us.

saelo:
Well I guess the approach above would be the "ideal way" from this comment: `TODO: The ideal way to handle this would be to include surrounding Blocks in the slice if they are required to create the necessary context.`, which can figure out which surrounding Blocks need to be included, and so can't fail (at least as long as the current context includes `.script`). I don't think the existing `ContextAnalyzer` could handle it, since it does "forward" execution, and we need something more like "backward" execution for this, but the logic could be moved into a new `ContextRequirementAnalyzer`.

With just the "forward" emulation (the code you linked), we wouldn't be able to splice the `LoadArguments` in your example if we are not currently inside a function definition. I don't know how much of a problem that would be in practice. I guess it get's worse the more context dependent operations we have and the longer the slice is, but currently it's probably Ok. So if we go with that approach, I think we should have some monitoring in place that keeps track of how often a splice is aborted due to an incompatible context.

amarekano:
Agreed, lets stick with doing this right (i.e. capturing the blocks with the right context). I'll try and find some time this week to get this over the line.

amarekano:
Not the prettiest code but this first version works for majority of cases. I'm still running a few more stress tests to single out any edge cases that I've not captured yet. I'll get around to tidying up the code once the algorithm is stable. I thought I'd give you preview of how things are shaping up and If you have thoughts on making this concise.

amarekano:
> I think we should have some monitoring in place that keeps track of how often a splice is aborted due to an incompatible context.

This is a good idea and we should have this regardless. Shall I move the splicing operation to its own mutator class? Then we can leverage the generator stats to monitor splicing performance.

saelo:
> > I think we should have some monitoring in place that keeps track of how often a splice is aborted due to an incompatible context.
> 
> This is a good idea and we should have this regardless. Shall I move the splicing operation to its own mutator class? Then we can leverage the generator stats to monitor splicing performance.

So I've also been thinking about this. We definitely want `ProgramBuilder.generate` to use splicing as well for the HybridEngine. But for the MutationEngine, it might be nice to separate "pure" code generation from splicing to be able to weigh them differently and for stats. There are also use cases in the HybridEngine for pure code generation, such as the [MapTransitionsTemplate](https://github.com/googleprojectzero/fuzzilli/blob/8983b7079c0ad8c5997cf9ac33cd6b9454cc8a63/Sources/FuzzilliCli/Profiles/V8Profile.swift#L112). Maybe we should discuss this in a separate issue?

amarekano:
Yeah that sounds good to me! We can tackle this as a separate PR.

I've committed all my for this PR and it's ready to review. Please have a look and let me know what you think.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -702,18 +702,45 @@ public class ProgramBuilder {
...
+                currentContext.formUnion(instr.op.requiredContext)
+                requiredContextStack.append(currentContext)
                 slice.insert(instr.index)
             }
 
             if instr.isBlock {
                 let group = BlockGroup(around: instr, in: source)
-                for instr in includeBlockContent ? group.includingContent() : group.excludingContent() {
+                let instructions = includeBlockContent ? group.includingContent() : group.excludingContent()
+                for instr in instructions.reversed() {
```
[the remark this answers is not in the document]
saelo: I'm also wondering if something breaks since this logic bypasses the context stack handling (starting in lines 791), and could potentially add multiple blocks (e.g. entire switch-case, class definitions, etc.) without modifying the stack. But it's a bit hard to wrap my head around what would actually happen...
Maybe it'd be easier to move this logic into the loop (starting in line 757) to reason about what happens. But I'll think about this again, maybe there's an easier/more obviously correct way.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -746,6 +783,35 @@ public class ProgramBuilder {
                     add(instr)
                 }
             }
+
+            // When we encounter a block begin:
+            // 1. We ensure that the context being opened removes atleast one required context
+            // 2. The required context is not empty
+            // 3. The required context is not equal to self.context (we don't want to capture blocks don't open atleast one requried context)
```
saelo, line 790: Is this really true? Say the host program is inside a `BeginAsyncFunction` and the target program is
```
BeginFunction -> v7
BeginAsyncFunction
Await <- v7
```
So we'd now decide that we don't need the `BeginAsyncFunction`, then take the `BeginFunction`, and so end up with
```
BeginAsyncFunction       // From host
BeginFunction
Await
```
Which is wrong.

saelo (commented):
Need to think about the logic of `add` in the case of Blocks again, but here's the first round of comments :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -702,18 +702,45 @@ public class ProgramBuilder {
...
+
+        // A stack of contexts that are required by the instruction in the slice
+        var requiredContextStack = [Context.empty]
+
         // Helper function to add an instruction, or possibly multiple instruction in the case of blocks, to the slice.
         func add(_ instr: Instruction, includeBlockContent: Bool = false) {
             guard !slice.contains(instr.index) else { return }
 
             func internalAdd(_ instr: Instruction) {
+                if !remainingInputs.isEmpty && instr.hasOutputs {
```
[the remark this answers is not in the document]
amarekano: Unfortunately, `remainingInputs` is not of type `Set` so we can't use `subtract` on it. Would have been really handy though.

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: > Maybe it'd be easier to move this logic into the loop (starting in line 757) to reason about what happens. But I'll think about this again, maybe there's an easier/more obviously correct way.

So one of the challenges with doing that is then we would have to duplicate the handling for the first instruction that we add to the slice (the one that's outside the loop).

Also, if there are scenarios that you want me to try out then let me know and I'll whip up some testcases for it.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -746,6 +783,35 @@ public class ProgramBuilder {
...
+                var requiredContext = requiredContextStack.popLast()!
+                if requiredContext.subtracting(instr.op.contextOpened) != requiredContext && requiredContext != .empty && requiredContext != self.context {
+                    requiredContextStack.append(requiredContext)
+                    add(instr)
+                    requiredContext = requiredContextStack.popLast()!
+                } 
+                requiredContext = requiredContext.subtracting(instr.op.contextOpened)
+
+                // We must have atleast one context on the stack
+                if requiredContextStack.count < 1 {
```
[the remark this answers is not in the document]
amarekano: I have a testcase for this scenario: `testBeginForSplicing`.
So when we capture the `ForLoop` block and as a result `Await`, as we walk back up the instruction stack we encounter `BeginAsyncFunction` which opens a context that we need for `Await` so we capture the block begin and end, without the block contents.

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Just pushed a couple of test for the scenario above: `testSameContextSplicing` and `testSameContextSplicing2`.

the algorithm will capture both `BeginAsync` and `BeginFunc` blocks. So the final program would be:

```
BeginAsync            // From the host program
 BeginAsync  v1         // From the target
  BeginFunction
   Await v1
  EndFunction
 EndAsync
EndAsync
```
So the reason the `BeginAsync` from the target is captured is due to the innerOutput `v1` that's required. 
In the scenario that we don't have a data dependency, then the algorithm would only capture `Await` and splice that into the host program.

saelo: > In the scenario that we don't have a data dependency, then the algorithm would only capture Await and splice that into the host program.

Yeah that's the scenario I had in mind (and `Await` needs to have a data dependency on the `BeginFunction`). Don't you then end up with
```
BeginAsyncFunction       // From host
BeginFunction v1         // From target
Await v1
...
```
Which is invalid?

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: Ah right, my bad. So how about we add that method to `VariableSet`? PTAL: https://github.com/googleprojectzero/fuzzilli/pull/265

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Just so that I understand the scenario correctly, the host program would be:
```
BeginAsyncFunc
// point of splice
EndAsyncFunc
```
The target program would be:
```
BeginAsyncFunc
 BeginPlainFunc v1
  Await v1
 EndPlainFunc
EndAsyncFunc
```
and we start the splice at Await. Would that be right?
We wouldn't be able to generate the target program since `Await` cannot exist within a `PlainFunc` block. Starting a function block creates a new context and does not propagate the previous context.

saelo: Not quite. Host program
```
BeginAsyncFunc
// point of splice
EndAsyncFunc
```

Target program:
```
BeginPlainFunc v1
  BeginAsyncFunc
    Await v1
  EndAsyncFunc
EndPlainFunc
```
and then start at the `Await`.

Now if we don't take the `BeginAsyncFunc` because `The required context is not equal to self.context` (from the comment), then we'd end up with

```
BeginAsyncFunc
  BeginPlainFunc v1
    Await v1
  EndPlainFunc
EndAsyncFunc
```

If I understand correctly, which would be incorrect.

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -251,6 +251,392 @@ class ProgramBuilderTests: XCTestCase {
...
+
+        let expectedSplice = b.finalize()
+
+        XCTAssertEqual(actualSplice, expectedSplice)
+    }
+
+    func testBeginForSplicing() {
+        let fuzzer = makeMockFuzzer()
+        let b = fuzzer.makeBuilder()
+        b.mode = .conservative
```
saelo, line 416: Nit: I think we can remove `b.mode = .conservative` now from the tests since the splicing no longer rewires variables

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: Hmm yeah, I think the example needs to be a bit more complicated than what I wrote above. Does this work:
```
v1 <- LoadInt 42
v2 <- BeginFunctionDefinition
  BeginForLoop
    v3 <- LoadArguments
    Reassign v1, v3
  EndForLoop
EndFunctionDefinition
CallFunction v2, v1
```
Then put the builder into `.conservative` mode and start the splicing at `CallFunction`. Then it'll push a new context onto the `requiredContextStack` when finding the `EndFunctionDefinition` and again for `EndForLoop`, then decide it needs to capture the `Reassign`, then take the `LoadArguments` as well, at which point the top of the context stack is `[.function]`. When it then finds the `BeginForLoop`, from what I can tell, the current code would just drop the top of the stack and not take the `BeginForLoop` (which is correct), but then it'll also not take the `BeginFunctionDefinition` later on, which is incorrect. So instead of dropping the context, I think it needs to merge it with the one below.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -702,18 +702,38 @@ public class ProgramBuilder {
...
+
+        // A stack of contexts that are required by the instruction in the slice
+        var requiredContextStack = [Context.empty]
+
         // Helper function to add an instruction, or possibly multiple instruction in the case of blocks, to the slice.
         func add(_ instr: Instruction, includeBlockContent: Bool = false) {
             guard !slice.contains(instr.index) else { return }
 
             func internalAdd(_ instr: Instruction) {
+                if !remainingInputs.isEmpty {
```
saelo, line 716: is the check here and below required? Isn't just `remainingInputs.subtract(instr.allOutputs)` enough?

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Right, so this scenario is currently captured in testcase: `testSameContextSplicing2` which is under ProgramBuilderTests.
So the splice in your example above would capture the entire target program and splice that into the host. So we end up with: 
```
BeginAsync            // From the host program
 BeginPlain  v1         // From the target
  BeginAsyncFunction // From the target
   Await v1
  EndAsyncFunction
 EndPlain
EndAsync
```
I think the comment I've added is misleading and that's probably the source of confusion. An example that demonstrates why this check is necessary can be found in testcase: `testSplicing3`. Without the check we end up capturing the `BeginFunc` and `EndFunc` block which isn't required for the host program since neither does it have data dependencies nor does it open a required context that doesn't already exist in the host program. Also, now that I've looked at it again, `testSplicing3` is the only testcase that would fail without this check.

Tests/FuzzilliTests/ProgramBuilderTest.swift
amarekano: I've left this in the testcases so that we exercise the code path in the splice loop that checks if an instruction mutates any of the requiredInputs:
```
if mode == .conservative || (mode == .aggressive && probability(0.5)) {
                if instr.mayMutate(requiredInputs) {
```

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: So would the following testcase work currently:
```
v1 <- BeginFunctionDefinition
  BeginForLoop
    Break
  EndForLoop
EndFunctionDefinition
CallFunction v1
```
And start at `CallFunction`. IIUC correctly, this would happen:
- It takes CallFunction
- It pushes a new entry onto the scope stack for `EndFunctionDefinition` and again for `EndForLoop`
- It removes one entry for `BeginForLoop`
- It takes the entire block content at `BeginFunctionDefinition`, and so adds the `.loop` requirement into the current entry of the stack for `Break`, but doesn't remove that requirement when adding the `BeginForLoop`
- It removes one entry for `BeginFunctionDefinition` and now adds the `.loop` requirement to the remaining context (see https://github.com/googleprojectzero/fuzzilli/pull/256/files#r709268658)
- So at the end, the requiredContext is still `[.loop]`, which is incorrect and prevents the splicing from succeeding

Tests/FuzzilliTests/ProgramBuilderTest.swift
saelo: Ah cool, yeah that makes sense, maybe add a comment about that for the first test that does it, but not really important.

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Ah great, yeah that test does indeed cover the scenario!
I think the reason why what I'm describing doesn't happen is because when the splicing loop reaches the `BeginAsyncFunction` from the target, the `requiredContext = [.async]`, but the `self.context = [.script, .async]`, and so `requiredContext != self.context`. This is because 
```swift
class Await: Operation {
    init() {
        super.init(numInputs: 1, numOutputs: 1, attributes: [], requiredContext: .asyncFunction)
    }
}
```
so probably, here `requiredContext` should really be `[.asyncFunction, .script]`, since the other operations also have `.script` as their `requiredContext` (since it's the default value if none is provided)?
From what I can tell, `testSplicing3` doesn't include an instruction that requires `[.function]` context, and so doesn't need the `BeginFunctionDefinition`.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -713,13 +713,9 @@ public class ProgramBuilder {
...
             func internalAdd(_ instr: Instruction) {
-                if !remainingInputs.isEmpty {
-                    remainingInputs.subtract(instr.allOutputs)
-                }
+                remainingInputs.subtract(instr.allOutputs)
 
-                if !remainingInputs.isEmpty && !instr.innerOutputs.isEmpty {
-                    remainingInputs.subtract(instr.innerOutputs)
-                }
+                remainingInputs.subtract(instr.innerOutputs)
```
saelo, line 718: `allOutputs` should include `innerOutputs`, so I think you can just remove this line :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -702,18 +702,37 @@ public class ProgramBuilder {
...
         // Helper function to add an instruction, or possibly multiple instruction in the case of blocks, to the slice.
         func add(_ instr: Instruction, includeBlockContent: Bool = false) {
             guard !slice.contains(instr.index) else { return }
 
             func internalAdd(_ instr: Instruction) {
+                remainingInputs.subtract(instr.allOutputs)
+
                 requiredInputs.formUnion(instr.inputs)
+                remainingInputs.formUnion(instr.inputs)
+                addContextRequired(requiredContext: instr.op.requiredContext)
```
amarekano, line 727: I've added the helper function here. I think from a previous comment you wanted me to add this further below but I'm not sure where you meant that to be?

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: Ah no, the above example doesn't work, my bad, this is what I meant:
```
v1 <- LoadInt 42
BeginFunctionDefinition
  BeginForLoop
    v3 <- LoadArguments
    Reassign v1, v3
  EndForLoop
EndFunctionDefinition
CreateArray [v1]
```
Then start at `CreateArray`

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: This splice in this scenario would be:
```
BeginPlainFunctionDefinition, inouts_: [v0, v1, v2, v0]
LoadInteger, inouts_: [v3, v1]
LoadInteger, inouts_: [v4, v2]
LoadInteger, inouts_: [v5, v3]
BeginFor, inouts_: [v3, v4, v5, v6, v4]
Break, inouts_: [v5]
EndFor, inouts_: [v6]
EndPlainFunctionDefinition, inouts_: [v7]
CallFunction, inouts_: [v0, v7, v8]
```

So it does remove the `.loop` context and at the end of the splice we are left with `.script` as the required context.

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: Good catch! this does lead to an invalid program. I'll debug this further and push a fix.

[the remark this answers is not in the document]
saelo: Cool, glad it worked :D Yeah I think the fix is to basically merge the top two states when you reach an `BeginX` instead of just removing the top entry. Here, merging means subtracting the states opened by the `BeginX` if it is kept, then combining (unioning) the result with the other state. Unless there is only one entry in the stack, in which case you'd just keep the top state after the subtraction, like you currently do.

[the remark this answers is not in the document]
amarekano: This is what currently happens when we encounter `BeginFor`:
stack: [[.script], [.empty], [.script, .function]]
we pop the stack to retrieve (`requiredContext`): [.script, .function]
context opened by BeginFor: [.scrtipt, .loop]
subtracting `contextOpened` from `requiredContext` we get: [.function]
and since the result is not equal to `requiredContext` we end up capturing the `BeginFor` instruction. So now I think the result of the subtraction should be a subset of `contextOpened` in order to keep the instruction.

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: After making the latest changes (i.e. merging of states) this is the one testcase that currently fails. I don't believe it has anything to do with the merging logic and it's more to do with how we handle adding entire blocks in `add`.

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: This should now be resolved with the latest commit

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: Ok, yeah I think this testcase should fail once the merging logic works correctly, but with `add` not taking blocks into account.
So I agree that `add` is the problem, and we probably need to somehow force the block handling to happen in the outer loop. :/ I'll think about how to best achieve that

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: This problem should now be resolved with adding the intersection check: 
```
 if requiredContext.subtracting(instr.op.contextOpened) != requiredContext &&
    requiredContext.intersection(instr.op.contextOpened) != .script && 
    requiredContext != .empty {
                    requiredContextStack.append(requiredContext)
                    add(instr)
                    requiredContext = requiredContextStack.removeLast()
} 
```
We make sure that the default context isn't the only one that's being removed from `requiredContexts`.

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
amarekano: I've duplicated the block handling in a helper function that is now being called from `add`. It's a naive solution that works but there's room for improvement here.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -402,44 +410,60 @@ class BeginArrowFunctionDefinition: BeginAnyFunctionDefinition {}
...
+    init(signature: FunctionSignature, isStrict: Bool) {
+        super.init(signature: signature, isStrict: isStrict, contextOpened: [.script, .function, .asyncFunction, .generatorFunction])
+    }
+}
 class EndAsyncGeneratorFunctionDefinition: EndAnyFunctionDefinition {}
 
 class Return: Operation {
     init() {
-        super.init(numInputs: 1, numOutputs: 0, attributes: [.isJump])
+        super.init(numInputs: 1, numOutputs: 0, attributes: [.isJump], requiredContext: .function)
```
saelo, line 446: I think all these non-default `requiredContext` should also include `.script`, since that's the default value (line 47), and they do require `.script` (it's just that there is currently no case where there'd be `.function` context and not also `.script` context). WDYT?

Tests/FuzzilliTests/AnalyzerTest.swift
```diff
@@ -31,15 +31,127 @@ class AnalyzerTests: XCTestCase {
...
+
+        XCTAssertEqual(b.context, .script)
+
+        let _ = b.definePlainFunction(withSignature: FunctionSignature(withParameterCount: 3)) { args in
+            XCTAssertEqual(b.context, [.script, .function])
+            let loopVar1 = b.loadInt(0)
+            b.doWhileLoop(loopVar1, .lessThan, b.loadInt(42)) {
+                XCTAssertEqual(b.context, [.script, .function, .loop])
+                b.definePlainFunction(withSignature: FunctionSignature(withParameterCount: 2)) { args in
+                    XCTAssertEqual(b.context, [.script, .function])
```
saelo, line 58: Maybe open a 2nd loop in here, e.g. a `for` loop or `while` loop?

Tests/FuzzilliTests/AnalyzerTest.swift
```diff
@@ -31,15 +31,127 @@ class AnalyzerTests: XCTestCase {
...
+                b.defineAsyncFunction(withSignature: FunctionSignature(withParameterCount: 2)) { _ in
+                    XCTAssertEqual(b.context, [.script, .function, .asyncFunction])
+                }
+                XCTAssertEqual(b.context, [.script, .classDefinition, .function])
+            }
+        }
+        XCTAssertEqual(b.context, .script)
+
+        let _ = b.finalize()
+    }
```
saelo, line 145: Maybe also add a test for `CodeStrings`, since you (correctly!) changed them to not propagate the surrounding context.
Also maybe add one testcase, `testTrivialBlocks` or so, that tests various blocks that don't create new contexts, such as block statements, if-else, try-catch. (just open some context like `.function` + `.loop`, then create each of these blocks inside it and verify the context inside is equal to the one outside)

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -251,6 +251,445 @@ class ProgramBuilderTests: XCTestCase {
...
+        let original = b.finalize()
+
+        superclass = b.defineClass() { cls in
+            cls.defineConstructor(withParameters: [.integer]) { params in
+            }
+        }
+        b.defineClass(withSuperclass: superclass) { cls in
+            cls.defineConstructor(withParameters: [.string]) { _ in
+                // Splicing at CallSuperConstructor
+                b.splice(from: original, at: original.code.lastInstruction.index - 5)
```
saelo, line 301: Even with the comments, I still find it a bit tricky to verify that these splicing indices are correct... how about this:
Add a method
```swift
    /// Returns the index of the next instruction added to the program. This is equal to the current size of the program.
    public func indexOfNextInstruction() {
        return code.count
    }
```
To the ProgramBuilder, for example just below `dumpCurrentProgram`.
Then change these tests to be structured like this:
```swift
var splicePoint = -1
let b = fuzzer.makeBuilder()

// ...

        let v1 = b.createObject(with: ["foo": v0])
        splicePoint = b.indexOfNextInstruction()
        b.callSuperConstructor(withArgs: [v1])      // Splice here
     }

// ...

        b.splice(from: original, at: splicePoint)
```
That'll make it obvious where the splicing happens and will also guarantee that the test still works correctly if new instructions are added to the program.

saelo (commented):
Ok cool! So I left one more round of comments. I'd suggest that you then also run the fuzzer in a debug build for a couple million iterations, just to see if we missed any cases where splicing now produces an incorrect result. Afterwards, I'm happy to merge this, and then I might take a stab at simplifying the splicing logic a bit more. I have some ideas, but probably I just need to play with it a bit. Really nice that we have all these testcases now though! WDYT?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -769,7 +842,7 @@ public class ProgramBuilder {
             counter += 1
             idx = Int.random(in: 0..<program.size)
             // Some instructions are less suited to be the start of a splice. Skip them.
-        } while counter < 25 && (program.code[idx].isJump || program.code[idx].isBlockEnd || !program.code[idx].hasInputs)
+        } while counter < 25 && (program.code[idx].isJump || program.code[idx].isBlockEnd || !program.code[idx].hasInputs || !program.code[idx].op.requiredContext.isSubset(of: self.context))
```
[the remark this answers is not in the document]
amarekano: Makes sense, I don't think we need this condition any longer and I'll get rid of it.

Sources/Fuzzilli/FuzzIL/Analyzer.swift
```diff
@@ -131,65 +131,56 @@ struct ScopeAnalyzer: Analyzer {
...
-    var context: ProgramContext {
+    var context: Context {
         return contextStack.last!
     }
     
+    init(contextStack: [Context] = [Context.script]) {
+        self.contextStack = contextStack
+    }
+
     mutating func analyze(_ instr: Instruction) {
```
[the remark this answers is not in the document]
amarekano: What are your thoughts on using an operation attribute rather than a bool flag for `propagatesSurroundingContext`?

Sources/Fuzzilli/FuzzIL/Analyzer.swift
```diff
@@ -131,65 +131,56 @@ struct ScopeAnalyzer: Analyzer {
...
-            contextStack.append([.classDefinition, .function])
+            contextStack.append([.script, .classDefinition, .function])
+        } else if instr.op is BeginCodeString {
+            contextStack.append([.script])
+        } else if instr.op is BeginSwitch {
+            contextStack.append([context, .switchCase])
         }
     }
+
+    func copy() -> ContextAnalyzer {
```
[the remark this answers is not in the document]
amarekano: This is no longer required, I'll remove this.

Tests/FuzzilliTests/AnalyzerTest.swift
```diff
@@ -31,15 +31,127 @@ class AnalyzerTests: XCTestCase {
                 contextAnalyzer.analyze(instr)
             }
 
-            XCTAssertEqual(contextAnalyzer.context, ProgramContext.script)
+            XCTAssertEqual(contextAnalyzer.context, .script)
         }
     }
+
+    /*
```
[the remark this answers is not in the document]
amarekano: yep, I've cleaned that up

Tests/FuzzilliTests/AnalyzerTest.swift
amarekano: testcases added

Sources/Fuzzilli/FuzzIL/Operations.swift
amarekano: Yep, sounds good! I'll add this context.

amarekano:
I've addressed the majority of the review comments. The one item that I need your thoughts on is on the use of `Attributes` vs bool flags to track context propagation. The current implementation uses `Attributes` which I thought would make for a cleaner implementation but if you see an issue with this implementation then I can revert it to using bool flags instead.

> I'd suggest that you then also run the fuzzer in a debug build for a couple million iterations, just to see if we missed any cases where splicing now produces an incorrect result. 

I haven't run into any debug assertion failures after about 2mil iterations so it appears to be pretty stable.

saelo:
Wohoo! Great work!

PR #258, opened 2021-08-31
Enable InputMutator to pick variables from outer scopes when mutating end-of-block instructions

amarekano:


Sources/Fuzzilli/Mutators/InputMutator.swift
```diff
@@ -28,7 +28,7 @@ public class InputMutator: BaseInstructionMutator {
         // Replace one input
         let selectedInput = Int.random(in: 0..<instr.numInputs)
         b.trace("Mutating input \(selectedInput)")
-        inouts[selectedInput] = b.randVar()
+        inouts[selectedInput] = b.randVarFromOuterScope()
```
saelo, line 31: We don't generally want to select variables from the outer scope I think, only when the instruction is the end of a block (because then, the inner variables are still visible at this point, but shouldn't be selected as inputs)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -302,13 +302,20 @@ public class ProgramBuilder {
...
+    public func randVarFromOuterScope() -> Variable {
+        assert(hasVisibleVariables)
+        return randVarInternal(fromOuterScope: true)!
+    }
+
     /// Returns a random variable satisfying the given constraints or nil if none is found.
-    private func randVarInternal(_ selector: ((Variable) -> Bool)? = nil) -> Variable? {
+    private func randVarInternal(_ selector: ((Variable) -> Bool)? = nil, fromOuterScope: Bool = false) -> Variable? {
         var candidates = [Variable]()
+        let scopes = fromOuterScope && scopeAnalyzer.scopes.count > 1 ? scopeAnalyzer.scopes.dropLast() : scopeAnalyzer.scopes
```
saelo, line 314: I think you can remove `&& scopeAnalyzer.scopes.count > 1` because "If the number of elements to drop exceeds the number of elements in the collection, the result is an empty subsequence." (From https://developer.apple.com/documentation/swift/array/1689751-droplast). So if the API is used incorrectly, then this function will either just crash or return nil (which is fine), but won't accidentally return variables from the inner scope (which is probably not fine, because it might go unnoticed).

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -302,13 +302,20 @@ public class ProgramBuilder {
...
 
         // Prefer inner scopes
         withProbability(0.75) {
-            candidates = chooseBiased(from: scopeAnalyzer.scopes, factor: 1.25)
+            candidates = chooseBiased(from: scopes, factor: 1.25)
             if let sel = selector {
                 candidates = candidates.filter(sel)
             }
```
saelo, line 321: Don't you also need to handle lines 324 - 330? Probably you could just do something like `let visibleVariables = fromOuterScope ? scopes.reduce([], +) : scopeAnalyzer.visibleVariables` or so?

amarekano:
Review comments addressed should be good to go.

Sources/Fuzzilli/Mutators/InputMutator.swift
```diff
@@ -28,7 +28,7 @@ public class InputMutator: BaseInstructionMutator {
         // Replace one input
         let selectedInput = Int.random(in: 0..<instr.numInputs)
         b.trace("Mutating input \(selectedInput)")
-        inouts[selectedInput] = b.randVar()
+        inouts[selectedInput] = instr.isBlockEnd ? b.randVarFromOuterScope() : b.randVar()
```
saelo, line 31: Nit: please add a short comment here like "Inputs to block end instructions must be taken from the outer scope since the scope closed by the instruction is currently still active."

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -302,23 +302,31 @@ public class ProgramBuilder {
...
 
+    /// Returns a random variable from the outer scope.
+    public func randVarFromOuterScope() -> Variable {
+        assert(hasVisibleVariables)
+        return randVarInternal(fromOuterScope: true)!
+    }
+
     /// Returns a random variable satisfying the given constraints or nil if none is found.
-    private func randVarInternal(_ selector: ((Variable) -> Bool)? = nil) -> Variable? {
+    private func randVarInternal(_ selector: ((Variable) -> Bool)? = nil, fromOuterScope: Bool = false) -> Variable? {
```
saelo, line 312: Since this variable selection logic is pretty sensitive and bugs in it wouldn't necessarily be detected, could you add a few short tests for this that just do something like:

- Open 2-3 scopes, have a single variable in the innermost scope (the outer ones are empty), assert that `randVar` gives you that one (and maybe that `randVarInternal(fromOuterScope: true)` returns nil if that's easy to do)
- Open e.g. 3 scopes, have at least one variable in the innermost scope and a single variable in the 2nd scope, assert that `randVarFromOuterScope` gives you that one
- Open 2-3 scopes, have a couple variables in each, assert that `randVarInternal({ $0 == X })` gives you the variable X for some variable X from each scope. You might need to remove `private` from this method for that to work, but I think that'd be fine.
- Same as above, but with `randVarInternal(fromOuterScope: true, { $0 == X })`

That'd be great, thanks!

amarekano:
Review comments addressed and testcases added, should be good to go here!

Note: I had to change the visibility on randVarInternal for some of the tests

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -184,6 +184,81 @@ class ProgramBuilderTests: XCTestCase {
         let float3 = b.reuseOrLoadFloat(4.2)
         XCTAssertNotEqual(floatOutOfScope!, float3)     // Variable went out of scope
     }
+
+    // Open 2-3 scopes, have a single variable in the innermost scope (the outer ones are empty),
```
saelo, line 188: Here and below, did you mean to leave these comments in here?
I guess they can stay here, but can be less vague (e.g. "Open 3 scopes", drop the "maybe ... if that's easy to do", etc.)

amarekano: ah no, these were mainly for my reference. I'll get rid of them now

PR #262, opened 2021-09-10
Implement LoadThis and LoadArguments operations

amarekano:


Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -319,6 +319,12 @@ public struct AbstractInterpreter {
         case is LoadNull:
             set(instr.output, .undefined)
 
+        case is LoadThis:
+            set(instr.output, .unknown)
```
saelo, line 323: I think we can be sure that it'll at least be `.object()`

saelo (commented):
Nice, thanks! I guess we should then also remove the pseudo-builtins for these two: https://github.com/googleprojectzero/fuzzilli/blob/590b6b0df2c0c467d3686da2ce2786385a61b8f0/Sources/Fuzzilli/Core/JavaScriptEnvironment.swift#L172

And I think you still need to add the new CodeGenerators to the CodeGeneratorWeights :)

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
amarekano: Sounds good to me!

PR #266, opened 2021-09-17
Remove duplicate case for CallFunction lifting

amarekano:
My bad, this one is on me! I forgot to clean up the previous `CallFunction` case when I pushed my spread element changes :(

### DIFF
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -337,10 +337,6 @@ public class JavaScriptLifter: Lifter {
             case is Await:
                 output = UnaryExpression.new() <> "await " <> input(0)
 
-            case is CallFunction:
-                let arguments = instr.inputs.dropFirst().map({ expr(for: $0).text })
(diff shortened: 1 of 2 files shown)


saelo:
Oh, nice catch! Maybe add a testcase:
```swift
    func testCallWithSpreadLifting(){
        let fuzzer = makeMockFuzzer()
        let b = fuzzer.makeBuilder()

        var initialValues = [Variable]()
        initialValues.append(b.loadInt(15))
        initialValues.append(b.loadInt(30))
        initialValues.append(b.loadString("Hello"))
        initialValues.append(b.loadString("World"))
        let v4 = b.createArray(with: initialValues)
        let v5 = b.loadFloat(13.37)

        let v6 = b.loadBuiltin("Array")
        let _ = b.call(v6, withArgs: [v4,v5], spreading: [true,false])

        let program = b.finalize()

        let lifted_program = fuzzer.lifter.lift(program)
        let expected_program = """
        const v4 = [15,30,"Hello","World"];
        const v7 = Array(...v4,13.37);
        
        """

    }
```
as well. This should've caught this, shouldn't it?

amarekano:
Testcase added, the new one will catch this bug.

saelo:
Great, thanks!

PR #267, opened 2021-09-18
Fuzzilli: use `libsocket.socket_recv` instead of `read`

compnerd:
This is a cleanup change that uses `socket_recv` instead of a raw `read`
from the libc.  THe intent here is to be able to support Windows which
has a separate sockets library  which is not part of libc.


saelo:
Ah great, thanks!

PR #268, opened 2021-09-19
Fuzzilli: replace `sleep` with `Thread.sleep(forTimeInterval:)`

compnerd:
Windows does not have a `sleep` function in the C library, though it does
provide a `Sleep` function in the Windows SDK.  Rather than having two separate
paths here, use the Foundation provided wrapper `Thread.sleep(forTimeInterval:)`
as a portable alternative.


saelo:
Thanks!

PR #269, opened 2021-09-19
libsocket: remove extraneous header includes (NFC)

compnerd:
This is simply a clean up change that removes unnecessary headers from being
included.


saelo:
nice, thanks!

PR #272, opened 2021-09-19
libsocket: port to Windows

compnerd:
This adds support for Windows to libsocket.


compnerd:
I'm not sure how to best handle this.  The API surface for Windows is sufficiently different that I simply effectively forked the file.  We could, I suppose guard the entire file with `#if !defined(_WIN32)` and add a separate file if that is preferable.  There are different approaches to organization and I think that everyone has their own preferences.  I don't really mind which approach is preferred if there is a preference.  I just did the simple brain-dead thing, because, well, laziness prevails.

saelo:
Nice, thanks for your help here and in the other PRs, really appreciated! 
So I think personally, I'd prefer to split socket.c into something like socket-posix.c and socket-win.c and guard the entire files with ifdefs as you suggested. I'm assuming that the build system will automatically compile any .c file it finds in the directory, but if that's not the case and making it work with multiple .c files is more tricky, then I'm also fine with just a single socket.c file.

compnerd:
You're welcome, I'm happy to help :)  Sure, splitting it up into two files is fine, I'll do that today.

PR #273, opened 2021-09-19
libreprl: rename `stdout` and `stderr` fields (NFC)

compnerd:
This renames the `stdout` and `stderr` fields to `child_stdout` and
`child_strerr` respectively.  `stdout` and `stderr` are reserved by the C
standard library, which may define them using a macro.  In such a case, the
re-use of the identifier literal as a field will be improperly expanded by the
CPP.  This actually occurs in practive with ucrt, which uses macros for the
indirection of the iostreams.  This makes the code more resilient to different
libc implementations.


saelo:
Thanks!

PR #276, opened 2021-09-21
librerpl: port to Windows

compnerd:


saelo:
Awesome! Yeah I think I'd prefer splitting it up into at least libreprl-posix and libreprl-win32 again. There seem to be almost no functions that don't need `ifdef`s, but those that don't could, I guess, move into a libreprl-common if that makes sense? But I'm also fine with just duplicating them.

compnerd:
I'll try to update this soon, sorry for the delay here.  I still have a few other patches that I am trying to drain.

saelo:
Sure, absolutely no worries!

Sources/libreprl/libreprl-windows.c
```diff
@@ -0,0 +1,439 @@
...
+    reprl_terminate_child(ctx);
+    if (bSuccess)
+        return status & 0xffff;
+    return reprl_error(ctx, "child in weird state after execution");
+}
+
+/// The 32bit REPRL exit status as returned by reprl_execute has the following format:
+///     [ 00000000 | did_timeout | exit_code | terminating_signal ]
+/// Only one of did_timeout, exit_code, or terminating_signal may be set at one time.
+int RIFSIGNALED(int status)
```
saelo, line 384: Would it make sense to move these 5 functions into a libreprl-common.c file?

saelo (commented):
One more question, but I think then this is ready to be merged. WDYT?

Sources/libreprl/libreprl-windows.c
compnerd: I was actually thinking that these functions don't make sense to have at all - we should put them into the header as `static inline`.  I didn't want to merge that into this change, but intended to do that as a follow up.

compnerd: #281 does that :)

PR #277, opened 2021-09-23
Fuzzilli: adapt `NetworkSync` for libsocket changes

compnerd:
Update the `NetworkSync` implementation to use the new `socket_t` type.
Additionally, adjust the dispatch usage to use the Windows extensions to
properly handle the socket source registration.

### DIFF
--- a/Sources/Fuzzilli/Modules/NetworkSync.swift
+++ b/Sources/Fuzzilli/Modules/NetworkSync.swift
@@ -78,7 +78,7 @@ protocol MessageHandler {
 /// A connection to a network peer that speaks the above protocol.
 class Connection {
     /// File descriptor of the socket.
-    let socket: socket_t
+    let socket: libsocket.socket_t
     
     // Whether this connection has been closed.
     private(set) var closed = false
@@ -104,7 +104,7 @@ class Connection {
     /// Pending outgoing data. Must only be accessed on this connection's dispatch queue.
     private var sendQueue: [Data] = []
     
-    init(socket: socket_t, handler: MessageHandler) {
+    init(socket: libsocket.socket_t, handler: MessageHandler) {
         self.socket = socket
         self.handler = handler
         self.queue = DispatchQueue(label: "Socket \(socket)")
@@ -302,9 +302,9 @@ class Connection {
 public class NetworkMaster: Module, MessageHandler {
     /// File descriptor of the server socket.
 #if os(Windows)
-    private var serverFd: socket_t = INVALID_SOCKET
+    private var serverFd: libsocket.socket_t = INVALID_SOCKET
 #else
-    private var serverFd: socket_t = -1
+    private var serverFd: libsocket.socket_t = -1
 #endif
     
     /// Associated fuzzer.
@@ -323,7 +323,7 @@ public class NetworkMaster: Module, MessageHandler {
     private var serverQueue: DispatchQueue? = nil
     
     /// Active workers. The key is the socket filedescriptor number.
-    private var workers = [socket_t: Worker]()
+    private var workers = [libsocket.socket_t: Worker]()
     
     /// Since fuzzer state can grow quite large (> 100MB) and takes long to serialize,
     /// we cache the serialized state for a short time.
@@ -397,7 +397,7 @@ public class NetworkMaster: Module, MessageHandler {
         }
     }
 
-    private func handleNewConnection(_ socket: socket_t) {
+    private func handleNewConnection(_ socket: libsocket.socket_t) {
         guard socket > 0 else {
             return logger.error("Failed to accept client connection")
         }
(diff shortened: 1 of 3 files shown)


Sources/Fuzzilli/Modules/NetworkSync.swift
```diff
@@ -78,7 +78,7 @@ protocol MessageHandler {
 /// A connection to a network peer that speaks the above protocol.
 class Connection {
     /// File descriptor of the socket.
```
saelo, line 80: Nit: maybe update this comment as well and state that it'll be a socket file descriptor on POSIX and a HANDLE on Windows? Then the code in lines 112-116 is also clear

Sources/Fuzzilli/Modules/NetworkSync.swift
```diff
@@ -293,7 +301,11 @@ class Connection {
 
 public class NetworkMaster: Module, MessageHandler {
     /// File descriptor of the server socket.
```
saelo, line 303: Nit: "File descriptor/Handle of the ..." maybe

Sources/Fuzzilli/Modules/NetworkSync.swift
```diff
@@ -647,19 +663,34 @@ public class NetworkWorker: Module, MessageHandler {
...
-        var fd: Int32 = -1
+#if os(Windows)
+        var fd: libsocket.socket_t = INVALID_SOCKET
+#else
+        var fd: libsocket.socket_t = -1
+#endif
         for _ in 0..<10 {
             fd = libsocket.socket_connect(masterHostname, masterPort)
+#if os(Windows)
+            if fd == INVALID_SOCKET {
```
saelo, line 674: Shouldn't this be `!=`?

Sources/Fuzzilli/Modules/NetworkSync.swift
```diff
@@ -293,7 +301,11 @@ class Connection {
 
 public class NetworkMaster: Module, MessageHandler {
     /// File descriptor of the server socket.
-    private var serverFd: Int32 = -1
+#if os(Windows)
+    private var serverFd: libsocket.socket_t = INVALID_SOCKET
```
saelo, line 305: I'm wondering, can we also define `INVALID_SOCKET` (or some similar constant) on posix as `-1` (from what I can tell, it's guaranteed to be -1 on error), then simplify a lot of this code here and below to always just use `INVALID_SOCKET` instead of having separate code for Windows and POSIX?

Sources/Fuzzilli/Modules/NetworkSync.swift
```diff
@@ -104,12 +104,16 @@ class Connection {
...
     
-    init(socket: Int32, handler: MessageHandler) {
+    init(socket: libsocket.socket_t, handler: MessageHandler) {
         self.socket = socket
         self.handler = handler
         self.queue = DispatchQueue(label: "Socket \(socket)")
-        
+
+#if os(Windows)
+        self.readSource = DispatchSource.makeReadSource(handle: HANDLE(bitPattern: UInt(socket))!, queue: self.queue)
```
saelo, line 113: Just for my understanding, why is the `HANDLE(bitPattern: UInt(socket))!` required here? Shouldn't `socket` already be a HANDLE on Windows? Or is `SOCKET` != `HANDLE` on Windows?

saelo (commented):
Thanks again!

Sources/Fuzzilli/Modules/NetworkSync.swift
compnerd: `SOCKET` and `HANDLE` are distinct types on Windows.  The reason is that `HANDLE` normally is something vended by the kernel.  `SOCKET` is vended by WinSock, which is just a library.  I do agree with you that the special handling here is ugly.

Sources/Fuzzilli/Modules/NetworkSync.swift
compnerd: I definitely prefer that.  I'll locally define `INVALID_SOCKET` to `-1` on non-Windows targets.

Sources/Fuzzilli/Modules/NetworkSync.swift
saelo: Cool, thanks for the explanation, that makes sense and I'm fine with the code as is! Feel free to add your reply as a comment to it if you want though :)

Sources/Fuzzilli/Modules/NetworkSync.swift
compnerd: Yes :-(.

Sources/libsocket/socket-posix.c
```diff
@@ -112,7 +112,7 @@ socket_t socket_connect(const char* address, uint16_t port) {
     freeaddrinfo(result);
 
     if (addr == NULL) {
-       return -2;
+       return -1;
```
saelo, line 115: Ah yes, I forget that these functions can return other values as well. Good catch! How about we move the `INVALID_SOCKET` definition into socket.h and turn all of these `-1` values into the `INVALID_SOCKET` constant? I think that'd be a bit nicer if it works.
Also, I think `socket_accept` can still return invalid values now.

compnerd: Sunk the definition into `libsocket` and updated the implementation.

PR #279, opened 2021-09-25
libcoverage: port to Windows

compnerd:
Use a file mapping for the page file to serve as the shared memory backing,
creating a mapped view of that file.  This should approximate the SysV shared
memory usage in libcoverage.

### DIFF
--- a/Sources/libcoverage/coverage.c
+++ b/Sources/libcoverage/coverage.c
@@ -19,11 +19,20 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/mman.h>
 #include <sys/stat.h>
+
+#if !defined(_WIN32)
+#include <sys/mman.h>
 #include <sys/time.h>
 #include <sys/types.h>
+#endif
+
+// `unistd.h` is the Unix Standard header.  It is available on all unices.
+// macOS wishes to be treated as a unix platform though does not claim to be
+// one.
+#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
 #include <unistd.h>
+#endif
 
 #include "libcoverage.h"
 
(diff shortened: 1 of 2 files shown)


compnerd:
@saelo this change is somewhat more interesting.  I don't think that this is as invasive as the other library changes.  I don't mind changing this to splitting out the files, but I'm not sure if that is as valuable here..

Sources/libcoverage/coverage.c
```diff
@@ -19,11 +19,17 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#if !defined(_WIN32)
```
saelo, line 22: Nit: I think I would slightly prefer to have fewer `#if`s here if possible, even if that breaks the sorting of the includes. Maybe we can just have a 2nd block of includes inside a single `#if !defined(_WIN32)` after line 33, would that work?
If not, I'm also fine leaving it as it is though!

Sources/libcoverage/coverage.c
```diff
@@ -19,11 +19,17 @@
...
 #include <sys/mman.h>
+#endif
 #include <sys/stat.h>
+#if !defined(_WIN32)
 #include <sys/time.h>
+#endif
 #include <sys/types.h>
+#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
```
saelo, line 30: Why is this not `#if !defined(_WIN32)` as above? Maybe add a short comment since I'm sure I'll ask myself that question again the next time I see this code =)

saelo (commented):
Nice! Yeah I agree, in this case I think it makes sense to just keep the two versions in a single file!

Sources/libcoverage/coverage.c
compnerd: Ah, so, this is purely pedantry :-D.  `unistd.h` is the *Unix* standard header.  That only makes sense to import on Unix platforms.  The identification for such platforms is ... `__unix__`.  The thing is, MacOS X and new wants to be treated as a unix but doesn't declare itself to be unix based.  So, we use `__APPLE__` to identify if we are targeting an Apple platform, and if it is using mach,, we assume that it is a Unix-like environment instead of older MacOS platforms.  Technically, I suppose that A/UX and Rhapsody would also support `unistd.h` but the latter never saw release (and thus I cannot verify if it ever declared to be unix-esque) and the former has been discontinued.  In theory, if someone wanted to port this to an alternative operating system (e.g. SerenityOS) this ensures that this is closer to correct.

Sources/libcoverage/coverage.c
compnerd: SGTM, updated.

Sources/libcoverage/coverage.c
compnerd: Added a brief version of this explanation as well.

saelo: Haha that's great, thanks for the explanation and comment! :D

PR #281, opened 2021-09-26
libreprl: extract macro-like functions

compnerd:
The status check "macro" functions are meant to simply encode the bit operations
required for checking the exit code.  There is no value for them to be
out-of-line, they can be fully inlined.  By placing them in the header, this can
be fully internalized into the Swift module as well.


saelo:
Nice, thanks!

PR #282, opened 2021-09-26
workflows: add Linux build

compnerd:
This adds Linux coverage to the CI to prevent accidental breakage.  The previous
changes to support Windows seems to have accidentally regressed Linux builds.
Add coverage to prevent that in the future.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -1,4 +1,4 @@
-name: Swift
+name: build
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


compnerd:
#281 will also fix the Linux build.

.github/workflows/swift.yml
```diff
@@ -1,4 +1,4 @@
-name: Swift
+name: build
```
saelo, line 1: Nit: maybe something like "test" or "RunTests" or so would be more descriptive? Also feel free to rename the file as well if that makes sense

saelo (commented):
Great! I saw a previous build failure on linux because `memfd_create` didn't exist. Was that maybe due to an older linux version? Did you change anything since?
It seems to work fine now though, so I'm happy to merge this in any case.

compnerd:
Yes, the change was in #281 for `_GNU_SOURCE` to be defined earlier.

I can certainly rename the file and job.  I tend to find build to be descriptive as it is the build test (where you may have other jobs for linting or documentation generation).  I think with quoting I can change it to "build and test", any suggestions for the file?

saelo:
Cool! Sounds good, so let's just go with "build" then :)

PR #291, opened 2021-10-10
Enhance Object creation

amarekano:
This is a ~WIP~ PR to address #253. 

On my TODO list I've still got the following to get through:
~* Update the AI to handle the new object creation ops~
~* More test for lifting, splicing and AI type tracking~
~* Remove the legacy object generators~
~* Potentially have additional generators for props, getters and setters.~ (Not possible with the current model)

Feel free to add your thoughts to these initial changes and I'll incorporate them as we progress with this PR.


amarekano:
@saelo this PR is ready for review, let me know your thoughts on the modelling and the handling of splicing in restricted contexts please. If you're happy with this approach, I'll go ahead and get started with implementing #8 using the same approach.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -208,17 +208,77 @@ class LoadRegExp: Operation {
...
-        }
-        super.init(numInputs: propertyNames.count, numOutputs: 1, attributes: flags)
+class BeginObjectDefinition: Operation {
+    init() {
+        super.init(numInputs: 0, numOutputs: 1, attributes: [.isBlockBegin], contextOpened: .objectDefinition)
+    }
+}
+
+class CreateProperty: Operation {
+	let propertyName: String
```
saelo, line 218: Nit: wrong indention (please use 4 spaces)

Tests/FuzzilliTests/AnalyzerTest.swift
```diff
@@ -126,7 +126,9 @@ class AnalyzerTests: XCTestCase {
             cls.defineConstructor(withParameters: [.string]) { _ in
                 XCTAssertEqual(b.context, [.script, .classDefinition, .function])
                 let v0 = b.loadInt(42)
-                let v1 = b.createObject(with: ["foo": v0])
+                let v1 = b.createObject { prop in
```
saelo, line 129: Let's unify the closure argument (`prop`) in this PR. How about just `obj` (if we also support defining methods at least)?
```swift
let v1 = b.createObject { obj in
    obj.addProperty("foo", withValue: v0)
}
```

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1069,31 +1215,156 @@ public class ProgramBuilder {
...
     }
 
-    @discardableResult
-    public func createObject(with initialProperties: [String: Variable]) -> Variable {
-        // CreateObject expects sorted property names
-        var propertyNames = [String](), propertyValues = [Variable]()
-        for (k, v) in initialProperties.sorted(by: { $0.key < $1.key }) {
-            propertyNames.append(k)
-            propertyValues.append(v)
+    public struct Property {
```
saelo, line 1218: What about methods? Do we also want to support those?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1069,31 +1215,156 @@ public class ProgramBuilder {
...
+            self.body = body
+        }
+    }
+
+    public struct ObjectBuilder {
+        public typealias GetterFunctionGenerator = () -> ()
+        public typealias SetterFunctionGenerator = (Variable) -> ()
+        fileprivate var properties: [Any] = []
+
+        public mutating func addProperty(_ name: String, v: Variable) {
```
saelo, line 1295: Nit: let's maybe use this signature here and below `(_ name: String, withValue v: Variable)`

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1069,31 +1215,156 @@ public class ProgramBuilder {
...
-        for (k, v) in initialProperties.sorted(by: { $0.key < $1.key }) {
-            propertyNames.append(k)
-            propertyValues.append(v)
+    public func createObject(body: (inout ObjectBuilder) -> ()) -> Variable {
+        var builder = ObjectBuilder()
+        body(&builder)
+
+        let instr = perform(BeginObjectDefinition(), withInputs: [])
+
+        for prop in builder.properties {
```
saelo, line 1331: I'm wondering if we can simplify this: how about having the `addXYZ` methods of the `ObjectBuilder` just append closures into a list:
```swift
public mutating func addProperty(_ name: String, withValue v: Variable) {
    generators.append({ b in b.perform(CreateProperty(propertyName: name), withInputs: [v]) })
}
```
Then just turn this into
```swift
for generator in builder.generators {
    generator(self)
}
```

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -55,6 +58,15 @@ public struct AbstractInterpreter {
         executeOuterEffects(instr)
 
         switch instr.op {
+        case is BeginObjectDefinition,
+            is BeginGetter,
```
saelo, line 62: Is this correct? I think only the `BeginObjectDefinition` creates a child state, the others create sibling states? Not 100% sure though.
A child state basically builds onto the parent state, so e.g. inside a `BeginIf`, the changes from the previous code are still visible. On the other hand, a sibling state discards the changes from its other siblings. E.g. inside the `else` branch of an `if`, the changes from the `if` body are no longer valid

Sources/Fuzzilli/Mutators/CodeGenMutator.swift
```diff
@@ -31,6 +31,10 @@ public class CodeGenMutator: BaseInstructionMutator {
     
     public override func mutate(_ instr: Instruction, _ b: ProgramBuilder) {
         b.adopt(instr, keepTypes: true)
+        guard b.context.contains(.script) else {
```
saelo, line 34: I think this should go into `canMutate` and just be an `assert` here

Sources/Fuzzilli/Mutators/CombineMutator.swift
```diff
@@ -30,6 +30,10 @@ public class CombineMutator: BaseInstructionMutator {
     public override func mutate(_ instr: Instruction, _ b: ProgramBuilder) {
         b.adopt(instr, keepTypes: true)
         let other = b.fuzzer.corpus.randomElementForSplicing()
+        guard b.context.contains(.script) else {
```
saelo, line 33: Same here

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -21,7 +21,7 @@ class ProgramBuilderTests: XCTestCase {
         let fuzzer = makeMockFuzzer()
         let b = fuzzer.makeBuilder()
 
-        for _ in 0..<1000 {
+        for _ in 0..<10000 {
```
saelo, line 24: maybe reset this back to 1000?

Tests/FuzzilliTests/ProgramBuilderTest.swift
```diff
@@ -313,7 +330,9 @@ class ProgramBuilderTests: XCTestCase {
         b.defineClass(withSuperclass: superclass) { cls in
             cls.defineConstructor(withParameters: [.string]) { _ in
                 let v0 = b.loadInt(42)
-                let v1 = b.createObject(with: ["foo": v0])
+                let v1 = b.createObject{ prop in
```
saelo, line 333: I kind of feel like we should have some convenience constructors since this "create object with normal properties" is very common. Can we just create an overload of `createObject` in the ProgramBuilder with the old signature?

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -827,12 +833,152 @@ public class ProgramBuilder {
             return
         }
 
-        // Finally, insert the slice into the current program.
-        for instr in source {
-            if slice.contains(instr.index) {
-                adopt(instr, keepTypes: true)
+
```
saelo, line 836: Uff.. :'D Can you give an example for why this additional logic is required? I think that'd help.

Ideally this method would be as simple as possible since it's pretty important to the overall fuzzer performance and somewhat hard to automatically test for correctness.

Maybe we need a different algorithm to handle splicing instead of trying to make the old algorithm work under new constraints. For example, would it help to do multiple passes and precompute some bits of information first, for example block groups, dataflow dependencies, or active contexts?
I'll try to think of something once I better understand the issue this new logic is trying to handle.

saelo (commented):
Did a first review round now. It's a pretty huge CL... :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: Here's an example that I've tried to simplify to explain the problem. Assume we have the following program that we want to splice at instruction `CallComputedMethod` which is at index 86, into a default context (.script):

```js
Splicing instruction 86 (CallComputedMethod) from 1F7EB535-2872-41AB-B2AF-170A9A2ED6FB
Instruction(op: Fuzzilli.LoadInteger, inouts_: [v0, v0])
...
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v6, v6])
Instruction(op: Fuzzilli.CreateArray, inouts_: [v6, v7, v7])
...
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v10, v10])
Instruction(op: Fuzzilli.CallFunction, inouts_: [v10, v11, v11])
...
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v26, v31])
...
Instruction(op: Fuzzilli.CreateArray, inouts_: [v0, v0, v31, v45])
Instruction(op: Fuzzilli.LoadInteger, inouts_: [v32, v46])
Instruction(op: Fuzzilli.StoreProperty, inouts_: [v31, v32, v47])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v33, v48])
Instruction(op: Fuzzilli.CallFunction, inouts_: [v33, v34, v49])
...
Instruction(op: Fuzzilli.BeginObjectDefinition, inouts_: [v38, v55])
Instruction(op: Fuzzilli.CreateSpreadProperty, inouts_: [v26, v56])
Instruction(op: Fuzzilli.CreateComputedProperty, inouts_: [v14, v32, v57])
Instruction(op: Fuzzilli.CreateProperty, inouts_: [v5, v58])
Instruction(op: Fuzzilli.BeginGetter, inouts_: [v59])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v39, v60])
Instruction(op: Fuzzilli.CallFunction, inouts_: [v39, v40, v61])
Instruction(op: Fuzzilli.EndGetterSetter, inouts_: [v62])
Instruction(op: Fuzzilli.BeginGetter, inouts_: [v63])
...
Instruction(op: Fuzzilli.CallMethod, inouts_: [v41, v31, v0, v10, v51, v72])
Instruction(op: Fuzzilli.EndGetterSetter, inouts_: [v73])
...
Instruction(op: Fuzzilli.BeginComputedSetter, inouts_: [v10, v53, v78])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v54, v79])
Instruction(op: Fuzzilli.LoadProperty, inouts_: [v54, v55, v80])
Instruction(op: Fuzzilli.StoreComputedProperty, inouts_: [v12, v55, v26, v81])
Instruction(op: Fuzzilli.LoadInteger, inouts_: [v56, v82])
Instruction(op: Fuzzilli.CreateArray, inouts_: [v56, v57, v83])
Instruction(op: Fuzzilli.StoreProperty, inouts_: [v57, v57, v84])
Instruction(op: Fuzzilli.LoadString, inouts_: [v58, v85])
Instruction(op: Fuzzilli.CallComputedMethod, inouts_: [v55, v58, v26, v7, v11, v33, v53, v6, v59, v86])
...
Instruction(op: Fuzzilli.EndGetterSetter, inouts_: [v89])
Instruction(op: Fuzzilli.EndObjectDefinition, inouts_: [v92])
```

With the existing splicing logic we end up with the following slice of instructions to be inserted:

```js
Instruction(op: Fuzzilli.LoadInteger, inouts_: [v0, v0])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v6, v6])
Instruction(op: Fuzzilli.CreateArray, inouts_: [v6, v7, v7])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v10, v10])
Instruction(op: Fuzzilli.CallFunction, inouts_: [v10, v11, v11])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v26, v31])
Instruction(op: Fuzzilli.CreateArray, inouts_: [v0, v0, v31, v45])
Instruction(op: Fuzzilli.LoadInteger, inouts_: [v32, v46])
Instruction(op: Fuzzilli.StoreProperty, inouts_: [v31, v32, v47])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v33, v48])
Instruction(op: Fuzzilli.BeginObjectDefinition, inouts_: [v38, v55])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v41, v64])
Instruction(op: Fuzzilli.CallMethod, inouts_: [v41, v31, v0, v10, v51, v72])
Instruction(op: Fuzzilli.BeginComputedSetter, inouts_: [v10, v53, v78])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v54, v79])
Instruction(op: Fuzzilli.LoadProperty, inouts_: [v54, v55, v80])
Instruction(op: Fuzzilli.LoadString, inouts_: [v58, v85])
Instruction(op: Fuzzilli.CallComputedMethod, inouts_: [v55, v58, v26, v7, v11, v33, v53, v6, v59, v86])
Instruction(op: Fuzzilli.EndGetterSetter, inouts_: [v89])
Instruction(op: Fuzzilli.EndObjectDefinition, inouts_: [v92])
```

Now if we were to adopt the instructions in sequential order, we would end up with syntactically invalid FuzzIL. This is because when we insert instruction `BeginObjectDef` it opens a restricted context and the instructions `LoadBuiltin` at index 64 and  and `CallMethod` at 72 are now going to be inserted into this restricted context (i.e. objectDefinition).

```
Instruction(op: Fuzzilli.BeginObjectDefinition, inouts_: [v38, v55])
Instruction(op: Fuzzilli.LoadBuiltin, inouts_: [v41, v64])
Instruction(op: Fuzzilli.CallMethod, inouts_: [v41, v31, v0, v10, v51, v72])
```

The reason these two instructions are captured in the slice is because they mutate an input which has a data dependency to one of the instructions included in the slice (in this instance `v10`).

To correctly handle these situations, we would need to hoist the defaulting instructions out of the restricted context and into a suitable instruction block. A suitable instruction block would be an open block (i.e we haven't processed its BlockEnd instruction yet) which opens and/or propagates a context that these instructions can be inserted into.

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: I suppose we can support methods natively. The existing approach to generate a function and then add it as an input to a property worked for us so I didn't think it necessary to add methods as a bunch of separate operations.

Having a rethink about this, adding methods would probably make type tracking easier (i.e. we can bundle this functionality in ObjUtils.swift and expose one API to get property and method information)? and we will now be able to use `super` within object method definitions. WDYT?

amarekano: On the flip side though, we'd have to generate object method operations for each function variant (i.e. PlainFunc, Async, Generator, etc). And at some point we'd also have to extend this to classes.

Tests/FuzzilliTests/ProgramBuilderTest.swift
amarekano: Can do, thinking along the same lines, should we also have multiple CodeGenerators as well (e.g. generate an object with no props, objects with just data props, objects will data props +getters/setters)? Currently it's all just under one generator.

Sources/Fuzzilli/Mutators/CodeGenMutator.swift
amarekano: Currently `canMutate` doesn't have a reference to the active `ProgramBuilder` instance. We can pass this as a parameter along with the instruction but that would mean changing `canMutate` in the parent class (i.e. `BaseInstructionMutator`). Should I go ahead and do that?

saelo: Ah right, yeah that won't work because the `ProgramBuilder` will be empty and so won't be able to provide useful context information :( Maybe for the splicing algorithm it'll turn out to be useful/necessary to have context information associated with every instruction in a `Program` (i.e. a `Array<Context>` in every `Program`), and then we'd also have it available here. So let's revisit this later

Tests/FuzzilliTests/ProgramBuilderTest.swift
saelo: I think I'm fine with one generator that can generate everything :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Right. I think we could go with just a single `addMethod` for now, then see how that plays out, and then see if we still want to support other types of methods as well later on

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Cool, that makes sense, thanks! Yeah I think we should try and come up with a simpler approach for computing slices, or this'll become unmaintainable :(

I think it could be possible to simplify things if we precompute for every instruction what the currently active context is (which we might need elsewhere, too, see other comment), and maybe also precompute all block instructions that belong together. Maybe we can further simplify things by completely ignoring blocks that don't surround the current slice (in your example above, the `BeginGetter` block), but I think it should work either way. If we have the context information for every instruction, we for example know whether we need to include a block begin instruction: if it opens a context that's required and doesn't exist at the instruction preceding it. Although we might anyway always want to select the closest block that creates a certain context, then it wouldn't really matter I think.

Another simplification might be to always include blocks entirely (with all instructions inside of them). This could make sense: if something depends e.g. on an argument to a class method, it probably doesn't really make sense to leave all other methods of the class empty. On the other hand, it might make slices significantly larger. I probably wouldn't mind that though if it kept the algorithm simpler. If we do that, things become a bit simpler: if we see an instruction with special context requirements (!= .script), then we know that we'll also include all instruction up to the next block that starts that context. Then we don't really need to push/pop entries onto/off our context stack until we find that opening block. Once we find it, we could "jump" to the end instruction (if we precomputed those), then include all instruction from there on up to the point where we started. Still without having to modify the context stack.

Finally, I think we should record more statistics about slices, as a start, maybe the total size of the slice and the relative size compared to the program it is taken from (`slice.count / program.count`).

But yeah, it's still all pretty vague. I'll try to come up with a proper algorithm. Let me know your thoughts!

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
amarekano: Thanks for the explanation, I think here (under `execute`) we don't need to handle state changes for `BeginObjectDef` and `EndObjectDef`. Although we are starting a new block, some of the non-block-opening instructions in the block (e.g. `CreateProperty`) don't affect the parent state or the sibling states. We only need to care about getter and setter blocks since they affect the parent state.

Note: `BeginObjDef` and `EndObjDef` are being analysed in `executeOuterEffects` so that we can track properties that have been added to the object being created.

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: For now I've added a `addMethod`, `addComputedMethod` and `addGeneratorMethod` functions. The first two occur quite frequently in the JSC test repo and there's a few samples that use generator methods. At some point I'd like to refactor class creation so that we can reuse method and property creation code from object creation.

saelo:
This is now implemented with https://github.com/googleprojectzero/fuzzilli/commit/840c09e07f342c2210694ececd23ef8ac3d669d6 so I'm closing this PR. Let me know if you see any features still missing! Thanks!

PR #294, opened 2021-11-01
Add support for JS void operator

amarekano:


saelo:
Thanks! So this is a bit weird: IIUC the `void` operator really only makes sense on a syntactical level: it evaluates its expression, then essentially drops the return value and replaces it with `undefined`. Now in FuzzIL, this doesn't really make that much sense:
```
v7 = SomeOperation
v8 = Void v7
```
There cannot (by construction) be anything to evaluate for the `Void` operation because it only receives a variable as input, not an expression. So you might as well just do `v8 = LoadUndefined`. On the other hand, it then does kind of make some sense again when the code is lifted back to JS, _but only if certain expressions are inlined_: `let v8 = void some_operation();`.

IMO Fuzzilli shouldn't attempt to fuzz the parser of the JS engine. You'd need a differently designed fuzzer for that. My current assumption is that any `void` operator is removed by the parser during bytecode generation and never makes it into the interpreter or a JIT compiler. In that case, and with similar reasoning as in https://github.com/googleprojectzero/fuzzilli/pull/297, I think I'd then be more in favor of just having the compiler deal with this, by turning `let v8 = void some_operation();` into 
```
v7 = SomeOperation
v8 = LoadUndefined
```
On the other hand, if the `void` operator does make it into JS bytecode or even JIT intermedate languages, then I think we should have it in FuzzIL as well. WDYT?

In any case, since you already have the code definitely try fuzzing with it for a bit and see if anything interesting comes up!

amarekano:
> On the other hand, if the void operator does make it into JS bytecode or even JIT intermedate languages, then I think we should have it in FuzzIL as well. WDYT?

Yeah so this was the main motivation (i.e. generate different bytecode/ JIT IR). Let me run a few tests and get back to you with data on bytecode generation/JIT IR.

amarekano:
So `void` has almost no effect on the bytecode generated (atleast for JSC). I've compared bycode generated for the following two samples:
```js
let x = function() { return 10 }
let y = void x()
```
and
```js
let x = function() { return 10 }
x()
let y = undefined
```
From diffing the bytecode dumps, `void` is essentially decomposed to a bunch of `mov undefined` operations. There's no new bytecode opcodes being generated for `void` and I doubt there's going to be JIT IL generated for the same. I think I'll close this PR and we can compile calls to `void` as you've mentioned in your previous comment.

Issue #213, opened 2021-06-23
AbstractInterpreter handles logical operators incorrectly


saelo:
The AbstractInterpreter currently assumes that the [result of binary logical operators is a .boolean](https://github.com/googleprojectzero/fuzzilli/blob/87e9142747dd76387ca9d0c8d02acf3a78c5196c/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift#L430). While that probably seems intuitive, that's not how JavaScript works:

```JavaScript
let a = 0 || "foobar";
print(a);     // "foobar"
let b = "" && 42;
print(b);     // ""
```

The correct semantics are explained for example [here for `||`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR) and [here for `&&`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND), but in essence, the operators always yield the value of one of the two input expressions, so probably the resulting type should simply be the union of the two inputs.

Dhananjayshukla1:
@saelo I am new to this Organisation, can you please assign this to me.

Dhananjayshukla1:
@amarekano I am new to this Organisation, can you please assign this to me

Dhananjayshukla1:
@saelo I want to work on this issue, can you assign this to me.

saelo:
That's great, thanks! We usually don't assign issues, but happy to do so if you like!

Dhananjayshukla1:
Sorry sir,
I forgot my github account so i was not able to reply your message and your valuable time.
let a = 0 || "foobar";
console.log(a);     // "foobar"
let b = "" && 42;
console.log(b);     // ""

Dhananjayshukla1:
I'm new to open source contributions but I'm well versed with C++, HTML, CSS, Javascript, React, and Node.js.
could you please tell me how to get started.

nfgrusk:
Looks like this issue was resolved a good while ago (see #295). Perhaps, it’s time to close it. 😅

saelo:
Indeed, looks like that fixes the issue. Thanks for pointing that out!

PR #295, opened 2021-11-02
Improve AI handling of logical operators

amarekano:
This PR should address #213 and some minor code refactoring.

### DIFF
--- a/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
+++ b/Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
@@ -276,6 +276,30 @@ public struct AbstractInterpreter {
         // basic types.
         // TODO: fetch all output types from the environment instead of hardcoding them.
 
+        // Helper function to set output type of binary/reassignment operations
+        func analyzeBinaryOperation(operator op: BinaryOperator, withInputs inputs: [Variable]) -> Type {
+            switch op {
+            case .Add:
+                return maybeBigIntOr(.primitive)
+            case .Sub,
+                 .Mul,
+                 .Exp,
+                 .Div,
+                 .Mod:
+                return maybeBigIntOr(.number)
+            case .BitAnd,
+                 .BitOr,
+                 .Xor,
+                 .LShift,
+                 .RShift,
+                 .UnRShift:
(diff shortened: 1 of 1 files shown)


Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -276,6 +276,30 @@ public struct AbstractInterpreter {
         // basic types.
         // TODO: fetch all output types from the environment instead of hardcoding them.
 
+        // Helper function to set output type of binary/reassignment operations
+        func analyzeBinaryOperation(operator op: BinaryOperator, withInputs inputs: [Variable]) -> Type {
```
saelo, line 280: Can `inputs` be `ArraySlice<Variable>` here so we avoid a copy when calling this function (just `instr.inputs` instead of `Array(instr.inputs)`)?

amarekano: Nice! all done now

amarekano:
Hmm, so I've noticed that intermittently this test fails due to an unexplained 'type confusion':
```
Test Case 'ProgramBuilderTests.testTypeInstantiation' started at 2021-11-03 09:00:22.498
Fuzzilli/ProgramBuilder.swift:520: Assertion failed: Unexpected type encountered .integer | .object()
Current stack trace:
``` 
Any thoughts on what the root cause might be. Perhaps incorrect AI modelling somewhere?
I'll debug this further when I've got some time and put together a fix.

saelo:
Ah yeah, I've noticed these before... I think the issue here is that `generateType` might generate e.g. a `TypedArray` from the JavaScriptEnvironment (that's correct), then `generateVariable` tries to call the corresponding constructor and attempts to obtain argument variables (that's also correct), but for `TypedArrays`, the constructor is defined to take `.integer | .object(...)`: https://github.com/googleprojectzero/fuzzilli/blob/1c33ac3cefb0771978a3f8f1185ac77fca703b14/Sources/Fuzzilli/Core/JavaScriptEnvironment.swift#L390 (also correct), but `generateVariable` can't yet handle type unions (that is, I think, the issue). Feel free to open a separate issue for this :)

PR #296, opened 2021-11-02
Add binary operation variants for StoreProperty, StoreElement, StoreComputedProperty and StoreSuperProperty

amarekano:


saelo (commented):
Great, thanks!

While I guess technically, these can also be implemented using temp variables and existing operations, I do think it makes sense to add these as they should probably improve the efficiency of the fuzzer (less variables to mutate), since it reduces e.g.
```
v7 = LoadProperty v0, 'x'
v8 = BinaryOperation v0, v3, '+'
StoreProperty v0, v8
```
to just
```
StorePropertyWithBinop v0, 'x', v3, '+'
```
This might also allow handling `obj.prop++` better, by converting it to `obj.prop += 1`, which at least is two instructions shorter than the alternative that uses a temporary variable.

amarekano:
Review comments addressed, I've also gone ahead and renamed the `BinaryOpAndReassign` operation instead of adding a TODO.

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -505,6 +508,11 @@ public struct AbstractInterpreter {
         case let op as LoadSuperProperty:
             set(instr.output, inferPropertyType(of: op.propertyName, on: currentSuperType()))
 
+        /* 
+        TODO: Add missing case for StoreSuperPoperty
```
saelo, line 512: What about this? :)

amarekano: So I've been thinking of rewriting the ClassDef building following the same model that I've proposed for ObjectDef building. I'm not sure if there's a trivial way to add AI rules for StoreSuper ops (tbh I haven't given this a great deal of thought) but if you've got some suggestions then I'll look into it.

saelo: Right, yeah I don't think updating the type of superclass properties is easily possible in the current model. So let's just remove these comments or turn them into `TODO support superclass property assignment` and deal with class definitions and superclass properties separately, wdyt?

amarekano: I think I'll turn this into a TODO so that we don't lose track of this piece of work.

PR #297, opened 2021-11-03
Always inline Unary Operations

amarekano:
Currently UnaryOperations are not inlined which make compilation to IL a bit more tedious. Consider the following program:
```
let arr = []
for(let i =0; i< 10; i++){
 arr[i++] = i
}
```
Which would compile to:
```
v0 <- CreateArray []
BeginFor v1 
    v2 <- UnaryOperation v1, '++'
    StoreComputedProperty v1, v2, v0
EndFor
```
Without inlining this IL would lift to:
```
let arr = []
for(let v1 =0; v1< 10; v1++){
 const v2 = v1++
 arr[v2] = v1
}
```
Which is incorrect. Of course this could be solved by generating additional IL instructions to emit a temporary variable and then reassign the temporary variable. Instead a more elegant solution would be to always inline UnaryOperations which saves us from unnecessary IL gymnastics with temporary variables.

### DIFF
--- a/Sources/Fuzzilli/Lifting/JSExpressions.swift
+++ b/Sources/Fuzzilli/Lifting/JSExpressions.swift
@@ -33,7 +33,7 @@ public let NumberLiteral        = ExpressionType(precedence: 17,                
 public let ObjectLiteral        = ExpressionType(precedence: 17,                        inline: .singleUseOnly)
 public let ArrayLiteral         = ExpressionType(precedence: 17,                        inline: .singleUseOnly)
 public let PostfixExpression    = ExpressionType(precedence: 16,                        inline: .singleUseOnly)
-public let UnaryExpression      = ExpressionType(precedence: 15, associativity: .right, inline: .singleUseOnly)
+public let UnaryExpression      = ExpressionType(precedence: 15, associativity: .right, inline: .always)
 public let BinaryExpression     = ExpressionType(precedence: 14, associativity: .none,  inline: .singleUseOnly)
 public let TernaryExpression    = ExpressionType(precedence: 4,  associativity: .none,  inline: .singleUseOnly)
 public let AssignmentExpression = ExpressionType(precedence: 3,                         inline: .never)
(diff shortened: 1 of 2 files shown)


saelo:
Don't these two JavaScript snippet behave identically though? It shouldn't matter whether you do `print(i++)` or `let t = i++; print(t)`.

One thing that does matter though:
```javascript
let a = [1,2,3];
a.length++;
```

vs

```javascript
let a = [1,2,3];
let l = a.length;
l++;
```

I'm not sure if we are handling that correctly currently? Probably it'd need to be compiled to something like

```javascript
let a = [1,2,3];
let l = a.length;
l++;
a.length = l;
```

which isn't super nice, either.

amarekano:
> Don't these two JavaScript snippet behave identically though? It shouldn't matter whether you do print(i++) or let t = i++; print(t).

You're right, in this instance inlining isn't required since they are semantically identical. However if we have a statement like the following:
```js
let x = 10;
let y = x+(x++)
```
we would need to use a temporary variable to lift this to:
```js
let x = 10;
let temp = x
x++
let y = temp  + temp
```

> One thing that does matter though: ...

Yeah, so this is an issue that we will have with LoadProp/LoadComputedProp and LoadElement operations :(
The current inlining approach that I've pushed doesn't address this issue. I've yet to think of a solution to handle these ops but I'm leaning in favour of handling this via the Lifter instead of generating additional IL. This is mainly to make compilation easier and have fewer instructions but I'm open to suggestions on other approaches.

saelo:
Right, thanks for the examples! I'm a bit hesitant to attempt to solve these issues in the Lifter. IMO, the algorithm used by the lifter should have no effect on the semantics of the program. The inlining is only performed to make the samples more readable, but it should ideally also work without any inlining at all. I also think it'd be weird for a generated FuzzIL program like
```
v0 = LoadProperty array, 'length'
v1 = UnaryOperation v0 '++'
```
To change the length of `array` in this example. It's more intuitive if the variable behave like, well, variables, and not potentially as mutable references to object properties :/

So I think even though it adds more instructions, I'm in favor of solving these issues in the compiler... Maybe we could even drop support for `++` and `--` entirely since it leads to these kinds of weirdnesses and that way just force any compiler to turn them into `+= 1` first (which is more or less what should happen in the compilation pipelines of actual JS engines as well, I'd expect). Then we could still add an optimization to the lifter to turn `+= 1` into `++` if possible, but then it is really just that, an optimization that doesn't change the semantics of the program, just makes it a bit more readable.

PR #298, opened 2021-11-12
Add entry for CVE-2021-30818

amarekano:


saelo:
Awesome! Congrats! Just to double check, https://bugs.webkit.org/show_bug.cgi?id=223278 really is the correct URL for this issue?

amarekano:
Apple Security didn't provide a URL so I looked up the commit that added my poc to the WebKit repo and used the bug id URL listed in the commit. I'm not 100% certain that this is the correct URL but since it's the first commit to add the poc, it's probably a good enough placeholder.

I plan to come back to this and update a link with a separate writeup on the bug which will probably be more helpful than the current URL.

saelo:
Cool that makes sense! Looking forward to the writeup! :)

PR #300, opened 2021-11-14
Add DestructArray and DestructArrayAndReassign operations

amarekano:
This implementation currently does not support the following JS constructs:
* ~Nested arrays (i.e. [[a,b],c] = arr)~ To be handled during compilation
* ~holes (i.e [a,,b] = arr)~
* ~rest element (i.e. [a, ...b] = arr)~

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -395,15 +395,27 @@ public let CodeGenerators: [CodeGenerator] = [
 
     CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
         // TODO: We need a way to expose the length of arr so that element selection is accurate
-        b.destruct(arr, selectFirst: Int.random(in: 0...5), hasRestElement: probability(0.2))
+        // Pick random indices
+        var indices: [Int] = []
+        for idx in 0..<Int.random(in: 0..<5) {
+            withProbability(0.7) {
+                indices.append(idx)
+            }
+        }
+
+        b.destruct(arr, selecting: indices, hasRestElement: probability(0.2))
     },
 
     CodeGenerator("DestructArrayAndReassignGenerator", input: .iterable) {b, arr in
         var candidates: [Variable] = []
-        for _ in 0..<Int.random(in: 0..<5) {
-            candidates.append(b.randVar())
+        var indices: [Int] = []
+        for idx in 0..<Int.random(in: 0..<5) {
+            withProbability(0.7) {
+                indices.append(idx)
+                candidates.append(b.randVar())
+            }
         }
-        b.destructAndReassign(candidates, to: arr, hasRestElement: probability(0.2))
+        b.destructAndReassign(candidates, atIndices: indices, from: arr, hasRestElement: probability(0.2))
     },
 
     CodeGenerator("ComparisonGenerator", inputs: (.anything, .anything)) { b, lhs, rhs in
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -394,8 +394,9 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
-        // TODO: We need a way to expose the length of arr so that element selection is accurate
-        // Pick random indices
+        // Fuzzilli generated arrays can have a length ranging from 0 to 10 elements,
+        // We want to ensure that 1) when destructing arrays we are within this length range
+        // and 2) The probability with which we select indices allows defining atleast 2-3 variables.
         var indices: [Int] = []
         for idx in 0..<Int.random(in: 0..<5) {
             withProbability(0.7) {
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -393,6 +393,19 @@ public let CodeGenerators: [CodeGenerator] = [
         b.reassign(target, to: val)
     },
 
+    CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
+        // TODO: We need a way to expose the length of arr so that element selection is accurate
+        b.destruct(arr, selectFirst: Int.random(in: 0...5))
+    },
+
+    CodeGenerator("DestructArrayAndReassignGenerator", input: .iterable) {b, arr in
+        var candidates: [Variable] = []
+        for _ in 0..<Int.random(in: 0..<5) {
+            candidates.append(b.randVar())
+        }
+        b.destructAndReassign(candidates, to: arr)
+    },
+
     CodeGenerator("ComparisonGenerator", inputs: (.anything, .anything)) { b, lhs, rhs in
         b.compare(lhs, rhs, with: chooseUniform(from: allComparators))
     },
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -395,7 +395,7 @@ public let CodeGenerators: [CodeGenerator] = [
 
     CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
         // TODO: We need a way to expose the length of arr so that element selection is accurate
-        b.destruct(arr, selectFirst: Int.random(in: 0...5))
+        b.destruct(arr, selectFirst: Int.random(in: 0...5), hasRestElement: probability(0.2))
     },
 
     CodeGenerator("DestructArrayAndReassignGenerator", input: .iterable) {b, arr in
@@ -403,7 +403,7 @@ public let CodeGenerators: [CodeGenerator] = [
         for _ in 0..<Int.random(in: 0..<5) {
             candidates.append(b.randVar())
         }
-        b.destructAndReassign(candidates, to: arr)
+        b.destructAndReassign(candidates, to: arr, hasRestElement: probability(0.2))
     },
 
     CodeGenerator("ComparisonGenerator", inputs: (.anything, .anything)) { b, lhs, rhs in
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -395,7 +395,7 @@ public let CodeGenerators: [CodeGenerator] = [
 
     CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
         // Fuzzilli generated arrays can have a length ranging from 0 to 10 elements,
-        // We want to ensure that 1) when destructing arrays we are within this length range
+        // We want to ensure that 1) when destructing arrays we are usually within this length range
         // and 2) The probability with which we select indices allows defining atleast 2-3 variables.
         var indices: [Int] = []
         for idx in 0..<Int.random(in: 0..<5) {
@@ -416,7 +416,7 @@ public let CodeGenerators: [CodeGenerator] = [
                 candidates.append(b.randVar())
             }
         }
-        b.destructAndReassign(candidates, atIndices: indices, from: arr, hasRestElement: probability(0.2))
+        b.destruct(arr, selecting: indices, into: candidates, hasRestElement: probability(0.2))
     },
 
     CodeGenerator("ComparisonGenerator", inputs: (.anything, .anything)) { b, lhs, rhs in
--- a/Sources/Fuzzilli/Core/ProgramBuilder.swift
+++ b/Sources/Fuzzilli/Core/ProgramBuilder.swift
@@ -1310,13 +1310,13 @@ public class ProgramBuilder {
     }
 
     @discardableResult
-    public func destruct(_ input: Variable, selectFirst numOutputs: Int) -> [Variable] {
-        let outputs = perform(DestructArray(numOutputs: numOutputs), withInputs: [input]).outputs
+    public func destruct(_ input: Variable, selectFirst numOutputs: Int, hasRestElement: Bool = false) -> [Variable] {
+        let outputs = perform(DestructArray(numOutputs: numOutputs, hasRestElement: hasRestElement), withInputs: [input]).outputs
         return Array(outputs)
     }
 
-    public func destructAndReassign(_ candidates: [Variable], to input: Variable) {
-        perform(DestructArrayAndReassign(numInputs: candidates.count), withInputs: [input] + candidates)
+    public func destructAndReassign(_ candidates: [Variable], to input: Variable, hasRestElement: Bool = false) {
+        perform(DestructArrayAndReassign(numInputs: candidates.count, hasRestElement: hasRestElement), withInputs: [input] + candidates)
     }
 
     @discardableResult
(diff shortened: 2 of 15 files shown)


Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -371,6 +371,31 @@ public let CodeGenerators: [CodeGenerator] = [
         b.reassign(target, to: val)
     },
 
+    CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
+        // TODO: We need a way to expose the length of arr so that element selection is accurate
```
saelo, line 375: This'd be quite complex to do, and I don't think we really need it: 
```javascript
let arr = [1,2,3];
let [x, y] = arr;
// x = 1, y = 2
let [v1, v2, v3, v4] = arr;
// v4 = undefined
```
seems to work fine.

Maybe just update this comment to say something like "we could improve this logic by tracking the lengths of arrays and using that, but it's not strictly necessary since array destructuring allows the number of variables to be different from the length of the array"

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -622,6 +622,31 @@ class Reassign: Operation {
     }
 }
 
+/// Destructs an array into n output variables
+class DestructArray: Operation {
+    let indices: [Int]
+    let hasRestElement: Bool
+    
+    init(indices: [Int], hasRestElement: Bool) {
+        self.hasRestElement = hasRestElement
```
saelo, line 631: `indices` must be sorted here, correct? Maybe add an `assert` to verify that? I guess it must also not have duplicates, although that might be a bit harder to assert (maybe convert to a `Set`, then compare against that?), so if it's more than one short-ish line, it's probably fine to just add a comment for that property

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -115,6 +115,44 @@ public class JavaScriptLifter: Lifter {
...
+                        indices = op.indices
+                        hasRestElement = op.hasRestElement
+                    case let op as DestructArrayAndReassign:
+                        outputs = instr.inputs.dropFirst().map({ $0.identifier })
+                        indices = op.indices
+                        hasRestElement = op.hasRestElement
+                    default:
+                        return arrayPattern
+                }
+                assert(indices.count == outputs.count)
```
saelo, line 137: Can we copy this assertion into the operation constructor as well?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -115,6 +115,44 @@ public class JavaScriptLifter: Lifter {
...
+                                hasRestElement ? arrayPattern.append(" ...\(outputs[outputIndex])") : arrayPattern.append(" \(outputs[outputIndex])")
+                            } else {
+                                arrayPattern.append("\(outputs[Int(outputIndex)]),")
+                            }
+                            outputIndex += 1
+                        } else {
+                            arrayPattern.append(" ,")
+                        }
+                    }
+                }
```
saelo, line 151: Can we simplify this a bit and make it O(N) (not O(N^2), not that it really matters though). Something like
```swift
var lastIndex = 0
for (index, output) in zip(indices, outputs) {
    let skipped = index - lastIndex
    lastIndex = index
    let dots = index == indices.last! && hasRestElement ? "..." : ""
    arrayPattern += String(repeating: ",", count: skipped) + dots + output + ","
}
```
maybe?

Sources/Fuzzilli/Protobuf/program.pb.swift
```diff
@@ -132,777 +132,799 @@ public struct Fuzzilli_Protobuf_Instruction {
   /// The operation is either encoded as an index, referring to the nth operation
   /// (so that shared operations are also only present once in the protobuf), or
   /// as one of the many concrete Operation messages.
-  public var inouts: [UInt32] = []
+  public var inouts: [UInt32] {
```
saelo, line 135: Are all of these changes due to a newer version of the Swift protobuf compiler? I'd slightly prefer to do that in a separate PR if possible, but it's fine if that's too much hassle (e.g. getting an older version to work)

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -622,6 +622,31 @@ class Reassign: Operation {
     }
 }
 
+/// Destructs an array into n output variables
+class DestructArray: Operation {
+    let indices: [Int]
```
saelo, line 627: Do we want the OperationMutator to be able to modify these (i.e. turn one index into another)? I think that could make sense, although it'll need to ensure the indices are still sorted and contain no duplicates

Sources/Fuzzilli/Lifting/FuzzILLifter.swift
```diff
@@ -220,6 +220,50 @@ public class FuzzILLifter: Lifter {
...
+                    } else {
+                        arrayPattern.append(",")
+                    }
+                }
+            }
+            w.emit("[\(arrayPattern)] <- DestructArray \(input(0))")
+
+        case let op as DestructArrayAndReassign:
+            var arrayPattern: String = ""
+            if op.indices.count > 0 {
```
saelo, line 247: Maybe similar code changes as in the JSLifter?

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -371,6 +371,31 @@ public let CodeGenerators: [CodeGenerator] = [
...
+        // Pick random indices
+        var indices: [Int] = []
+        for idx in 0..<Int.random(in: 0..<5) {
```
saelo, line 378: Maybe add a short comment about how these random values are determined (5 and 0.7). I think 5 makes sense since it's the default max length of array literals (is it?). Then the probability should probably be selected so that on average, we define 2-3 new variables?

saelo (commented):
Nice! The main thing I'm still wondering about is whether we really need the `AndReassign` variant of this. On the one hand, it most likely gets converted early on in the JS pipeline into a destruct and regular reassignments, so probably doesn't really cover any new code. On the other hand, compiling a destruct + reassign to a normal destruct and a bunch of reassignments is kind of inefficient, and so if that happens frequently enough, it's probably worth keeping the additional instruction for that. WDYT?

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Actually, I can just drop the comment since it's strictly not necessary that we cap the array length.

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: Nice! I learn a new swift construct with each PR I send :D
I'll update this logic.

Sources/Fuzzilli/Protobuf/program.pb.swift
amarekano: Yeah, I've updated my protobuf compiler so there's a whole load of this changes.

amarekano:
> The main thing I'm still wondering about is whether we really need the AndReassign variant of this.
Let me compare the bytecode generate from `destruct + reassign` and `destruct with multiple assignments`. Perhaps we can take a decision using that data.

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: I think you mean adding this in ProgramBuilder right? In the Operation constructor we won't have access to the output variables list.

Sources/Fuzzilli/Protobuf/program.pb.swift
saelo: Hmm are you sure that's the latest release? For me I have
```
> protoc-gen-swift --version
protoc-gen-swift 1.18.0
```
which seems to be the latest version according to https://github.com/apple/swift-protobuf

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
saelo: Ah yeah right, nevermind. The Operation specifies `numOutput` and that is asserted to be equal to the number of outputs in the Instruction, so we don't need a separate check.

Sources/Fuzzilli/Protobuf/program.pb.swift
amarekano: AH! Turns out I picked the last listed tag to compile from which was (1.9.0) and that appears after 1.18.0. I'll recompile and rebase my PRs :(

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: The max length for array generation that we currently have is 10 elements, Should we continue with 5 as the upper limit for array destruct?

saelo: Right ok, yeah maybe still leave it at 5 then, since that's the expected number of elements in an array if the max is 10. WDYT?

amarekano: Yeah that works for me too

amarekano:
I think we should support the `AndReassign` variant for a couple of reasons: the first being that this op generates more interesting bytecode compared to just using simple reassignments. Here's two programs that I've used to test this:

```js
const v4 = [15,30,"Hello"];
let v5 = 1000;
[v5] = v4;

//VS

const v4 = [15,30,"Hello"];
let v5 = 1000;
v5 = v4[0];

```
The bc from the two samples are linked below:
[bc_with_destruct_array.log](https://github.com/googleprojectzero/fuzzilli/files/7573017/bc_with_destruct_array.log)
[bc_without_destruct_array.log](https://github.com/googleprojectzero/fuzzilli/files/7573024/bc_without_destruct_array.log)

The second reason is that handling `restElement` assignments would become tricky. Thinking off the top of my head, I imagine you'd need to introduce a loop (or some slice operation) that would simulate the same reassignment, which would change the semantics of the program.

amarekano:
Wrt to nested arrays in destruct operations, I think we can handle this during compilation using temporary variables. For example, if we have the following program:
```js
let [x,[a,b]] = [1,[2,3]]
```
we can re-write this during compilation to:

```js
let [x, _temp0] = [1,[2,3]]
let [a,b] = _temp0
```
Diffing the bytecode dumps from the [orginal](https://github.com/googleprojectzero/fuzzilli/files/7576412/orginal_bc.log) and [modified sample](https://github.com/googleprojectzero/fuzzilli/files/7576413/rewrite_bc.log) above doesn't reveal a significant difference so I think we should be fine on that front as well.

All other review comments have been addressed so we should be good to go here. Let me know if you'd like me to incorporate any other changes or improvements.

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -117,6 +117,22 @@ public class OperationMutator: BaseInstructionMutator {
             newOp = BinaryOperation(chooseUniform(from: allBinaryOperators))
         case is ReassignWithBinop:
             newOp = ReassignWithBinop(chooseUniform(from: allBinaryOperators))
+        case let op as DestructArray:
+            var newIndices: [Int] = []
```
saelo, line 121: I think I'd prefer if these did smaller changes. Also I'm not sure that this currently works? The number of outputs might now be different than the number of indices. How about something like:
```swift
var newIndices = Set(op.indices)
if newIndices.count > 0 {
    newIndices.remove(newIndices.randomElement()!)
    while newIndices.count != op.indices.count {
        newIndices.insert(Int.random(in: 0..<10))
    }
}
newOp = ... (newIndices.sorted(), ...)
...
```

saelo: I think it's ok to use a larger maximum number (10) here than in the generator

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -393,6 +393,32 @@ public let CodeGenerators: [CodeGenerator] = [
         b.reassign(target, to: val)
     },
 
+    CodeGenerator("DestructArrayGenerator", input: .iterable) { b, arr in
+        // Fuzzilli generated arrays can have a length ranging from 0 to 10 elements,
+        // We want to ensure that 1) when destructing arrays we are within this length range
```
saelo, line 398: Nit: maybe "when destructing arrays we are usually within", because we might still generate arrays with only 1 or 2 elements

saelo (commented):
Cool thanks, that makes sense, so let's keep both variants!

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1309,6 +1309,16 @@ public class ProgramBuilder {
         perform(Reassign(), withInputs: [output, input])
     }
 
+    @discardableResult
+    public func destruct(_ input: Variable, selecting indices: [Int], hasRestElement: Bool = false) -> [Variable] {
+        let outputs = perform(DestructArray(indices: indices, hasRestElement: hasRestElement), withInputs: [input]).outputs
+        return Array(outputs)
+    }
+
+    public func destructAndReassign(_ candidates: [Variable], atIndices indices: [Int], from input: Variable, hasRestElement: Bool = false) {
```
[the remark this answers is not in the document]
amarekano: thoughts on renaming this function to just `destruct` and updating the signature as you've suggested?

Sources/Fuzzilli/Protobuf/operations.pb.swift
```diff
@@ -950,6 +950,34 @@ public struct Fuzzilli_Protobuf_Reassign {
...
 }
 
+public struct Fuzzilli_Protobuf_DestructArray {
+  // SwiftProtobuf.Message conformance is added in an extension below. See the
+  // `Message` and `Message+*Additions` files in the SwiftProtobuf library for
+  // methods supported on all messages.
+
+  public var indices: [Int32] = []
+
+  public var hasRestElement_p: Bool = false
```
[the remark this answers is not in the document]
amarekano: I think this generated by the swift proto complier (which I've now correctly upgraded to 1.18.0, thanks for pointing that out :)) and it appears to be a standard representation for `bool` fields. We've also got this notation in BeginClassDef:
```swift
case .beginClassDefinition(let p):
            op = BeginClassDefinition(hasSuperclass: p.hasSuperclass_p,
```

Sources/FuzzilliCli/CodeGeneratorWeights.swift
```diff
@@ -70,6 +70,8 @@ let codeGeneratorWeights = [
     "ReassignWithBinopGenerator":               25,
     "DupGenerator":                             2,
     "ReassignmentGenerator":                    25,
+    "DestructArrayGenerator":                   2,
+    "DestructArrayAndReassignGenerator":        7,
```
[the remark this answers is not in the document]
amarekano: Not really, these are just arbitrary values at the moment. Happy to change them to something else if you've got suggestions.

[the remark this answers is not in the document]
amarekano: I've updated these to now have the same value of 7 which is the average of weights from `ArrayGenerator` and `ArrayWithSpreadGenerator`. Again, this is purely empirical :)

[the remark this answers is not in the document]
saelo: Sounds good, thanks!

Sources/Fuzzilli/Core/ProgramBuilder.swift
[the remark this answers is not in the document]
saelo: sure, that works as well!

PR #301, opened 2021-11-22
Function parameters now support default assignments

amarekano:
This implementation has some limitations:
1. We currently can't use `super` and `arguments` as default args since these ops have context restrictions and we won't always be able to emit them into the outer scope. One way to solve this could be to use `LoadFromScope` operations.
2. Now that we have default args, should we ensure that `inputTypes` is in sync with the default arg types? Not sure if that's a good idea since the default args are generated randomly and narrowing the type from `.anything` doesn't make a lot of sense. Happy to hear arguments (no pun intended!) both for an against this.


amarekano:
@saelo this PR is ready for review. Please see the note on limitations and let me know what you think.

saelo (commented):
Cool, thanks for working on this!

So from what I understand, the semantics of default parameter values in JS are essentially like this:
```javascript
function foo(x = something) {
    ...
}
```
is equivalent to
```javascript
function foo(x) {
    if (typeof(x) == "undefined") {
        x = something;
    }
    ....
}
```

One interesting implication is that default-value expressions are re-evaluated at each invocation:
```javascript
function foo(x = console.log("Hello!")) { return x; }
foo()
// Hello!
foo()
// Hello!
```

With that, I broadly see two options for implementing default values in Fuzzilli:

1. Essentially the way you have it now: default values are just inputs to the function. The nice thing here is that we can easily lift those default values to "real" JS default values, basically like you are doing now. The drawback is that we don't capture the semantics of default values properly: in FuzzIL, it'll look like default values are never re-evaluated because they are just a variable that already exists, while in the generated JS code, it'll depend on expression inlining whether values are re-evaluated. Example:

```
v0 <- CreateObject ...
BeginFunctionDefinition <- v0 -> v1
    ...
EndFunctionDefinition
```
could become either
```js
const v0 = { ... };
function foo(v1 = v0) { ... }
```
or 
```js
functino foo(v1 = {...}) { ... }
```
which are pretty different (in the 2nd example, every invocation will create a fresh object). However, the inlining decisions also affect semantics elsewhere, so this wouldn't be the only place...

2. We lower default values to these `if (param === undefined) { param = defaultValue; }` constructs, either through a dedicated operation, or just with an if-else. This properly captures the semantics of default values, but requires quite a bit more FuzzIL operations, which causes other problems, so that's also not great.

Maybe there are other options yet?

Now if we decide for approach 1 (I think I'm leaning towards that), and we anyway refactor `FunctionSignature`, then I think I'd be in favor of moving the information about which parameters have default values into the FunctionSignature (so we'd have a Parameter type `.paramWithDefault` or so). This will again enable us to support default values in every type of function without any further changes.

saelo:
> This implementation has some limitations:
> 
> 1. We currently can't use `super` and `arguments` as default args since these ops have context restrictions and we won't always be able to emit them into the outer scope. One way to solve this could be to use `LoadFromScope` operations.
> 2. Now that we have default args, should we ensure that `inputTypes` is in sync with the default arg types? Not sure if that's a good idea since the default args are generated randomly and narrowing the type from `.anything` doesn't make a lot of sense. Happy to hear arguments (no pun intended!) both for an against this.

So for 1. I think it's fine to ignore it. I'm also not sure if this won't fix itself: of `super` and `arguments` have special context requirements (in class and in function), and we decide for approach 1, then they shouldn't be available before the function. Or is the problem that they'll be inlined, and then that causes syntax errors? In either case, I think it's fine to ignore that...

For 2. I think it would make sense to try and find matching (i.e. correct type) default values, at least in .conservative mode. Maybe this'll be fairly easy though if we have all the information in the `FunctionSignature` (i.e. the information about the expected type and whether there's a default value)

amarekano:
> Maybe there are other options yet?

I don't think there's an easy way to handle call-time-invocation without using some form of inlining. There's a potentially simpler approach using inlining that we could consider. How about we add a property to `Variable` such as `isInlined` to help the lifter determine which outputs should be cached and which outputs should be emitted immediately to the script buffer. For example consider the following program:

```js
let v0 = CreateObject...
BeginFunctionDefinition <- v1 -> v0
....
EndFunctionDefinition
```

At the time of code gen we could set the `isInlined` flag on the output of the CreateObject operation i.e `v0`. At the time of lifting, if we determine that the output from the operation has the `isInlined` property set then we generate an expression for `v0` and store this in a 'expression cache' rather than emitting it to the script buffer.

Now when we get to parsing default function parameters, we first look up the inline cache for an expression that matches the default assignment variable and if that lookup fails then we simply lift the variable identifier. In this way we move the inlining property from the lifter to the output variables. The FuzzIL would essentially stay the same and if we wanted to we can add additional analysers to track these inline variables if we wanted to (I currently can't think of a reason why that would be useful.) Wdyt?

> Now if we decide for approach 1 (I think I'm leaning towards that), and we anyway refactor FunctionSignature, then I think I'd be in favor of moving the information about which parameters have default values into the FunctionSignature (so we'd have a Parameter type .paramWithDefault or so). This will again enable us to support default values in every type of function without any further changes.

This makes sense but we would still need to update the various function operations (regular, objects, classes) to have the numInputs match the number of default parameter assignments. I don't think that's a big deal but something to consider.

saelo:
Now that we have improved `FunctionSignatures`, do we want to use them here to express whether there are default arguments? I could imagine that to work pretty nicely, since default arguments should be supported in all the different function types we have

saelo:

PR #302, opened 2021-11-26
CallSuperConstructor now supports spread arguments

amarekano:


Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1410,8 +1410,14 @@ public class ProgramBuilder {
         return classDefinition.output
     }
 
-    public func callSuperConstructor(withArgs arguments: [Variable]) {
-        perform(CallSuperConstructor(numArguments: arguments.count), withInputs: arguments)
+    @discardableResult
+    public func callSuperConstructor(withArgs arguments: [Variable]) -> Variable {
```
saelo, line 1414: Does this really have an output, and what is it?

saelo (commented):
One question, but otherwise looks good!

Sources/Fuzzilli/Core/ProgramBuilder.swift
amarekano: So this would return an object of the super class but if we don't have a super class then this will generate an invalid program and error out. We probably need a check to ensure that we don't add a `CallSuperConstructor` op when we superClass is null but I'm not sure there's a simple way to do that.

saelo: Doesn't `super()` just return `this`? At least, that's what my quick test seems to indicate. In any case, do we really need that (like, are the JS samples that actually use the return value of `super()`)? It seems like it might be easier to just leave this operation with zero outputs since we already have `this` available in a local variable here.

amarekano: So here's a test that I've just tried:
```js
let c = class C extends String {
	constructor(){
		let x = super()
		print(typeof x)
		print(x instanceof String)
		print(x instanceof this)
	}
}

let y = new c()
```
which generates the following output:
```js
object
true
Exception: TypeError: this is not a function. (evaluating 'x instanceof this')
C@test.js:6:22
global code@test.js:10:14
```
How did you check that it returns this? Perhaps I'm doing something incorrectly.

saelo: can you insert `print(x === this)` into the constructor? I think `print(x instanceof C)` might also work, although it'd probably need to be a `class` statement, not expression.

amarekano: > It seems like it might be easier to just leave this operation with zero outputs
Yeah I agree, I don't see samples using the output so we can drop that.

amarekano: > can you insert print(x === this) into the constructor? I think print(x instanceof C) might also work, although it'd probably need to be a class statement, not expression.

these two statements evaluate to true

amarekano: I've dropped the output from this op, should be good to go.

saelo:
Thanks!

PR #303, opened 2021-11-28
Add array destruct support to ForOf

amarekano:
This PR allows us to generate for-of loops of the form:
```js
let arr = [[key,value], [key,value]]
for( let [x,y] of arr ){
// stuff
}
```

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -588,14 +588,12 @@ public let CodeGenerators: [CodeGenerator] = [
             }
         }
 
-        if indices.count == 0 {
-            b.forOfLoop(obj) { _ in
-                b.generateRecursive()
-            }
-        } else {
-            b.forOfLoop(obj, selecting: indices, hasRestElement: probability(0.2)) { _ in
-                b.generateRecursive()
-            }
+        if indices.isEmpty {
+            indices = [0]
+        }
+        
+        b.forOfLoop(obj, selecting: indices, hasRestElement: probability(0.2)) { _ in
+            b.generateRecursive()
         }
     },
 
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -573,6 +573,14 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("ForOfLoopGenerator", input: .iterable) { b, obj in
+        b.forOfLoop(obj) { _ in
+            b.generateRecursive()
+        }
+    },
+
+    CodeGenerator("ForOfWithDestructLoopGenerator", input: .iterable) { b, obj in
+        // Don't run this generator in conservative mode, until we can track array element types
+        guard b.mode != .conservative else { return }
         var indices: [Int] = []
         for idx in 0..<Int.random(in: 1..<5) {
             withProbability(0.8) {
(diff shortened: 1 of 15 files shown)


Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -573,8 +573,21 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("ForOfLoopGenerator", input: .iterable) { b, obj in
-        b.forOfLoop(obj) { _ in
-            b.generateRecursive()
+        var indices: [Int] = []
```
saelo, line 576: I suspect that the ForOfLoop with destructuring has a pretty high failure rate, since stuff like
```javascript
var a = [1,2,3]
for ([x, y] of a) { console.log(x, y); }
// Uncaught:
// TypeError: number 1 is not iterable (cannot read property Symbol(Symbol.iterator))
```
throws an exception.

So I think there should at least be two separate CodeGenerators, and the one that does destructuring probably shouldn't run in `.conservative` mode at all (maybe with a comment "until we can track array element types").

I'm not sure if having two separate FuzzIL Operations would make sense as well. Maybe it does: if the OperationMutator is able to mutate the `indices` of a `ForOfLoop` (it probably should?), then for a "regular" for-of loop (one with `indices = [0]`), it would most of the time construct something that's invalid (i.e. `for ([,,,x] of y)`, where `y` is a normal array, not a nested one). Further, it probably would make lifting a tiny bit easier (get's rid of `if op.indices.count > 1 {` below). So I think I'm in favour of adding a separate `ForOfLoopWithDestruct` operation. WDYT?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -614,8 +614,13 @@ public class JavaScriptLifter: Lifter {
                 w.decreaseIndentionLevel()
                 w.emit("}")
 
-            case is BeginForOf:
-                w.emit("for (\(decl(instr.innerOutput)) of \(input(0))) {")
+            case let op as BeginForOf:
+                if op.indices.count > 1 {
```
saelo, line 618: Is this correct? What if that index is != 0?

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Yeah this makes sense and I agree, I'll add a new op,

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: This was incorrect and should now be fixed with the latest commit (i.e. two unique ops for for-of loops).
I've also included this as part of the lifting test case.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -578,6 +578,27 @@ public let CodeGenerators: [CodeGenerator] = [
...
+        // Don't run this generator in conservative mode, until we can track array element types
+        guard b.mode != .conservative else { return }
+        var indices: [Int] = []
+        for idx in 0..<Int.random(in: 1..<5) {
+            withProbability(0.8) {
+                indices.append(idx)
+            }
+        }
+
+        if indices.count == 0 {
```
saelo, line 591: Nit: can we instead just ensure that `indices.count > 0` and drop this case? How about just adding
```swift
if indices.isEmpty {
    indices = [0]
}
```

which has the same effect, but still generates a `ForOpWithDestruct` in this generator

amarekano: sounds good, I've added the the changes

PR #309, opened 2021-12-28
Add DestructObject and DestructObjectAndReassign operations

amarekano:
This PR adds support for [object destructing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring)

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -421,21 +421,15 @@ public let CodeGenerators: [CodeGenerator] = [
 
     CodeGenerator("DestructObjectGenerator", input: .object()) { b, obj in
         var properties = Set<String>()
-        for _ in 0...5 {
-            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
+        for _ in 0..<Int.random(in: 2...6) {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop) {
                 properties.insert(prop)
+            } else {
+                properties.insert(b.genPropertyNameForRead())
             }
         }
 
-        // If we haven't added any properties, add a random property name
-        if properties.isEmpty {
-            properties.insert(b.genPropertyNameForRead())
-        }
-
         let hasRestElement = probability(0.2)
-
-        // Ensure that we have at least one output
-        guard !properties.isEmpty || hasRestElement else { return }
 
         b.destruct(obj, selecting: properties.sorted(), hasRestElement: hasRestElement)
     },
@@ -443,24 +437,17 @@ public let CodeGenerators: [CodeGenerator] = [
     CodeGenerator("DestructObjectAndReassignGenerator", input: .object()) { b, obj in
         var candidates: [Variable] = []
         var properties = Set<String>()
-        for _ in 0...5 {
-            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
+        for _ in 0..<Int.random(in: 2...6) {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop) {
                 properties.insert(prop)
+                candidates.append(b.randVar())
+            } else {
+                properties.insert(b.genPropertyNameForRead())
                 candidates.append(b.randVar())
             }
         }
 
-        // If we haven't added any properties, add a random property name
-        if properties.isEmpty {
-            properties.insert(b.genPropertyNameForRead())
-            candidates.append(b.randVar())
-        }
-
         let hasRestElement = probability(0.2)
-
-        // Ensure that we have at least one input variable to reassign
-        guard !properties.isEmpty || hasRestElement else { return }
-
         if hasRestElement {
             candidates.append(b.randVar())
         }
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -419,6 +419,40 @@ public let CodeGenerators: [CodeGenerator] = [
         b.destruct(arr, selecting: indices, into: candidates, hasRestElement: probability(0.2))
     },
 
+    CodeGenerator("DestructObjectGenerator", input: .object()) { b, obj in
+        var properties = Set<String>()
+        for _ in 0..<Int.random(in: 0...b.type(of: obj).numProperties) {
+            properties.insert(b.type(of: obj).randomProperty()!)
+        }
+
+        let hasRestElement = probability(0.2)
+
+        // Ensure that we have at least one output
+        guard !properties.isEmpty || hasRestElement else { return }
+
+        b.destruct(obj, selecting: properties.sorted(), hasRestElement: hasRestElement)
+    },
+
+    CodeGenerator("DestructObjectAndReassignGenerator", input: .object()) { b, obj in
+        var candidates: [Variable] = []
+        var properties = Set<String>()
+        for _ in 0..<Int.random(in: 0...b.type(of: obj).numProperties) {
+            properties.insert(b.type(of: obj).randomProperty()!)
+            candidates.append(b.randVar())
+        }
+
+        let hasRestElement = probability(0.2)
+
+        // Ensure that we have at least one input variable to reassign
+        guard !properties.isEmpty || hasRestElement else { return }
+
+        if hasRestElement {
+            candidates.append(b.randVar())
+        }
+
+        b.destruct(obj, selecting: properties.sorted(), into: candidates, hasRestElement: hasRestElement)
+    },
+
     CodeGenerator("ComparisonGenerator", inputs: (.anything, .anything)) { b, lhs, rhs in
         b.compare(lhs, rhs, with: chooseUniform(from: allComparators))
     },
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -435,16 +435,17 @@ public let CodeGenerators: [CodeGenerator] = [
     },
 
     CodeGenerator("DestructObjectAndReassignGenerator", input: .object()) { b, obj in
-        var candidates: [Variable] = []
         var properties = Set<String>()
         for _ in 0..<Int.random(in: 2...6) {
             if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop) {
                 properties.insert(prop)
-                candidates.append(b.randVar())
             } else {
                 properties.insert(b.genPropertyNameForRead())
-                candidates.append(b.randVar())
             }
+        }
+
+        var candidates = properties.map{ _ in
+            b.randVar()
         }
 
         let hasRestElement = probability(0.2)
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -421,8 +421,15 @@ public let CodeGenerators: [CodeGenerator] = [
 
     CodeGenerator("DestructObjectGenerator", input: .object()) { b, obj in
         var properties = Set<String>()
-        for _ in 0..<Int.random(in: 0...b.type(of: obj).numProperties) {
-            properties.insert(b.type(of: obj).randomProperty()!)
+        for _ in 0...5 {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
+                properties.insert(prop)
+            }
+        }
+
+        // If we haven't added any properties, add a random property name
+        if properties.isEmpty {
+            properties.insert(b.genPropertyNameForRead())
         }
 
         let hasRestElement = probability(0.2)
@@ -436,8 +443,16 @@ public let CodeGenerators: [CodeGenerator] = [
     CodeGenerator("DestructObjectAndReassignGenerator", input: .object()) { b, obj in
         var candidates: [Variable] = []
         var properties = Set<String>()
-        for _ in 0..<Int.random(in: 0...b.type(of: obj).numProperties) {
-            properties.insert(b.type(of: obj).randomProperty()!)
+        for _ in 0...5 {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
+                properties.insert(prop)
+                candidates.append(b.randVar())
+            }
+        }
+
+        // If we haven't added any properties, add a random property name
+        if properties.isEmpty {
+            properties.insert(b.genPropertyNameForRead())
             candidates.append(b.randVar())
         }
 
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -131,8 +131,7 @@ public class JavaScriptLifter: Lifter {
                 return arrayPattern
             }
 
-            // Helper function to lift destruct object operations
-            func liftObjectPattern(properties: [String], outputs: [String], hasRestElement: Bool) -> String {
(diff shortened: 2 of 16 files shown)


amarekano:
```
Run swift test -v
C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swiftc.exe -print-target-info
TSCUtility/Triple.swift:154: Fatal error: could not determine host triple: malformed
error: unable to invoke subcommand: C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift-test.exe ()
Error: Process completed with exit code 1.
```
Not quite sure why the Win build is failing

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -135,6 +135,18 @@ public class OperationMutator: BaseInstructionMutator {
                 }
             }
             newOp = DestructArrayAndReassign(indices: newIndices.sorted(), hasRestElement: !op.hasRestElement)
+        case let op as DestructObject:
+            var newProperties = Set(op.properties)
+            if newProperties.count > 0 {
+                newProperties.remove(newProperties.randomElement()!)
+            }
```
saelo, line 142: Shouldn't this also add a new property to the set (here and below)?

Also now that I see it, in the two cases above (`DestructArray`, `DestructArrayAndReassign`), let's try to get rid of the `while` loop: in case of an imported program (compiled from JS, for example), we could end up with a `DestructArray` that has indices 0 through 20 or so, and then this would turn into an infinite loop :(

How about we make the logic for these four cases something like
- remove a random element from the set
- attempt X times (let's say 10) to insert a random new element, then just give up

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -131,6 +131,21 @@ public class JavaScriptLifter: Lifter {
                 return arrayPattern
             }
 
+            // Helper function to lift destruct object operations
+            func liftObjectPattern(properties: [String], outputs: [String], hasRestElement: Bool) -> String {
```
saelo, line 135: Nit: maybe rename here and in FuzzILLifter to `liftObjectDestructPattern`? Then you can also drop the comment. If you want, you can also pass in the destructured object variable.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -419,6 +419,40 @@ public let CodeGenerators: [CodeGenerator] = [
         b.destruct(arr, selecting: indices, into: candidates, hasRestElement: probability(0.2))
     },
 
+    CodeGenerator("DestructObjectGenerator", input: .object()) { b, obj in
+        var properties = Set<String>()
+        for _ in 0..<Int.random(in: 0...b.type(of: obj).numProperties) {
```
saelo, line 424: I'm not sure this works well in practice. I think we might have some objects with a lot's of properties (e.g. Functions or the `Number` constructor) and then we'll probably have lot's of objects without properties (e.g. because the AI can't properly track them).
How about we just again make a loop with 1-5 iterations, then include a `randomProperty()` if available (and if we don't already have it), otherwise just a random property name?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
amarekano: I'd like to leave the function signature as is, the reason being that I want to reuse this when I create ops that lift a function param list as destructible object. Ex:
```js
function f({"prop1":v1, "prop2":v2}) {
  return v1 + v2;
}
```

amarekano:
review comments addressed with the latest commit. I've tweaked the property selection logic, that you've recommended, a tiny bit. Let me know if you'd like that changed.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -419,6 +419,55 @@ public let CodeGenerators: [CodeGenerator] = [
         b.destruct(arr, selecting: indices, into: candidates, hasRestElement: probability(0.2))
     },
 
+    CodeGenerator("DestructObjectGenerator", input: .object()) { b, obj in
+        var properties = Set<String>()
+        for _ in 0...5 {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
```
saelo, line 425: Nit: I don't think you need the `!properties.contains(prop)` condition, `properties.insert(prop)` will already need to do that check, and be a nop if the element already exists

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -419,6 +419,55 @@ public let CodeGenerators: [CodeGenerator] = [
...
+
+        // If we haven't added any properties, add a random property name
+        if properties.isEmpty {
+            properties.insert(b.genPropertyNameForRead())
+        }
+
+        let hasRestElement = probability(0.2)
+
+        // Ensure that we have at least one output
+        guard !properties.isEmpty || hasRestElement else { return }
```
saelo, line 438: Is this check still needed?

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -419,6 +419,55 @@ public let CodeGenerators: [CodeGenerator] = [
...
+
+        // If we haven't added any properties, add a random property name
+        if properties.isEmpty {
+            properties.insert(b.genPropertyNameForRead())
+            candidates.append(b.randVar())
+        }
+
+        let hasRestElement = probability(0.2)
+
+        // Ensure that we have at least one input variable to reassign
```
saelo, line 461: Same here

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -419,6 +419,55 @@ public let CodeGenerators: [CodeGenerator] = [
...
+        guard !properties.isEmpty || hasRestElement else { return }
+
+        b.destruct(obj, selecting: properties.sorted(), hasRestElement: hasRestElement)
+    },
+
+    CodeGenerator("DestructObjectAndReassignGenerator", input: .object()) { b, obj in
+        var candidates: [Variable] = []
+        var properties = Set<String>()
+        for _ in 0...5 {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
```
saelo, line 447: Same here

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -119,22 +119,72 @@ public class OperationMutator: BaseInstructionMutator {
             newOp = ReassignWithBinop(chooseUniform(from: allBinaryOperators))
         case let op as DestructArray:
             var newIndices = Set(op.indices)
-            if newIndices.count > 0 {
-                newIndices.remove(newIndices.randomElement()!)
-                while newIndices.count != op.indices.count {
-                    newIndices.insert(Int.random(in: 0..<10))
+            var removedIndex = -1
```
saelo, line 122: Ok cool, this logic makes sense to be, but I'm wondering if we can factor it out instead of essentially duplicating it four times.

How about something like this (with some slight simplifications as well):
```swift
        func replaceRandomElement<T>(in set: inout Set<T>, generatingRandomValuesWith generator: () -> T) {
            guard let removedElem = set.randomElement() else { return }
            set.remove(removedElem)

            for _ in 0...5 {
                let newElem = generator()
                // Ensure that we neither add an element that already exists nor add one that we just removed
                if !set.contains(newElem) && newElem != removedElem {
                    set.insert(newElem)
                    break
                }
            }
        }
```

then call it like
```swift
var newIndices = Set(op.indices)
replaceRandomElement(in: &newIndices, generatingRandomValuesWith: { return Int.random(in 0..<10) })
```

Do you think that'd work?

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -119,22 +119,72 @@ public class OperationMutator: BaseInstructionMutator {
...
+            var removedProperty = ""
+            if newProperties.count > 0, let prop = newProperties.randomElement() {
+                removedProperty = prop
+                newProperties.remove(prop)
+            }
+            // Append a random property
+            let properties = b.type(of: instr.input(0)).properties
+            for _ in 0...5 {
+                // Ensure that we don't add the property that we just removed
+                if properties.count  > 0, let prop = properties.randomElement(), prop != removedProperty {
```
saelo, line 164: I kind of think we should just choose properties here with `b.genPropertyNameForRead()`. These basic mutators are currently mostly ignoring type information (except for the type-aware InputMutator)

Sources/Fuzzilli/Core/CodeGenerators.swift
saelo: Alternatively, you could perform `properties.insert(b.genPropertyNameForRead())` when the property already exists or if none was found? Maybe that'd be a better option.

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -419,6 +419,55 @@ public let CodeGenerators: [CodeGenerator] = [
         b.destruct(arr, selecting: indices, into: candidates, hasRestElement: probability(0.2))
     },
 
+    CodeGenerator("DestructObjectGenerator", input: .object()) { b, obj in
+        var properties = Set<String>()
+        for _ in 0...5 {
```
saelo, line 424: Do we really always want to destruct 6 properties, not a random number between, say, 2 and 6?

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: yes, you're right it is a Set :)

I'm a little hesitant to add random properties using genPropertyName*, mainly because I feel it might cause a high failure rate if the property does not exist, especially when running this generator on imported compiled programs. I think we might need a better way to track properties across objects/classes so that assignments are more accurate, wdyt?

saelo: So I think it's fine (but maybe only if `!conservative_mode`). Due to mutations, e.g. InputMutator, we'll anyway quickly (?) end up with different objects being destructured or "wrong" property names being used. But that's "fine", it'll just result in a variable with the `undefined` value afaik. On the other hand, I'm a bit worried about only limiting us to what the AI knows about object properties. Improvements to the AI will still automatically improve this logic later on (because we'll only use `b.genPropertyNameForRead()` as fallback).

amarekano: Fair, that makes sense to me

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: yeah could do, I'll update the range

amarekano: should we also do this for the op mutator as well?

saelo: I think it's fine for the op mutator to just always change one property. Or do you mean something different?

amarekano: I meant the fixed loop that tries to find a new element:
```swift
for _ in 0...5 {
     let newElem = generator()
     ...
}
```
But now that I think about it, it shouldn't matter since we only need the selection to succeed once :)

Sources/Fuzzilli/Mutators/OperationMutator.swift
amarekano: Changes added, I've replaced the guard statement with and if-else block. If we can't remove an element, say if the set is empty, then instead of bailing we should attempt to add a new property.

Sources/Fuzzilli/Mutators/OperationMutator.swift
amarekano: I've added this here:
```swift
replaceRandomElement(in: &newProperties, generatingRandomValuesWith: { return b.type(of: instr.input(0)).properties.randomElement() ?? b.genPropertyNameForRead() })
newOp = DestructObject(properties: newProperties.sorted(), hasRestElement: !op.hasRestElement)
```

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -421,46 +421,33 @@ public let CodeGenerators: [CodeGenerator] = [
...
         var properties = Set<String>()
-        for _ in 0...5 {
-            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop){
+        for _ in 0..<Int.random(in: 2...6) {
+            if let prop = b.type(of: obj).properties.randomElement(), !properties.contains(prop) {
                 properties.insert(prop)
                 candidates.append(b.randVar())
+            } else {
+                properties.insert(b.genPropertyNameForRead())
+                candidates.append(b.randVar())
```
saelo, line 446: I think here it could happen that the `insert` doesn't actually insert anything (because the property is already in there), in which case you'd end up with more `candidates` than `properties`.
Maybe it's easier to just compute the candidates separately, after the properties. E.g. with something like

```swift
var variables = properties.map { b.randVar() }
```
or just a regular for-loop

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -26,6 +26,22 @@ public class OperationMutator: BaseInstructionMutator {
         var newOp: Operation
         
         b.trace("Mutating next operation")
+
+        func replaceRandomElement<T>(in set: inout Set<T>, generatingRandomValuesWith generator: () -> T) {
+            let removedElem: T? = nil
+            if let removedElem = set.randomElement() {
```
saelo, line 32: Is this really fine? (instead of the `let removedElem = set.randomElement() else { return }`)? When you change the size of the set, don't you then also need to change the inputs of the instruction? I think it's easier to just bail and not mutate an empty set at all (we should anyway never see that in practice)

However, that also means that the loop in lines 36+ must succeed, or else the new set will be smaller... Maybe replace the `break` in line 41 with a `return`, then put

```swift
// Failed to insert a new element, so just insert the removed element again as we must not change the size of the set
set.insert(removedElem)
```
after the loop?

Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -119,22 +135,20 @@ public class OperationMutator: BaseInstructionMutator {
...
-                newIndices.remove(newIndices.randomElement()!)
-                while newIndices.count != op.indices.count {
-                    newIndices.insert(Int.random(in: 0..<10))
-                }
-            }
+            replaceRandomElement(in: &newIndices, generatingRandomValuesWith: { return Int.random(in: 0..<10) })
             newOp = DestructArrayAndReassign(indices: newIndices.sorted(), hasRestElement: !op.hasRestElement)
+        case let op as DestructObject:
+            var newProperties = Set(op.properties)
+            replaceRandomElement(in: &newProperties, generatingRandomValuesWith: { return b.type(of: instr.input(0)).properties.randomElement() ?? b.genPropertyNameForRead() })
```
saelo, line 146: Let's just use `b.genPropertyNameForRead()`. The issue is that if you have an object for which AI knows only one property, then this mutation will just pick that one property 5 times, then give up. So then I think it's better to just choose completely random property names, which we also do elsewhere in this mutator (instead of relying on type information).

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: yeah good catch, I'll add the changes

amarekano:
review comments addressed in the latest commit

PR #310, opened 2022-01-01
Fix incorrect handling of BeginXFunctionDefinition in OperationMutator

amarekano:


Sources/Fuzzilli/Mutators/OperationMutator.swift
```diff
@@ -169,8 +169,18 @@ public class OperationMutator: BaseInstructionMutator {
             } else {
                 newOp = BeginFor(comparator: op.comparator, op: chooseUniform(from: allBinaryOperators))
             }
-        case let op as BeginAnyFunctionDefinition:
-            newOp = BeginAnyFunctionDefinition(signature: op.signature, isStrict: !op.isStrict)
+        case let op as BeginPlainFunctionDefinition:
+                newOp = BeginPlainFunctionDefinition(signature: op.signature, isStrict: !op.isStrict)
```
saelo, line 173: Nit: wrong indention here and the line below

saelo (commented):
Ah, nice catch! So this would've left programs with `BeginAnyFunctionDefinition` operations in them, which then probably crashed the Lifter?

Sources/Fuzzilli/Mutators/OperationMutator.swift
amarekano: fixed

amarekano:
> which then probably crashed the Lifter?
This this is the bit that confused me! I haven't noticed any crashes in the Lifter but I suspect the swift compiler matches BeginAnyFunction with BeginPlainFunction and that probably why we don't see any crashes. Or it may be that the mutation rarely triggered so this crash does not occur as often.

saelo:
Ah, actually I think the problem might be that `BeginAnyFunctionDefinition` isn't marked with `.isParametric` and so will never be mutated by that Mutator? At least in my local tests, I never see it being mutated so far.

PR #313, opened 2022-01-14
Update WebKit revision

timoludwig:
Thanks a lot for your work!

I'm not sure if I'm missing something, but to me it seems like the revision WebKit/WebKit@60a7c2b4b1c92f4ea84b85dcf81e83280d122007 specified in [Targets/JavaScriptCore/REVISION](https://github.com/googleprojectzero/fuzzilli/blob/cbac00cb49d24b455ea727b10061348be5fe8dcf/Targets/JavaScriptCore/REVISION) doesn't exist.
However, the patch also applies cleanly to the latest existing revision WebKit/WebKit@aa0060cae4f6e9c1f3a4825b985dc3ca3e8db3dc.

Thanks in advance for your feedback!


saelo:
Hi! Thanks for this. The issue is that the different WebKit git mirrors seem to have different commit histories, and so different commit hashes... the commit specified in REVISION is valid for this mirror: https://github.com/googleprojectzero/fuzzilli/blob/cbac00cb49d24b455ea727b10061348be5fe8dcf/Cloud/Docker/JSCBuilder/Dockerfile#L12
while the README says to use the Github mirror, where it apparently isn't valid... 

In the past, the Github mirror was frequently not in sync with the main repository, so I switched it at some point, but apparently forgot to update the [README](https://github.com/googleprojectzero/fuzzilli/blob/main/Targets/JavaScriptCore/README.md).

Since https://webkit.org/getting-the-code/ now suggest to use the Github mirror, I'm fine with switching back to that one again, but could you then also please update the Dockerfile above to use the Github mirror as well? Thanks!

timoludwig:
@saelo Thanks for your quick reply and explanation, I have now updated the Dockerfile and added a small note to the README.

saelo:
Thanks!

Issue #299, opened 2021-11-12
`generateVariable` unable to handle type unions


amarekano:
The following test intermittently fails with the following error:

```
Test Case 'ProgramBuilderTests.testTypeInstantiation' started at 2021-11-03 09:00:22.498
Fuzzilli/ProgramBuilder.swift:520: Assertion failed: Unexpected type encountered .integer | .object()
```

What follows is saelo's analysis of this issue:

I think the issue here is that generateType might generate e.g. a TypedArray from the JavaScriptEnvironment (that's correct), then generateVariable tries to call the corresponding constructor and attempts to obtain argument variables (that's also correct), but for TypedArrays, the constructor is defined to take .integer | .object(...):

https://github.com/googleprojectzero/fuzzilli/blob/1c33ac3cefb0771978a3f8f1185ac77fca703b14/Sources/Fuzzilli/Core/JavaScriptEnvironment.swift#L390
 
(also correct), but generateVariable can't yet handle type unions (that is, I think, the issue).

saelo:
This issue appears to happen elsewhere, too, not just during tests. During a fuzzing session with a debug build, I got this assertion failure: `Fuzzilli/ProgramBuilder.swift:520: Assertion failed: Unexpected type encountered .integer | .float | .string`

saelo:
This is now fixed since we no longer use `generateVariable()`: https://github.com/googleprojectzero/fuzzilli/commit/e264abb403c3c63b713d20c8649f0840a59a08cd

PR #314, opened 2022-01-15
Replace Types with Parameters in Function Signatures

amarekano:


saelo (commented):
Nice! This should be much better than using those weird flag types :)

One thing I'm wondering is whether there's some way we could still write function signatures like `[.integer, .opt(.string), .anything...] => .undefined` or something similar (instead of `[.plain(.integer), .opt(.string), rest(.anything)] => .undefined`)). But I couldn't really come up with an easy way of achieving that. The main problem is that the array parameters array would have to contain both `Type` and `Parameter`, which doesn't work with the type inferencing I think. Probably something like this would still work `.integer => .plain(.string) => .anything... => .undefined` (kind of like haskell), but I'm not convinced that'd be better than what we have now... Let me know if you see some way of making this work, but otherwise, I'm fine with the way it is now, and if we later come up with some clever way of doing this, we can still change things around :)

amarekano:
> The main problem is that the array parameters array would have to contain both Type and Parameter, which doesn't work with the type inferencing I think. 

iiuc you mean the array destruct parameters right? If so then I've addressed this in #311. There are two functions `type()` and `inputTypes` that can be called on a `Parameter`. `type()` is called when we want to construct arguments for the function so if we have a function signature with three parameters of the form `.plain(.anything), .destructArray(descriptor), .rest(.anything) => .unknown` which lifts to:
```js
function f(v0, [v1,v2], ...v3) {
}
```

If we wanted to generate arguments for `f` then we query each Parameter in the signature and extract type information using `type()` which would return `.anything`, `.iterable`, and `.anything + .iterable` for the three parameters respectively.

However, when the AI is tracking individual types on innerOutputs (v0 to v3), we call `inputTypes` on each parameter which would return type information for each inner output.

Effectively we have an 'outer view' for each parameter type and an 'inner view' for each individual innerOutput type.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -850,53 +789,172 @@ class TypeExtension: Hashable {
...
+    }
+
+    public init(from proto: ProtobufType) throws {
+        try self.init(from: proto, with: nil)
+    }
+}
+
 public struct FunctionSignature: Hashable, CustomStringConvertible {
-    public let inputTypes: [Type]
+    public let paramTypes: [Parameter]
```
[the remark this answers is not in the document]
amarekano: yeah makes sense, also I think it would be less confusing.

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -252,16 +252,14 @@ public struct AbstractInterpreter {
     private mutating func executeInnerEffects(_ instr: Instruction) {
         // Helper function to process parameters
         func processParameterDeclarations(_ params: ArraySlice<Variable>, signature: FunctionSignature) {
-            for (i, param) in params.enumerated() {
-                let paramType = signature.inputTypes[i]
```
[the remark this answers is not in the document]
amarekano: So we won't be able to use the same logic once we introduce destructArray and destructObject parameters. See latest comment on `type` and `inputTypes` below.

Also I think I it would be a good idea to replace `inputTypes` with `calleeTypes` so that it's less confusing.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -850,53 +789,172 @@ class TypeExtension: Hashable {
...
+               return ".rest(\(t.format(abbreviate: abbreviate)))"
+        }
+    }
+
+    func type() -> Type {
+        switch self {
+            case .plain(let t):
+                return t
+            case .opt(let t):
+                return t
```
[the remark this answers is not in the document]
amarekano: This makes sense for .opt, do we also need to do this for .plain? Also, I've changed .rest to return .iterable.

Yeah i like the idea of renaming the functions to calleeType and callerType. That would definitely make things less confusing.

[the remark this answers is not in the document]
amarekano: > I'm wondering, should this be t | .undefined?

I think we will run into issue when we are unable to merge types. Similar to the issue we have with #299

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -377,31 +377,25 @@ public class ProgramBuilder {
...
             for _ in 0..<Int.random(in: 0...5) {
-                parameterTypes.append(varargsParam.removingFlagTypes())
+                callerTypes.append(restType)
             }
         }
 
-        for var param in parameterTypes {
-            if param.isOptional {
+        for type in callerTypes {
+            if type == .undefined {
```
saelo, line 391: This doesn't quite work, does it? The type would be a union with `.undefined`. But I think it'd anyway be better to call `isOptional` here instead.

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -377,31 +377,25 @@ public class ProgramBuilder {
 
     // This expands and collects types for arguments in function signatures.
     private func prepareArgumentTypes(forSignature signature: FunctionSignature) -> [Type] {
-        var parameterTypes = signature.inputTypes
+        var callerTypes = signature.callerTypes
```
saelo, line 380: I think this function should be rewritten to use `Parameter` directly, something like this:
```swift
var argumentTypes = [Type]()
for param in signature {
    if param.isOptional {
        // It's an optional argument, so stop here in some cases
        ...
    }
    if param.isRest {
        // "Unroll" the rest parameter
        for _ in 0..<Int.random(in: 0...5) {
            argumentTypes.append(param.callerType)
        }
        // Rest parameter must be the last one
        break
    }
    argumentTypes.append(param.callerType)
}
```

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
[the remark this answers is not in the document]
saelo: Ok yeah, that makes sense. We also don't need to union with `.undefined` for the caller (the caller will either pass an argument of `t` or no argument at all). So I guess `callerType` for now is just `return t` in all cases then.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -850,53 +789,168 @@ class TypeExtension: Hashable {
...
+    }
+
+    fileprivate func callerType() -> Type {
+        switch self {
+            case .plain(let t):
+                return t
+            case .opt(let t):
+                return t
+            case .rest(_):
+               return .iterable
```
saelo, line 815: The caller doesn't pass an iterable, it just passes potentially multiple arguments of type `t`. So let's just return `t` here as well

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -850,53 +789,168 @@ class TypeExtension: Hashable {
...
+        }
+    }
+
+    fileprivate func calleeType() -> [Type] {
+        return [self.callerType()]
```
saelo, line 820: I think here optionals should get the union with `.undefined` and rest should get `.iterable`?

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -850,53 +789,168 @@ class TypeExtension: Hashable {
...
+            case .plain(let t):
+                return ".plain(\(t.format(abbreviate: abbreviate)))"
+            case .opt(let t):
+                return ".opt(\(t.format(abbreviate: abbreviate)))"
+            case .rest(let t):
+               return ".rest(\(t.format(abbreviate: abbreviate)))"
+        }
+    }
+
+    fileprivate func callerType() -> Type {
```
saelo, line 808: Nit: I think we could turn these two methods into getters (`var callerType: Type { ...`) to save a few keystrokes when using them, since they are essentially trivial getters.

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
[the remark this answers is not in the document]
saelo: Right, that makes sense!

So IIUC, basically:
- we potentially have more inner outputs (and thus types) than parameters, due to destructuring
- we need to turn rest parameters into whatever the array type of the environment is (a plain JS array in our case)
- we need to turn optional parameters into unions with `.undefined`
- we need to turn plain `.anything` parameters into `.unknown`

So due to the 2nd point, we actually can't properly implement `calleeType` in `Parameter`, since we don't have access to the environment. Right now, we set it to `.iterable` as a workaround, then rely on AI to turn `.iterable` into the actual type here... I think instead, probably we should just implement `calleeType` in the AbstractInterpreter, where we have access to all the data. This is also the only place where it's used from what I can tell.

Then, I think this can just become
```swift
    let types = calleeTypes(for: signature)
    assert(types.count == params.count)
    for (param, type) in zip(params, types) {
        set(param, type)
    }
```

With

```swift
private func calleeTypes(for signature: FunctionSignature) {
    // Basically what is now FunctionSignature.calleeTypes, with the additional steps listed above
}
```

WDYT?

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
saelo: See comment in `AbstractInterpreter`. I think we should just move this method into the AI entirely, where we e.g. have access to the correct ArrayType to use for rest parameters

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -1007,7 +1054,7 @@ extension FunctionSignature: ProtobufConvertible {
     }
 
     init(from proto: ProtobufType, with typeCache: TypeCache?) throws {
-        self.init(expects: try proto.inputTypes.map({ try Type(from: $0, with: typeCache) }),
+        self.init(expects: try proto.parameters.map({ try Parameter(from: $0) }),
```
saelo, line 1057: Do you need to pass the `typeCache` here into the Parameter constructor as well?

saelo:
What I meant above (with `The main problem is that the array parameters array would have to contain both Type and Parameter, which doesn't work with the type inferencing I think.`) is that it'd be awesome if we could still write FunctionSignatures something like `[.integer, .integer...] => .integer` (i.e. with both `Type` and `Parameter` in the array literal) or `[.string, .opt(.string)] => .object` instead of `[.plain(.integer), .rest(integer)] => .integer` and `[.plain(.string), .opt(.string)] => .object` respectively.
But I think this will be pretty hard and/or require somewhat evil hacks to get to work. At least I couldn't really think of a nice way to make it work... :/

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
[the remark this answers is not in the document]
amarekano: Yeah this approach is ideal. I'll add these changes.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
amarekano: > I think here optionals should get the union with .undefined and rest should get .iterable?

I tried this before and I think I ran into a number of assertion failures when merging types. I'll give it another shot and report back.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
amarekano: nice! yeah you're right and this is the solution to the `typeCache` serialisation bug that I was trying to work out.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
amarekano: I'm at 200k iterations and I haven't seen any assertion failures yet. I think this is good to go!

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -219,6 +219,32 @@ public struct AbstractInterpreter {
         setType(of: v, to: t)
     }
 
+    private func calleeTypes(for signature: FunctionSignature) -> [Type] {
+
+        func processType(_ type: Type) -> Type {
+            if type == .anything {
```
saelo, line 225: I think this function should just be
```swift
if type == .anything {
  // .anything in the caller maps to .unknown in the callee
  return .unknown
}
return type
```
And we'll deal with the array type below, which should just be used for `.rest`

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -219,6 +219,32 @@ public struct AbstractInterpreter {
...
+            }
+        }
+
+        var types: [Type] = []
+        signature.parameters.forEach { param in
+            switch param {
+                case .plain(let t):
+                    types.append(processType(t))
+                case .opt(let t):
+                    types.append(processType(t) | .undefined)
```
saelo, line 240: I think this should be `processType(t | .undefined)`, so that `.opt(.anything)` just turns into `.unknown` and not `.unknown | .undefined` (`.unknown` already means that we don't know what it is, so adding in `.undefined` doesn't really make sense and might screw other code that checks for `.unknown`). Maybe make a comment about this behaviour, since it's another tricky bit of logic necessary due to the distinction between `.anything` and `.unknown`...

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
```diff
@@ -219,6 +219,32 @@ public struct AbstractInterpreter {
...
+                case .rest(let t):
+                    types.append(processType(t) | .iterable)
```
saelo, line 242: I think here we always just need `environment.arrayType` and no `processType(t)` at all. After all, a rest parameter will just be an array. Currently, we don't support nested array types (i.e. `.iterable(of: .integer)`) or so, but once we do, we'd need to update this logic. So maybe add a comment about that, too?

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
```diff
@@ -850,53 +789,155 @@ class TypeExtension: Hashable {
...
     }
 
     public var description: String {
         return format(abbreviate: false)
     }
 
-    public init(expects parameterTypes: [Type], returns returnType: Type) {
-        self.inputTypes = parameterTypes
+    // Returns the number of innerOutputs that are generated from the function signature
+    public var numParameters: Int {
```
saelo, line 898: Taking into account future PRs about parameter destructuring, is this really true? I guess for something like
`function foo(a, [b, c, d], e))` we'd have `numParameters = 3` (caller side) but `numInnerOutputs (or numCreatedVariables) = 5` (callee side). So I think we should clarify whether these counts are valid for the caller or the callee :)

Sources/FuzzilliCli/CodeGeneratorWeights.swift
```diff
@@ -34,11 +34,11 @@ let codeGeneratorWeights = [
     "ArrayWithSpreadGenerator":                 5,
     "TemplateStringGenerator":                  1,
     "PlainFunctionGenerator":                   15,
-    "ArrowFunctionGenerator":                   3,
-    "GeneratorFunctionGenerator":               3,
-    "AsyncFunctionGenerator":                   3,
-    "AsyncArrowFunctionGenerator":              1,
-    "AsyncGeneratorFunctionGenerator":          1,
+    "ArrowFunctionGenerator":                   10,
```
saelo, line 37: Accidental changes? :)

Sources/Fuzzilli/FuzzIL/AbstractInterpreter.swift
saelo: I'll create an issue about the `.unknown` vs. `.anything` weirdness

Sources/FuzzilliCli/CodeGeneratorWeights.swift
amarekano: Yeah, I bumped up these values to do some stress testing. I'll roll these back to the original values.

Sources/Fuzzilli/FuzzIL/TypeSystem.swift
amarekano: So the direction I was going with this was that `numParameters` would represent callee side variables. In the example above `numParameters` would be 5 and to calculate the number of caller side arguments (which is what's being done in ProgramBuilder) I use the count from `callerTypes`.
Happy to change this though if that would help simplify this.

saelo: Ok, yeah WDYT about renaming it to something like `numOutputVariablesInCallee`? Maybe a bit verbose, but at least it's fairly obvious (I think) what it expresses. I think it'll be confusing to have `sig.numParameters != sig.parameters.count` in the future... :)

saelo:
\o/ thanks!

PR #319, opened 2022-01-28
ObjectDestructAndReassign can support empty patterns

amarekano:
This PR drops an assert statement since object destruct and reassign operations can support empty patterns. Eg:

```js
[] = {foo:10, bar:20}
```


saelo:
Thanks! This is now included in https://github.com/googleprojectzero/fuzzilli/commit/700f669a1d38b787968229f7d7ab2a4d0a7bf2b7

PR #322, opened 2022-02-02
Add support for optional chaining

amarekano:
This PR adds support for [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -185,7 +185,7 @@ public let CodeGenerators: [CodeGenerator] = [
 
     CodeGenerator("PropertyRetrievalGenerator", input: .object()) { b, obj in
         let propertyName = b.type(of: obj).randomProperty() ?? b.genPropertyNameForRead()
-        b.loadProperty(propertyName, of: obj)
+        b.loadProperty(propertyName, optional: probability(0.25), of: obj)
     },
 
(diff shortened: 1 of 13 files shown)


saelo:
Sorry for the super late response here, I had deprioritized adding new language features to Fuzzilli (as you have probably guessed :)). If you'd still like to merge this PR, I'd suggest the following changes:
- (mostly cosmetic): maybe change the isOptional boolean to a mode enum ("throwing", "nonThrowing" or so?) or something like that? So that it's more clear that this is about returning undefined vs. throwing an exception (IIUC). In the program builder etc. we should default to the throwing mode
- don't let CodeGenerators randomly generate operations with the optional mode. Instead only do that for example when `type(of: object).MayBe(.null | .undefined)` or something like that (i.e. when we know that a non-optional operation may throw at runtime) or for example also if a method may not exist (where we currently use try-catch instead, but this would be nicer)
- don't mutate the optional/non-optional mode in the OperationMutator since that's probably not very useful
- instead, let the SimplifyingMutator attempt to turn optional operations into non-optional ones (since there's a good chance that the operation is valid if it is important)

saelo:
I've made the modifications I've suggested above and submitted it as https://github.com/googleprojectzero/fuzzilli/commit/39d268b4a6e7ca11162339726f2e2ddf8df549cd so I'm closing this now. Thanks for the PR!

PR #323, opened 2022-02-03
Add support for tagged templates

amarekano:
This PR adds support for [tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates)

### DIFF
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -120,7 +120,7 @@ public let CodeGenerators: [CodeGenerator] = [
         b.createTemplateString(from: parts, interpolating: interpolatedValues)
     },
 
-    CodeGenerator("CallTaggedTemplateGenerator", input: .function()) { b, f in
+    CodeGenerator("TaggedTemplateStringGenerator") { b in
         var interpolatedValues = [Variable]()
         for _ in 1..<Int.random(in: 1...5) {
             interpolatedValues.append(b.randVar())
@@ -131,6 +131,13 @@ public let CodeGenerators: [CodeGenerator] = [
             // For now we generate random strings
             parts.append(b.genString())
         }
+
+        // Generate a tag function
+        let f = b.definePlainFunction(withSignature: FunctionSignature(withParameterCount: 1 + interpolatedValues.count, hasRestParam: probability(0.1)), isStrict: probability(0.1)) { _ in
+            b.generateRecursive()
+            b.doReturn(value: b.randVar())
+        }
+
         b.callTaggedTemplate(f, with: parts, interpolating: interpolatedValues)
     },
 
--- a/Sources/Fuzzilli/Core/CodeGenerators.swift
+++ b/Sources/Fuzzilli/Core/CodeGenerators.swift
@@ -120,6 +120,20 @@ public let CodeGenerators: [CodeGenerator] = [
         b.createTemplateString(from: parts, interpolating: interpolatedValues)
     },
 
+    CodeGenerator("CallTaggedTemplateGenerator", input: .function()) { b, f in
+        var interpolatedValues = [Variable]()
+        for _ in 1..<Int.random(in: 1...5) {
+            interpolatedValues.append(b.randVar())
+        }
+
+        var parts = [String]()
+        for _ in 0...interpolatedValues.count {
+            // For now we generate random strings
+            parts.append(b.genString())
+        }
+        b.callTaggedTemplate(f, with: parts, interpolating: interpolatedValues)
+    },
+
     CodeGenerator("BuiltinGenerator") { b in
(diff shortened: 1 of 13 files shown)


Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -279,6 +279,24 @@ class CreateTemplateString: Operation {
...
 
+class CallTaggedTemplate: Operation {
+    // Stores the string elements of the temaplate literal
+    let parts: [String]
+
+    var numInterpolatedValues: Int {
+        return numInputs - 1
+    }
+
+    // This operation isn't parametric since it will most likely mutate imported templates (which would mostly be valid JS snippets) and
```
saelo, line 290: I'm wondering if this is really necessary (also for `CreateTemplateString`). _If_ a string really is valid JavaScript and is subsequently executed, and we mutate it, we'll just end up with an invalid program and undo the mutation. On the other hand, I feel like if we mutate regular strings, we should also mutate template strings. WDYT?
An alternative would be to extend the existing `CodeString` mechanism to support interpolation, but in that case the FuzzIL compiler would have to know when a string is supposed to be JavaScript code, which probably isn't really feasible.

Sources/Fuzzilli/FuzzIL/Operations.swift
```diff
@@ -279,6 +279,24 @@ class CreateTemplateString: Operation {
     }
 }
 
+class CallTaggedTemplate: Operation {
```
saelo, line 282: Nit: maybe this should be called `CreateTaggedTemplateString` to be consistent with `CreateTemplateString`?

Sources/Fuzzilli/Core/CodeGenerators.swift
```diff
@@ -120,6 +120,20 @@ public let CodeGenerators: [CodeGenerator] = [
         b.createTemplateString(from: parts, interpolating: interpolatedValues)
     },
 
+    CodeGenerator("CallTaggedTemplateGenerator", input: .function()) { b, f in
```
saelo, line 123: I'm wondering if this code generator should also generate a random tag function  similar to the other function generators, with something like
```swift
            b.generateRecursive()
            b.doReturn(value: b.randVar())
```
as body. The main benefit I see is that the tag function could be given the appropriate signature here (first arg: array of string, remaining args: anything), because we know it'll be used as a tag. The `InputMutator` will take care of using other functions as tags eventually. Let me know what you think!

Sources/Fuzzilli/FuzzIL/Operations.swift
amarekano: I think the reason we didn't want to do this was due to the high failure rate it would have. I'm happy to change that to enable mutations.
> An alternative would be to extend the existing CodeString mechanism to support interpolation, but in that case the FuzzIL compiler would have to know when a string is supposed to be JavaScript code, which probably isn't really feasible.

Yeah that's going to require some gnarly hacks to parse and compile :(

Sources/Fuzzilli/Core/CodeGenerators.swift
amarekano: Yeah that's a good idea! I'll drop the current input to the generator in that case and add a tag generator.

Sources/Fuzzilli/FuzzIL/Operations.swift
amarekano: Unfortunately no, I don't have data on this.

amarekano: @saelo any thoughts on the above? I can start gathering stats if we need that to make changes to this PR

saelo: Yeah I'd suggest something like running 4-8 Instances of Fuzzilli, half of them with mutation of TaggedTemplateStrings allowed, the other half with it forbidden, and keeping track of the failure rates. If they are not significantly different, I'd prefer if TaggedTemplates were mutable as well (I guess that includes the existing TemplateStrings: https://github.com/googleprojectzero/fuzzilli/blob/4b0df78015bbfb41ff261d366683e3d6f39b3eee/Sources/Fuzzilli/FuzzIL/Operations.swift#L273)

amarekano: Sounds good to me, any ideas on capturing failure rates? I could add a counter to record valid and invalid program generation from the two generators (tagged template and template string). I could also just use the existing correctness rate but that's probably less accurate.

saelo: With some of the recent changes, if you add --logLevel=verbose, you'll get detailed statistics about the correctness rates of every CodeGenerator, Mutator, etc. That's probably good enough

saelo:

PR #325, opened 2022-02-06
Adding missing ConcatMutator

DeamonSpawn:
Pull request to add missing ConcatMutator in Mutators weighted list

### DIFF
--- a/Sources/Fuzzilli/Mutators/ConcatMutator.swift
+++ b/Sources/Fuzzilli/Mutators/ConcatMutator.swift
@@ -14,6 +14,8 @@
 
 /// A mutator that concatenates two programs together.
 public class ConcatMutator: Mutator {
+    public override init() {}
+
     override func mutate(_ program: Program, using b: ProgramBuilder) -> Program? {
         let suffix = b.fuzzer.corpus.randomElementForSplicing()
         
--- a/Sources/Fuzzilli/Mutators/ConcatMutator.swift
+++ b/Sources/Fuzzilli/Mutators/ConcatMutator.swift
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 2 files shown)


Sources/FuzzilliCli/main.swift
```diff
@@ -368,6 +368,7 @@ func makeFuzzer(for profile: Profile, with configuration: Configuration) -> Fuzz
         (CodeGenMutator(),                  3),
         (InputMutator(isTypeAware: false),  2),
         (InputMutator(isTypeAware: true),   1),
+        (ConcatMutator(),                   1),
```
saelo, line 371: The `ConcatMutator` is currently disabled (mostly) on purpose since it's just a more limited version of the `CombineMutator` (the `CombineMutator` will insert one program into another at a random position, the `ConcatMutator` does the same, but only at the end). That said, how about adding the mutator to this list with weight `0` (or commented out), with a short comment explaining that it's disabled as the `CombineMutator` is more general? Then people can easily enable it for experiments if desired.

DeamonSpawn: I have added the comment as you suggested.

Sources/Fuzzilli/Mutators/ConcatMutator.swift
```diff
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// A mutator that concatenates two programs together.
+/// A mutator that concatenates one programs at the end of the other.
```
saelo, line 15: Nit: I think leaving the comment as it was is fine, "concatenating" should already imply that it's inserted at the end :)

DeamonSpawn: I have reverted the comment. :)

saelo:
Thanks!

PR #328, opened 2022-02-20
Update readme, corpus + mutation engine path

Microsvuln:
Hi, fix readme, adding the proper path of mutation engine and corpus .


saelo:
Nice, thanks!

PR #332, opened 2022-03-25
add XS target and profile

phoddie:


Targets/XS/README.md
```diff
@@ -0,0 +1,15 @@
+# Target: XS
+
+To build XS for fuzzing with Fuzzilli:
+
+1. Clone the Moddable SDK from [https://github.com/Moddable-OpenSource/moddable](https://github.com/Moddable-OpenSource/moddable)
+2. Build `xst`, the XS test tool:
```
saelo, line 6: I think these instructions aren't quite complete: I had to add an `export MODDABLE=<path>` as described [here](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/Moddable%20SDK%20-%20Getting%20Started.md#macos) and the path seems to be `$MODDABLE/xs/makefiles/mac` (not `.../xs/make/...`). But then everything seemed to work fine! Maybe just link to the official build instructions, then add the bit about building `xst` via `FUZZILLI=1 make` in `$MODDABLE/xs/makefiles/mac`?

saelo (commented):
Cool, thanks!

phoddie:
Thank for you for giving it a try. Very cool that it (eventually) worked for you. Apologies for the errors in the readme. I've tried to address those.

saelo:
That's great, thanks!

PR #340, opened 2022-04-03
Update QJS target for latest revision and fix broken link.

jessesomerville:
The official Github mirror for QuickJS was moved and the current link in the QJS target README is broken because of it.  The changes in this PR update the README to point to the new location of the official mirror (which I got from https://bellard.org/quickjs/) and updates the QuickJS patch to work with the latest revision in that repo.  In addition to that, I made a small change to the Makefile patch to ensure that `make qjs` uses `clang` rather than `gcc` when running on non-Darwin systems.

Side note: The QuickJS source uses 4 space indentation rather than 2, so I made the patch compliant with that.  This is just to prevent formatting oddities after applying the patch caused by discrepancies in indentation.


Targets/QJS/Patches/Fuzzilli-instrumentation-for-QJS.patch
```diff
@@ -131,198 +144,192 @@ index 7e4839f..7dbf2b6 100644
...
-+    return JS_FALSE;
-+  }
-+  if (!strcmp(str, "FUZZILLI_CRASH")) {
-+    printf("js_fuzzilli CRASH\n");
-+    int type;
-+    if (JS_ToInt32(ctx, &type, argv[1])) {
-+      JS_FreeCString(ctx, str);
-+      printf("js_fuzzilli wrong type\n");
-+      return JS_FALSE;
++    printf("js_fuzzilli\n");
```
saelo, line 286: Nit: should we drop these printfs? The crash testers in https://github.com/googleprojectzero/fuzzilli/blob/8431a409dc431b04438511a684af598428aa147f/Sources/FuzzilliCli/Profiles/QjsProfile.swift#L34 should notice if this logic breaks I think

Targets/QJS/Patches/Fuzzilli-instrumentation-for-QJS.patch
```diff
@@ -131,198 +144,192 @@ index 7e4839f..7dbf2b6 100644
...
-  * THE SOFTWARE.
-  */
-+
- #include <stdlib.h>
- #include <stdio.h>
+@@ -27,6 +27,7 @@
  #include <stdarg.h>
  #include <inttypes.h>
  #include <string.h>
 +#undef NDEBUG
```
saelo, line 275: I realize this was here before, but is this actually required? Shouldn't debug or no-debug configuration be configured elsewhere?

saelo (commented):
Great, thanks! Just two questions

Targets/QJS/Patches/Fuzzilli-instrumentation-for-QJS.patch
jessesomerville: Yep, these were in the original patch as well.  I've removed them.

Targets/QJS/Patches/Fuzzilli-instrumentation-for-QJS.patch
jessesomerville: Yeah I didn't think these did anything because NDEBUG isn't being defined anywhere else as far as I can tell, but I left them just in case.  I've removed this line in both the qjs.c and quickjs.c files in the patch.

jessesomerville:
Force push was to amend the previous commit because I had added the wrong patch file

saelo:
Perfect, thanks!

PR #342, opened 2022-05-12
use #ifdef __cplusplus macro

dramthy:


saelo:
Oh looks like this PR slipped, sorry! I'm going to close this for now since it's quite old, but feel free to just reopen if it's still relevant. I guess my question would be what the use case for checking the `__cplusplus` macro is in this case?

Issue #172, opened 2020-11-19
Implement More Intelligent Mutation/Code Generator/Constant Selection Scheduler


WilliamParks:
Fuzzilli current uses hard-coded weights to select between the various [mutators](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/FuzzilliCli/main.swift#L295), [code generators](https://github.com/googleprojectzero/fuzzilli/blob/1cd76bf9e8f32fa4ce5d350a9122228d20f913a5/Sources/FuzzilliCli/CodeGeneratorWeights.swift), and [constants](https://github.com/googleprojectzero/fuzzilli/blob/ce4738fc571e2ef2aa5a30424f32f7957a70b5f3/Sources/Fuzzilli/Core/ProgramBuilder.swift#L128). These weights are just approximations, and a more intelligent mutator scheduler could help improve performance, by giving additional weight to those likely to find new coverage, or be successful.

One possible example would be something like [MOpt](https://www.usenix.org/system/files/sec19-lyu.pdf), although it may be overkill at this time.

First thoughts on requirements for a new scheduler:

- Applies more emphasis to mutators/code generators/constant selectors that are more likely to run successfully  or produce new coverage
- Have a "weight floor", intermittent re-normalization, etc, such that some no mutator/code generator/constant is ever completely ignored

saelo:
Great idea! Some more initial thoughts:
* Besides the weights of the mutators, it would also be nice to automatically determine a somewhat optimal "aggressivity" for each mutator, i.e. [how many changes are performed by one invocation of the mutator](https://github.com/googleprojectzero/fuzzilli/blob/ce4738fc571e2ef2aa5a30424f32f7957a70b5f3/Sources/Fuzzilli/Mutators/BaseInstructionMutator.swift#L17)
* While (as you already noted) the success rate could also be used as a second input to the algorithm besides the "interesting" rate, I think it's quite possible that a mutator with a fairly high "failure" rate (producing lots of semantically invalid samples) actually has a very good "interesting" rate. Probably something to evaluate later on

ducphanduyagentp:
Hi. I am working on this but I am quite new to developing Fuzzilli. Should I make a WIP PR so I can get more feedback? My code is now probably really messy 😅Thank you!

saelo:
Hey! Sure, feel free to open a WIP PR for early feedback if you think it helps :)

DeamonSpawn:
[...]

PR #343, opened 2022-06-11
Intelligent Selection of Code Generation, Mutation and Seed Selection with Multi-Armed Bandit

DeamonSpawn:
Coverage-based guidance mechanism for code generation, mutation, and seed selection tasks to optimize coverage growth. Utilizes the [Multi-Armed Bandit Algorithms](https://en.wikipedia.org/wiki/Multi-armed_bandit) to navigate the search spaces for selection of tasks.  

Addressing issue #172 . 

Thesis with design and implementation:
[Intelligent Code Generation/ Mutation to aid fuzzing of JavaScript engines](https://github.com/DeamonSpawn/amenezes.github.io/blob/b29efd8b89a1945ed743826547e87081e474ab92/Docs/IntelligentCodeGenerationAndMutationToAidFuzzingOfJavaScriptEngines.pdf)

### DIFF
--- a/Cloud/GCE/push.sh
+++ b/Cloud/GCE/push.sh
@@ -7,5 +7,5 @@ set -e
 
 source config.sh
 
-docker tag $CONTAINER_NAME:latest $REGION-docker.pkg.dev/$PROJECT_ID/fuzzilli-docker-repo/$CONTAINER_NAME
-docker push $REGION-docker.pkg.dev/$PROJECT_ID/fuzzilli-docker-repo/$CONTAINER_NAME
+docker tag fuzzilli gcr.io/$PROJECT_ID/$CONTAINER_NAME
+docker push gcr.io/$PROJECT_ID/$CONTAINER_NAME
--- a/Cloud/GCE/push.sh
+++ b/Cloud/GCE/push.sh
@@ -7,5 +7,5 @@ set -e
 
 source config.sh
 
-docker tag fuzzilli gcr.io/$PROJECT_ID/$CONTAINER_NAME
-docker push gcr.io/$PROJECT_ID/$CONTAINER_NAME
+docker tag $CONTAINER_NAME:latest $REGION-docker.pkg.dev/$PROJECT_ID/fuzzilli-docker-repo/$CONTAINER_NAME
+docker push $REGION-docker.pkg.dev/$PROJECT_ID/fuzzilli-docker-repo/$CONTAINER_NAME
--- a/Sources/FuzzilliCli/main.swift
+++ b/Sources/FuzzilliCli/main.swift
@@ -31,7 +31,7 @@ Options:
     --jobs=n                    : Total number of fuzzing jobs. This will start one master thread and n-1 worker threads. Experimental!
     --engine=name               : The fuzzing engine to use. Available engines: "mutation" (default), "hybrid", "multi".
                                   Only the mutation engine should be regarded stable at this point.
-    --corpus=name               : The corpus scheduler to use. Available schedulers: "basic" (default), "markov"
+    --corpus=name               : The corpus scheduler to use. Available schedulers: "basic" (default), "markov", "mab"
     --minDeterminismExecs=n     : The minimum number of times a new sample will be executed when checking determinism (default: 3)
     --maxDeterminismExecs=n     : The maximum number of times a new sample will be executed when checking determinism (default: 50)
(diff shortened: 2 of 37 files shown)


Cloud/GCE/push.sh
```diff
@@ -7,5 +7,5 @@ set -e
 
 source config.sh
 
-docker tag fuzzilli gcr.io/$PROJECT_ID/$CONTAINER_NAME
-docker push gcr.io/$PROJECT_ID/$CONTAINER_NAME
+docker tag $CONTAINER_NAME:latest $REGION-docker.pkg.dev/$PROJECT_ID/fuzzilli-docker-repo/$CONTAINER_NAME
```
saelo, line 10: I'd prefer if changes to these files could be a separate CL (if you want to include them) :)

Sources/FuzzilliCli/main.swift
```diff
@@ -593,9 +600,17 @@ signalSources.append(source)
 #endif
 
 // Finally, start fuzzing.
-for fuzzer in instances {
-    fuzzer.sync {
-        fuzzer.start(runFor: numIterations)
+let master = fuzzer
```
saelo, line 603: Ah, is this an independent bug fix? Would the thread workers otherwise not get the initial corpus? This should probably be it's own PR as well then

saelo (commented):
That's amazing, thanks! A warning up front, this is quite a large PR so it'll take me some time to get through it :D

Do you want me to run some (more) fuzzing sessions (against v8) for evaluation with this enabled vs disabled? I could probably do a few sessions up to 1B iterations (each time enabled vs. disabled). Would it make sense to test e.g. only MABCorpus or CodeGeneration at a time or is it ok to enable all MAB-ed "things" at the same time?

I think the first step towards merging this is to split up the PR into multiple smaller ones. I could imagine:
- 1-2 PRs for various unrelated fixes (see e.g. comments)
- 1 PR for the CodeGenerator changes (the CodeGenerationMode)
- Either one big PR for the rest, or one PR per MAB "thing". I'm not yet sure which makes more sense

I think it'd also be good to have a short, high-level description of how the MAB algorithm works somewhere in the code, similar to https://github.com/googleprojectzero/fuzzilli/blob/main/Sources/Fuzzilli/Corpus/MarkovCorpus.swift (and also link to the paper of course!). Would it be possible to have one "generic" MAB implementation that is then used for the corpus, the code generators, and the mutators? From a quick look it seems like at least some of the MAB-related logic is duplicated in a few places. Probably it's not going to be completely generic, but maybe it'll be good enough with the right abstractions. WDYT?

Cloud/GCE/push.sh
DeamonSpawn: I'll revert this change it is not relevant to the PR. :)

Sources/FuzzilliCli/main.swift
DeamonSpawn: The reason this is done, is to synchronize the MAB state and the corpora across each worker.
Without this change, only interesting programs would be distributed to the worker nodes and not the MAB state and compiled seeds.

DeamonSpawn:
To answer your first question,
I have performed tests for a single node setup of Fuzzilli without any distributed instances (no master-worker network or threading nodes) and no compiled/imported seeds used in the corpus.
The tests where evaluated over 24 hours.
I have conducted tests with the MAB implementation against instances with the basic (default) corpus and the Markov corpus respectively.
These tests include MAB Code Gen\Mutator and MAB Corpus with their combined and individual contributions.

**Summary Observations:**
The performance of each baseline instances (Basic and Markov) and MAB implementations (MAB Code Gen\Mutator and MAB Corpus) eventually converges after a period of 24 hours to the same rate of coverage discovery.
In the isolated evaluation of MAB Code Gen\Mutator, the instance is found to reach the convergence point faster than Basic and Markov.
In the isolated evaluation of MAB Corpus, the performance against baseline instances shows a faster coverage growth rate with higher coverage discovered in early time intervals before convergence.
With the combined performance of MAB CodeGen\Mutator and MAB Corpus the higher coverage growth rate and convergence rate show an improved exploration of the corpus search space reducing the time taken for coverage discovery

I am waiting on the assessment of my Masters thesis before I upload my report with details to my implementation.

Networked nodes have been tested locally and are very much capable of operating with MAB Corpus.
However, stats have not been collected for distributed nodes over 24 hours.

PR #347, opened 2022-07-14
Docker fix for misssing packages

DeamonSpawn:


Cloud/Docker/JSCBuilder/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM swift:focal
 ENV DEBIAN_FRONTEND=noninteractive
 ENV SHELL=bash
 
-RUN apt-get -y update && apt-get -y upgrade
+RUN apt-get -y update --fix-missing && apt-get -y upgrade
```
saelo, line 6: So in which situation is this needed? (I'm not sure I understand what --fix-missing actually does, it sounds like it would let you ignore errors due to missing packages, but that shouldn't happen for us?)

saelo:
Closing for now since there were some changes to the dockerfiles in the meantime, but feel free to reopen!

PR #355, opened 2022-08-09
Add `cause` property to `Error` and add variant of `URIError` to `Error` object

0xedward:


Sources/Fuzzilli/Core/JavaScriptEnvironment.swift
```diff
@@ -1116,6 +1116,7 @@ public extension ObjectGroup {
                 "constructor" : .function(),
                 "message"     : .jsString,
                 "name"        : .jsString,
+                "cause"       : .jsError,
```
saelo, line 1119: IIUC, `cause` doesn't need to be an Error but could be anything: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause
```
The value can be of any type. You should not make assumptions that the error you caught has an Error as its cause
```
So maybe this should be `.anything` instead?

0xedward: Ah, thanks! Sorry I missed that in the docs earlier and it also looks there are [some test cases in test262](https://github.com/tc39/test262/tree/main/test/built-ins/Error) for when `cause` is not a `Error` object

saelo:
Thanks! :)

PR #358, opened 2022-08-26
Fix `Symbol.unscopables` in runtime environment

0xedward:
The runtime environment currently has `Symbol.unscopable` instead of [`Symbol.unscopables`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/unscopables)


saelo:
Nice catch, thanks!

PR #370, opened 2022-09-14
Add `--statisticsExportInterval option` for specifying interval for saving fuzzing stats to disk

0xedward:
### Test Plan
I ran the following commands, we expect Fuzzilli to begin fuzzing:
- `swift run -c release FuzzilliCli --profile=v8 --storagePath=./test --exportStatistics d8` to test that `--statisticsExportInterval` is an optional parameter

- `swift run -c release FuzzilliCli --profile=v8 --storagePath=./test --exportStatistics --statisticsExportInterval=1 d8` and let it run for 1 minute, then stopped Fuzzilli and checked if `./test/stats` contains a single json file

I ran the following commands to check for invalid options:

- `swift run -c release FuzzilliCli --profile=v8 --storagePath=./test --statisticsExportInterval=1 d8` will print `statisticsExportInterval requires --exportStatistics` 

- `swift run -c release FuzzilliCli --profile=v8 --storagePath=./test --exportStatistics --statisticsExportInterval=0 d8` will print `statisticsExportInterval needs to be > 0`

### DIFF
--- a/Sources/FuzzilliCli/main.swift
+++ b/Sources/FuzzilliCli/main.swift
@@ -26,73 +26,73 @@ Usage:
 \(args.programName) [options] --profile=<profile> /path/to/jsshell
 
 Options:
-    --profile=name              : Select one of several preconfigured profiles.
-                                  Available profiles: \(profiles.keys).
-    --jobs=n                    : Total number of fuzzing jobs. This will start one master thread and n-1 worker threads.
-    --engine=name               : The fuzzing engine to use. Available engines: "mutation" (default), "hybrid", "multi".
-                                  Only the mutation engine should be regarded stable at this point.
-    --corpus=name               : The corpus scheduler to use. Available schedulers: "basic" (default), "markov"
-    --minDeterminismExecs=n     : The minimum number of times a new sample will be executed when checking determinism (default: 3)
-    --maxDeterminismExecs=n     : The maximum number of times a new sample will be executed when checking determinism (default: 50)
-    --noDeterministicCorpus     : Don't ensure that samples added to the corpus behave deterministically.
-    --maxResetCount=n           : The number of times a non-deterministic edge is reset before it is ignored in subsequent executions.
-                                  Only used as part of --deterministicCorpus.
-    --logLevel=level            : The log level to use. Valid values: "verbose", info", "warning", "error", "fatal"
-                                  (default: "info").
-    --numIterations=n           : Run for the specified number of iterations (default: unlimited).
-    --timeout=n                 : Timeout in ms after which to interrupt execution of programs (default: 250).
-    --minMutationsPerSample=n   : Discard samples from the corpus only after they have been mutated at least this many times (default: 100).
-    --minCorpusSize=n           : Keep at least this many samples in the corpus regardless of the number of times
-                                  they have been mutated (default: 1024).
-    --maxCorpusSize=n           : Only allow the corpus to grow to this many samples. Otherwise the oldest samples
-                                  will be discarded (default: unlimited).
-    --markovDropoutRate=p       : Rate at which low edge samples are not selected, in the Markov Corpus Scheduler,
-                                  per round of sample selection. Used to ensure diversity between fuzzer instances
-                                  (default: 0.10)
-    --consecutiveMutations=n    : Perform this many consecutive mutations on each sample (default: 5).
-    --minimizationLimit=p       : When minimizing interesting programs, keep at least this percentage of the original instructions
-                                  regardless of whether they are needed to trigger the interesting behaviour or not.
-                                  See Minimizer.swift for an overview of this feature (default: 0.0).
-    --storagePath=path          : Path at which to store output files (crashes, corpus, etc.) to.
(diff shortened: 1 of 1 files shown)


Sources/FuzzilliCli/main.swift
```diff
@@ -217,6 +220,14 @@ if exportStatistics && storagePath == nil {
     configError("--exportStatistics requires --storagePath")
 }
 
+if args.double(for: "--statisticsExportInterval") == nil || statisticsExportInterval <= 0.0 {
```
saelo, line 223: I think you can just drop the `args.double(for: "--statisticsExportInterval") == nil ||` part entirely since the default value is valid

Sources/FuzzilliCli/main.swift
```diff
@@ -217,6 +220,14 @@ if exportStatistics && storagePath == nil {
...
+    configError("statisticsExportInterval needs to be > 0.0")
```
saelo, line 224: Nit maybe just make it `0` instead of `0.0`

Sources/FuzzilliCli/main.swift
```diff
@@ -135,6 +137,7 @@ let storagePath = args["--storagePath"]
 var resume = args.has("--resume")
 let overwrite = args.has("--overwrite")
 let exportStatistics = args.has("--exportStatistics")
+let statisticsExportInterval = args.double(for: "--statisticsExportInterval") ?? 10.0
```
saelo, line 140: Can we just make it a UInt instead? If there's need for sub-minute intervals, we should probably just specify it in seconds, but I think minutes is good enough.

Sources/FuzzilliCli/main.swift
```diff
@@ -217,6 +220,14 @@ if exportStatistics && storagePath == nil {
     configError("--exportStatistics requires --storagePath")
 }
 
+if args.double(for: "--statisticsExportInterval") == nil || statisticsExportInterval <= 0.0 {
+    configError("statisticsExportInterval needs to be > 0.0")
+}
+
+if args.double(for: "--statisticsExportInterval") != nil && !exportStatistics  {
```
saelo, line 227: Nit: maybe just `args.has("--statisticsExportInterval")` for better readability?

Sources/FuzzilliCli/main.swift
```diff
@@ -26,72 +26,74 @@ Usage:
...
+                                      per round of sample selection. Used to ensure diversity between fuzzer instances
+                                      (default: 0.10)
+    --consecutiveMutations=n        : Perform this many consecutive mutations on each sample (default: 5).
+    --minimizationLimit=p           : When minimizing interesting programs, keep at least this percentage of the original instructions
+                                      regardless of whether they are needed to trigger the interesting behaviour or not.
+                                      See Minimizer.swift for an overview of this feature (default: 0.0).
+    --storagePath=path              : Path at which to store output files (crashes, corpus, etc.) to.
+    --resume                        : If storage path exists, import the programs from the corpus/ subdirectory
+    --overwrite                     : If storage path exists, delete all data in it and start a fresh fuzzing session
+    --exportStatistics              : If enabled, fuzzing statistics will be collected and saved to disk every 10 minutes.
```
saelo, line 59: Maybe replace the `every 10 minutes` part with something like ` in regular intervals`

Sources/FuzzilliCli/main.swift
```diff
@@ -26,72 +26,74 @@ Usage:
...
-                                  executions on the current REPRL instance.
-    --inspect=opt1,opt2,...     : Enable inspection options. The following options are available:
-                                      history: Additional .fuzzil.history files are written to disk for every program.
-                                               These describe in detail how the program was generated through mutations,
-                                               code generation, and minimization
-                                        types: Programs written to disk also contain variable type information as
-                                               determined by Fuzzilli as comments
-                                          all: All of the above
-""")
+    --profile=name                  : Select one of several preconfigured profiles.
```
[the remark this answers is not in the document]
0xedward: Oh, I added a tab to all options in a separate commit (https://github.com/googleprojectzero/fuzzilli/pull/370/commits/98a67bcabd451c5e73fb985abc02a5cf142dbed8), because `statisticsExportInterval` is just 1 character too long for keeping the `:` aligned with everything else. Let me know if that's okay with you :)

https://github.com/googleprojectzero/fuzzilli/blob/169fef7e1041816454e8689799485a863bac90d7/Sources/FuzzilliCli/main.swift#L59-L66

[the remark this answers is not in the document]
saelo: Ah, hm we use spaces instead of tabs throughout the project, so I'd prefer if we can keep the spaces here :/ I guess that'd mean adding a space in front of every `:` here to fix the alignment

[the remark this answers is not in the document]
0xedward: Oh, got it! I deleted the tab I added and replaced it with a space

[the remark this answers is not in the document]
saelo: perfect, thanks! :)

PR #377, opened 2022-10-05
Update Jerryscript patches to work against v2.4.0.

jamie-garside:
jerry-main.c has undergone some significant refactoring between 0f0041d and v2.4.0. Now that jerry-main.c has been refactored into separate modules, we can easily split all the Fuzzilli-related logic into its own file and patch that into main() to make it easier to rebase in future.


saelo:
Nice, thanks!

PR #378, opened 2022-10-10
Differential testing

bernhl:
Warning ahead: this PR is not yet complete as it modifies the reprl interface and only v8, spidermonkey and jsc are updated accordingly.
Alas, the PR adds differential testing to Fuzzilli. If differential testing is active, there will be two instances of the JS engine, one "regular" engine + another one with JIT disabled for comparison. Fuzzilli now inserts calls to a `fuzzilli_hash(x)` function into the generated samples. The input parameter `x` is hashed and updates a (non-crypto) hash value. At the end of the execution the hash value is send back to the Fuzzer (via the reprl interface). The JS engines can implement this by any mean deemed suitable, examples are included for v8, spidermonkey and jsc.

Right now, there are two options of activating differential testing; the difference being only the locations where `fuzzilli_hash` is inserted. Either via `--differentialRate=p`, which will cause insertion of calls to `fuzzilli_hash` only at the end of generated samples. This is supposed to be the "default", as the calls to `fuzzilli_hash` might disturb jitting. There is a second option,  `--differentialWeaveRate=p`, which inserts  `fuzzilli_hash` throughout the entire sample (at random locations). This is currently only recommended for spidermonkey as the included patches decrease jit disturbances.
So, please let me know what you think regarding the PR so I can iron out the remaining rough edges or clarify the design.


Sources/FuzzilliCli/Profiles/V8Profile.swift
```diff
@@ -270,10 +275,24 @@ let v8Profile = Profile(
         return args
     },
 
+    processArgumentsReference: [
```
saelo, line 278: Here and in the other two profiles, I think it'd be nice to split these into `commonArguments`, `optimizingArguments`, and `referenceArguments` or so, so that e.g. `--harmony` is only specified in one place.

Sources/FuzzilliCli/main.swift
```diff
@@ -317,8 +329,19 @@ for (generator, var weight) in (additionalCodeGenerators + regularCodeGenerators
 //
 
 func makeFuzzer(for profile: Profile, with configuration: Configuration) -> Fuzzer {
+    let differentialTesting = isNetworkMaster || configuration.differentialRate > 0.0 || configuration.differentialWeaveRate > 0.0
```
saelo, line 332: Why the `isNetworkMaster` flag?

Sources/Fuzzilli/Mutators/ExplorationMutator.swift
```diff
@@ -88,6 +89,9 @@ public class ExplorationMutator: Mutator {
                 // exploration, leading to a failed execution. Failed executions are therefore expected, but if the failure rate appears unreasonably high, one could call
                 // maybeLogFailingExecution here and enable verbose mode to investigate.
                 return failure(.instrumentedProgramFailed)
+            } else if case .differential = execution.outcome {
```
saelo, line 92: If we disable differential fuzzing for the instrumentedProgram (see above), just `fatalError` if we see a `.differential` result here instead.

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -545,6 +600,41 @@ public class Fuzzer {
...
+                program.comments.add("STDOUT:\n" + stdout, at: .footer)
+                program.comments.add("ARGS: \(runner.processArguments.joined(separator: " "))\n", at: .footer)
+                program.comments.add("REFERENCE ARGS: \(referenceRunner!.processArguments.joined(separator: " "))\n", at: .footer)
+            }
+            assert(program.comments.at(.footer)?.contains("DIFFERENTIAL INFO") ?? false)
+
+            // Check for uniqueness only after minimization
+            let execution = execute(program, withTimeout: self.config.timeout * 2)
+            if case .differential = execution.outcome {
+                // TODO don't use crash eval
```
saelo, line 619: I agree. I'm not sure there's any value in trying to determine if a differential is unique (but maybe there is...?), so I'm also fine with just treating all differential programs as unique.

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -545,6 +600,41 @@ public class Fuzzer {
...
+                dispatchEvent(events.DifferentialFound, data: (program, .flaky, true, origin))
+            }
+        }
+
+        if !origin.requiresMinimization() {
+            return processCommon(program)
+        }
+
+        fuzzGroup.enter()
+        minimizer.withMinimizedCopy(program, withAspects: ProgramAspects(outcome: .differential), limit: 0) { minimizedProgram in
```
saelo, line 632: Nit: can drop the `limit: 0` as that's the default value now.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -453,6 +453,15 @@ class Return: JsOperation {
     }
 }
 
+class DifferentialHash: Operation {
+    let allowInnerScope: Bool
```
saelo, line 457: Nit: maybe add a comment explaining what `allowInnerScope` means

Sources/Fuzzilli/Execution/Execution.swift
```diff
@@ -58,5 +64,12 @@ public protocol Execution {
...
     /// Execution time in microseconds
-    var execTime: TimeInterval { get }
+    var execTime: TimeInterval { get set }
+
+    /// A serialized version of (some of) the programs variables after
+    /// execution is finished. Might be empty if the result was not requested
+    /// or could not be computed.
+    var differentialResult: Int { get }
+
+    var differentialResultInputs: Int { get }
```
saelo, line 74: Nit: and what is this? :)

Sources/Fuzzilli/Core/ProgramBuilder.swift
```diff
@@ -1685,6 +1748,23 @@ public class ProgramBuilder {
         // Update type information
         jsTyper.analyze(instr)
 
+        if let op = instr.op as? DifferentialHash, op.allowInnerScope {
```
saelo, line 1751: I'm wondering if it'd be nicer to have this (and the logic in `appendDifferentialProbes`) run as a "PostProcessing" step. How about:
- Create a new file "DifferentialFuzzing.swift" in maybe Core/
- Add a class "DifferentialFuzzing" with a static function `instrumentForDifferentialFuzzing(_ program: Program) -> Program` and move this logic in there (create a new ProgramBuilder, append all instructions from the old program, and randomly insert these `DifferentialHash` operations)
- Call that function from fuzzer.execute if we're doing differential fuzzing

WDYT?

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -453,6 +453,15 @@ class Return: JsOperation {
     }
 }
 
+class DifferentialHash: Operation {
```
saelo, line 456: One thing I'm wondering about here is whether we expect this operation to be present in corpus samples or not. Is this operation just emitted whenever a fuzzer testcase is executed, then removed afterwards (e.g. during minimization), similar to the `Explore` operation for example, or will it persist in the samples and be part of the corpus?

Sources/libreprl/include/libreprl.h
```diff
@@ -51,7 +51,7 @@ void reprl_destroy_context(struct reprl_context* ctx);
 /// @param execution_time A pointer to which, if execution succeeds, the execution time in microseconds is written to
 /// @param fresh_instance if true, forces the creation of a new instance of the target
```
saelo, line 52: Nit: maybe add the parameters to this comment?

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -158,6 +164,9 @@ class REPRLExecution: Execution {
     private var cachedStderr: String? = nil
     private var cachedFuzzout: String? = nil
 
+    var differentialResult: Int
+    var differentialResultInputs: Int
```
saelo, line 168: I guess I know what `differentialResult` is, but what is `differentialResultInputs`?

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: Another thought, could it make sense to also add a `Return` to every instrumented function (if there isn't one already) to return a random local variable, then insert these hashes for all return values?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -438,6 +461,12 @@ public class Fuzzer {
         return corpus.supportsFastStateSynchronization
     }
 
+    private func containsPoison(execution: Execution, differentialPoison: [String]) -> Bool {
```
saelo, line 464: What does this do? Does it somehow mark samples as invalid so they aren't compared for differential results?

bernhl: Correct; some sample trigger behavior within the engines which renders any differentials invalid. E.g., the amount of stack memory differs between interpreter/jit, so one of them might overflow whereas the other doesn't. We must not compare the result of these executions as this would be a false positive.

Sources/Fuzzilli/Execution/REPRL.swift
bernhl: `differentialResultInputs` is the number of inputs which contributed to the hash value (i.e., the number of calls zu fuzzilli_hash). I found this information useful during debugging but we might want to remove it for production

Sources/Fuzzilli/FuzzIL/JsOperations.swift
bernhl: Removing it entirely from the corpus (but obviously not from samples producing differentials) seems reasonable. How would I handle interesting samples where the removal of `DifferentialHash` changes the coverage?

Sources/FuzzilliCli/main.swift
bernhl: If an instance receives differential samples via the network we need the reference runner to determine whether they're flaky or deterministic.

Sources/Fuzzilli/Mutators/ExplorationMutator.swift
```diff
@@ -46,12 +46,13 @@ public class ExplorationMutator: Mutator {
         case success = "Success"
         case cannotInstrument = "Cannot instrument input"
         case instrumentedProgramCrashed = "Instrumented program crashed"
+        case instrumentedProgramDifferential = "Instrumented program produced a differential"
```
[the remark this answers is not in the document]
bernhl: This is a good point; I had one instance producing non-reproducible differentials and this might be the root cause. In the next version of the PR I'll include the required functionality.

Sources/Fuzzilli/Fuzzer.swift
bernhl: I'll update the PR s.t. all differentials are treated as unique; we'd only indicate whether they're flaky or deterministic.

Sources/FuzzilliCli/main.swift
saelo: I think I'd prefer to just log an error in that case and save the sample as flaky crash without trying to reproduce it. I think that should be fairly easy to do? This way we avoid spinning up unused processes for non-differential fuzzing sessions

Sources/Fuzzilli/Mutators/ExplorationMutator.swift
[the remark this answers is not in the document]
saelo: The same is now true for the ProbingMutator :)

Sources/Fuzzilli/Configuration.swift
```diff
@@ -23,6 +23,28 @@ public struct Configuration {
     /// Used to verify that crashes can be detected.
     public let crashTests: [String]
 
+    /// Code snippets that cause an observable difference of output
+    /// in the target engine. Used to verify that crashes can be detected.
+    public let differentialTests: [String]
+
+    /// Code snippets that must not cause an observable difference of output
+    /// in the target engine. Used to verify that common sources of
+    /// entrpy (Math.random, ...) are deterministic.
```
saelo, line 32: Nit: `entropy`

Sources/Fuzzilli/Fuzzer.swift
saelo: Makes sense, thanks! Maybe we should rename it to something more descriptive, e.g. `shouldExecutionBeIgnoredForDifferentialTesting`, and then I guess it can take the `differentialPoison` directly from the Configuration?

Sources/Fuzzilli/Core/ProgramBuilder.swift
saelo: That's a good point! I think the logic should move into the FuzzEngines, since that's where samples are initially generated. I could imagine something along the lines of:
- Make FuzzEngine a class instead of a protocol
- Add a `doDifferentialFuzzTesting` boolean member and `instrumentForDifferentialTesting` method. The latter contains this logic here
- In FuzzEngine.execute, check if we're doing differential fuzzing, and if so instrument the program

WDYT?

Sources/Fuzzilli/FuzzIL/JsOperations.swift
saelo: 👍  in that case please make it inherit from `InternalOperation`, then we'll assert that it doesn't get into the Corpus. I think we don't really need to worry about the coverage. The edges that this triggers will be marked as non-deterministically fairly quickly, at which point they will no longer trigger. Might be worth adding a comment that explains this at the place where the `DifferentialHash` operations are removed.

Sources/Fuzzilli/Core/ProgramBuilder.swift
bernhl: DifferentialHash is not an internal operation and should never be emitted into the corpus. It is added within the FuzzEngine code. What do you think of the updated design?

Sources/Fuzzilli/Engines/FuzzEngine.swift
```diff
@@ -20,15 +20,68 @@ public protocol FuzzEngine: ComponentBase {
 }
 
 extension FuzzEngine {
+    private func instrumentForDifferentialTesting(_ program: Program) -> Program {
```
saelo, line 23: Could you add a short comment describing what this function is trying to achieve? From what I can tell it does
1. Create a number of global variables that will be used to store "regular" differential values
2. Interleave random `fuzzilli_hash` calls for local variables
3. Randomly assign local variables to the global variables created in step 1
4. Hash all of the variables created in step 1

Sources/Fuzzilli/Engines/FuzzEngine.swift
```diff
@@ -20,15 +20,68 @@ public protocol FuzzEngine: ComponentBase {
...
+        return Program(code: b.finalize().code, parent: program.parent,
+                       comments: program.comments, contributors: program.contributors)
+    }
+
     public func execute(_ program: Program) -> ExecutionOutcome {
         fuzzer.dispatchEvent(fuzzer.events.ProgramGenerated, data: program)
 
-        let execution = fuzzer.execute(program)
+        let postprocessedProgram: Program
+        if fuzzer.config.differentialRate > 0.0 || fuzzer.config.differentialWeaveRate > 0.0 {
```
saelo, line 70: Nit: maybe add a single `fuzzer.config.performDifferentialTesting` or so member (can be computed and return this expression here)?

Sources/Fuzzilli/Engines/FuzzEngine.swift
```diff
@@ -20,15 +20,68 @@ public protocol FuzzEngine: ComponentBase {
...
+
+        return Program(code: b.finalize().code, parent: program.parent,
+                       comments: program.comments, contributors: program.contributors)
+    }
+
     public func execute(_ program: Program) -> ExecutionOutcome {
         fuzzer.dispatchEvent(fuzzer.events.ProgramGenerated, data: program)
 
-        let execution = fuzzer.execute(program)
+        let postprocessedProgram: Program
```
saelo, line 69: Another small nit: would it be better to use
```
let program = fuzzer.config.performDifferentialTesting ? instrumentForDifferentialTesting(program) : program
```
That way there's no risk of mixing up `program` and `postprocessedProgram` below.

Sources/Fuzzilli/Engines/FuzzEngine.swift
saelo, line 23: Maybe add an `assert(fuzzer.config.performDifferentialTesting)` here?

Sources/Fuzzilli/Engines/FuzzEngine.swift
```diff
@@ -20,15 +20,68 @@ public protocol FuzzEngine: ComponentBase {
...
+                        let v = b.randVar()
+                        if !reassignVars.contains(v) {
+                            b.differentialHash(v)
+                            break
+                        }
+                    }
+                }
+
+                guard reassignedVars < reassignVars.count else { continue }
+                if probability(fuzzer.config.differentialRate) {
```
saelo, line 51: IIUC, we could end up not using some of the variables created above?
I'm wondering if we can implement this by setting properties/elements on a global object instead of setting global variables? I.e. at the start we'd create a single object, then here we'd set a property/element on it, and below we'd hash all properties/elements of that object. This way, we don't have to create up-front global variables for the hashes. WDYT?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -426,16 +467,47 @@ public class Fuzzer {
...
+        if differentialResult && shouldExecutionBeIgnoredForDifferentialTesting(execution: execution) {
+            // deliberate crash by the engine due to a benign, non-deterministic execution
+            execution.outcome = .failed(0)
+        }
+        else if differentialResult && execution.outcome == .succeeded &&
+           !shouldExecutionBeIgnoredForDifferentialTesting(execution: execution) {
+
+            assert(referenceRunner != nil)
+            assert(referenceRunner!.isInitialized)
+            let diff = referenceRunner!.run(script, withTimeout: timeout ?? config.timeout)
```
saelo, line 499: Should this get a higher timeout since the reference runner is (I guess?) assumed to be slower? E.g. just 2x - 4x more time

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -426,16 +467,47 @@ public class Fuzzer {
...
+        }
 
+        let script = lifter.lift(program)
         dispatchEvent(events.PreExecute, data: program)
-        let execution = runner.run(script, withTimeout: timeout ?? config.timeout)
+        var execution = runner.run(script, withTimeout: timeout ?? config.timeout)
         dispatchEvent(events.PostExecute, data: execution)
 
+        if differentialResult && shouldExecutionBeIgnoredForDifferentialTesting(execution: execution) {
+            // deliberate crash by the engine due to a benign, non-deterministic execution
```
saelo, line 491: I'm not sure I understand this comment. What causes a crash?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -426,16 +467,47 @@ public class Fuzzer {
...
     ///   - timeout: The timeout after which to abort execution. If nil, the default timeout of this fuzzer will be used.
     /// - Returns: An Execution structure representing the execution outcome.
-    public func execute(_ program: Program, withTimeout timeout: UInt32? = nil) -> Execution {
+    public func execute(_ program: Program, withTimeout timeout: UInt32? = nil,
+                        differentialExecute: DifferentialExecute = .auto) -> Execution {
         dispatchPrecondition(condition: .onQueue(queue))
         assert(runner.isInitialized)
 
-        let script = lifter.lift(program)
+        let differentialResult: Bool
```
saelo, line 475: Nit: maybe name this `performDifferentialTesting` or so? That'll make the `if` conditions below a bit more intuitive I think

Sources/Fuzzilli/Configuration.swift
```diff
@@ -23,6 +23,28 @@ public struct Configuration {
...
+    public let differentialTests: [String]
+
+    /// Code snippets that must not cause an observable difference of output
+    /// in the target engine. Used to verify that common sources of
+    /// entropy (Math.random, ...) are deterministic.
+    public let differentialTestsInvariant: [String]
+
+    /// During differential fuzzing, this is the rate of variables probed
+    /// at the end of the generated programs.
+    public let differentialRate: Double
```
saelo, line 37: What's your impression of how important these are? Aren't they basically a subset of the "weaved" hash operations?
I think the name is a bit confusing. Intuitively, I'd for example think that this is the rate of differential tests being performed during fuzzing, i.e. if this is 0.5, half of the generated programs would be for differential testing, the other half would be "regular" samples. Maybe they should be called something like "differentialOutOfLineHashRate" and "differentialInlineHashRate" or something along these lines?

Sources/FuzzilliCli/main.swift
```diff
@@ -303,8 +315,20 @@ for (generator, var weight) in (additionalCodeGenerators + regularCodeGenerators
...
-    let runner = REPRL(executable: jsShellPath, processArguments: profile.getProcessArguments(randomizingArguments), processEnvironment: profile.processEnv, maxExecsBeforeRespawn: profile.maxExecsBeforeRespawn)
+    let runner = REPRL(executable: jsShellPath, processArguments: profile.getProcessArguments(randomizingArguments, differentialTesting), processEnvironment: profile.processEnv, maxExecsBeforeRespawn: profile.maxExecsBeforeRespawn)
+
+    var referenceRunner: ScriptRunner? = nil
+    if differentialTesting {
+        referenceRunner = REPRL(executable: jsShellPath,
+                                processArguments: profile.processArgumentsReference,
+                                processEnvironment: profile.processEnv,
+                                maxExecsBeforeRespawn: profile.maxExecsBeforeRespawn)
+        // we never actually evaluate the coverage no need to attach the
```
saelo, line 328: So why are we creating the evaluator in the first place?

Targets/V8/Patches/v8.patch
```diff
@@ -0,0 +1,188 @@
+diff --git a/BUILD.bazel b/BUILD.bazel
```
saelo, line 1: If you want, feel free to send a CL to V8 to upstream these changes. Otherwise I'll do it after we land this PR.

Sources/Fuzzilli/Execution/Execution.swift
```diff
@@ -59,4 +65,8 @@ public protocol Execution {
 
     /// Execution time in microseconds
     var execTime: TimeInterval { get }
+
+    /// A serialized version of (some of) the programs variables after
```
saelo, line 69: Maybe we should call this `executionHash` or so? I think calling it something with `hash` makes it more clear that this is a hash of a number of values that were produced during the program's execution.

saelo (commented):
Thanks! Overall looks good, mostly nits and some suggestions for improvements. I'm wondering though, what ensures that the new `DifferentialHash` operations don't end up in the Corpus? Should they all be removed prior to/during minimization or so?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -426,16 +467,47 @@ public class Fuzzer {
...
+            // deliberate crash by the engine due to a benign, non-deterministic execution
+            execution.outcome = .failed(0)
+        }
+        else if differentialResult && execution.outcome == .succeeded &&
+           !shouldExecutionBeIgnoredForDifferentialTesting(execution: execution) {
+
+            assert(referenceRunner != nil)
+            assert(referenceRunner!.isInitialized)
+            let diff = referenceRunner!.run(script, withTimeout: timeout ?? config.timeout)
+            dispatchEvent(events.PostExecute, data: diff)
```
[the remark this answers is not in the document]
bernhl: The reason I added dispatching of `PostExecute` is to increase the `totalExecs`, buts its not really necessary. Omitting the event would result in `totalExecs` not including the executions in the reference runner.
Dispatching `PreExecute` is unnecessary work at this would cause a call to `cov_clear_bitmap`.

Sources/Fuzzilli/Configuration.swift
bernhl: I'll rename the variables/parameters as you suggested. Do you think there should be an additional variable controlling the rate of programs which are differentially tested?

Sources/Fuzzilli/Fuzzer.swift
```diff
@@ -418,6 +447,18 @@ public class Fuzzer {
         return corpus.supportsFastStateSynchronization
     }
 
+    private func shouldExecutionBeIgnoredForDifferentialTesting(execution: Execution) -> Bool {
+        return config.differentialPoison.contains { poison in
```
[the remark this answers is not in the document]
bernhl: I updated the code s.t. `shouldExecutionBeIgnoredForDifferentialTesting` is always called because even if a particular program doesn't include a differential test. This is necessary because the engine might be instantiated with command line flags for differential testing and crash because it detected a non-recoverable non-determinism.

Sources/libreprl/libreprl-posix.c
```diff
@@ -410,11 +411,15 @@ int reprl_execute(struct reprl_context* ctx, const char* script, uint64_t script
     }
     
     // Poll succeeded, so there must be something to read now (either the status or EOF).
+    struct {
```
saelo, line 414: So since these changes to the REPRL interface are not backwards compatible, they will break all the existing REPRL implementations for the other engines. I don't think we need to fix them, but it'd be nice if we could at least print a message along the lines of "The target engine uses an old version of the REPRL protocol, please update it to the latest version".

My suggestion would be the following:

1. Add this code at the start of this file (and I guess into the client-side code)
```C
// The REPRL protocol version. Must be identical between server and client.
// Bump this up whenever there are non-backwards-compatible changes to the protocol
#define REPRL_PROTOCOL_VERSION 2
```

2. Change this code: https://github.com/googleprojectzero/fuzzilli/blob/3af52d7f2a58754097b2920b9c9f5ab30dc63d37/Sources/libreprl/libreprl-posix.c#L262 to something like
```C
    char handshake[5] = { 0 };
    if (read(ctx->ctrl_in, handshake, 4) != 4) {
        reprl_terminate_child(ctx);
        return reprl_error(ctx, "Did not receive handshake message from child: %s", strerror(errno));
    }
    
    if (strncmp(handshake, "HELO", 4) == 0) {
        // Client is using protocol version 1.
        reprl_terminate_child(ctx);
        return reprl_error(ctx, "The client uses an old version of the REPRL protocol, please update it to the latest version");
    }

    if (strncmp(handshake, "RPRL", 4) != 0) {
        reprl_terminate_child(ctx);
        return reprl_error(ctx, "Received invalid handshake message from child: %s", helo);
    }

    int client_protocol_version;
    static_assert(sizeof(client_protocol_version) == 4);
    if (read(ctx->ctrl_in, &client_protocol_version, 4) != 4) {
        reprl_terminate_child(ctx);
        return reprl_error(ctx, "Did not receive client protocol version from child: %s", strerror(errno));
    }

    if (client_protocol_version != REPRL_PROTOCOL_VERSION) {
        reprl_terminate_child(ctx);
        return reprl_error(ctx, "The client uses an old version of the REPRL protocol, please update it to the latest version");
    }

    // The server side just sends back the handshake message, but not the protocol version.
    if (write(ctx->ctrl_out, handshake, 4) != 4) {
        reprl_terminate_child(ctx);
        return reprl_error(ctx, "Failed to send handshake reply message to child: %s", strerror(errno));
    }
```

3. Update the client-side code accordingly, i.e. send `"RPRL"` (or some other string) instead of `"HELO"` and send the protocol version number afterwards.

WDYT about this?

PR #389, opened 2023-03-28
add .propagatesSurroundingContext to loop related ILs

birdg0:
I think loop related ILs should inherit context.

### DIFF
--- a/Tests/FuzzilliTests/AnalyzerTest.swift
+++ b/Tests/FuzzilliTests/AnalyzerTest.swift
@@ -71,7 +71,7 @@ class AnalyzerTests: XCTestCase {
             XCTAssertEqual(b.context, [.javascript, .subroutine])
             let loopVar = b.loadInt(0)
             b.buildWhileLoop({
-                XCTAssertEqual(b.context, .javascript)
+                XCTAssertEqual(b.context, [.javascript, .subroutine])
                 return b.compare(loopVar, with: b.loadInt(10), using: .lessThan)
 
             }) {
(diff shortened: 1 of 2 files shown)


saelo:
Hi! Thanks for the PR!
Good point, you can indeed have something like
```
function *foo() {
    while (yield 42) {}
}
```
What doesn't work is `return`, since that'll not be able to return from any outer function. But that's probably fine.
Could you also add one or two tests to LifterTest that show scenarios where this is useful (such as the one with `yield` from above)?

saelo:
Ah yes, that's an excellent example! Could you add a testcase like this to LifterTests.swift:

```
    func testForLoopLifting10() {
        let fuzzer = makeMockFuzzer()
        let b = fuzzer.makeBuilder()

        b.buildPlainFunction(with: .parameters(n: 0)) { _ in
            let s = b.loadInt(0)
            // Test that context-dependent operations such as LoadArguments are handled correctly inside loop headers
            b.buildForLoop(i: { b.loadInt(0) }, { i in b.compare(i, with: b.getProperty("length", of: b.loadArguments()), using: .lessThan) }, { i in b.unary(.PostInc, i) }) { i in
                let arg = b.getComputedProperty(i, of: b.loadArguments())
                b.reassign(s, to: arg, with: .Add)
            }
            b.doReturn(s)
        }

        let program = b.finalize()
        let actual = fuzzer.lifter.lift(program)

        let expected = """
        function f0() {
            let v1 = 0;
            for (let i3 = 0; i3 < arguments.length; i3++) {
                v1 += arguments[i3];
            }
            return v1;
        }

        """

        XCTAssertEqual(actual, expected)
    }
```

? Also please run `swift test`, I think there will now be some test for the context analyser that need to be fixed. Thanks!

birdg0:
Yes, I have fixed AnalyzerTest.

PR #405, opened 2023-12-29
Add SpiderMonkey CVEs to README

bernhl:
Credit Fuzzilli with a couple of issues


saelo:
Wow awesome work, congrats! And thanks for sending the PR! FWIW it looks like these were mostly/all hard crashes, but if some of these bugs were found by JIT Picker, feel free to mark them as such (e.g. with something like `... (found by [JIT-Picker](https://github.com/RUB-SysSec/JIT-Picker))`)

saelo (commented):
Awesome work!

PR #417, opened 2024-02-16
Add missing Array and TypedArray prototype properties to environment

0xedward:
Although ProbingMutator can probably find these properties, I thought it might be a good idea to give CodeGenMutator a chance to also generate code with the following properties:
- https://262.ecma-international.org/14.0/#sec-array.prototype.findlast
- https://262.ecma-international.org/14.0/#sec-array.prototype.findlastindex
- https://262.ecma-international.org/14.0/#sec-array.prototype.toreversed
- https://262.ecma-international.org/14.0/#sec-array.prototype.tosorted
- https://262.ecma-international.org/14.0/#sec-array.prototype.tospliced
- https://262.ecma-international.org/14.0/#sec-array.prototype.with

- https://262.ecma-international.org/14.0/#sec-%typedarray%.prototype.at
- https://262.ecma-international.org/14.0/#sec-%typedarray%.prototype.findindex
- https://262.ecma-international.org/14.0/#sec-%typedarray%.prototype.findlastindex
- https://262.ecma-international.org/14.0/#sec-%typedarray%.prototype.toreversed
- https://262.ecma-international.org/14.0/#sec-%typedarray%.prototype.tosorted

### DIFF
--- a/Sources/Fuzzilli/Environment/JavaScriptEnvironment.swift
+++ b/Sources/Fuzzilli/Environment/JavaScriptEnvironment.swift
@@ -583,7 +583,7 @@ public extension ILType {
 
     /// Type of a JavaScript TypedArray object of the given variant.
     static func jsTypedArray(_ variant: String) -> ILType {
-        return .iterable + .object(ofGroup: variant, withProperties: ["buffer", "byteOffset", "byteLength", "length"], withMethods: ["copyWithin", "fill", "find", "findIndex", "reverse", "slice", "sort", "includes", "indexOf", "keys", "entries", "forEach", "filter", "map", "every", "set", "some", "subarray", "reduce", "reduceRight", "join", "lastIndexOf", "values", "toLocaleString", "toString"])
+        return .iterable + .object(ofGroup: variant, withProperties: ["buffer", "byteOffset", "byteLength", "length"], withMethods: ["at", "copyWithin", "fill", "find", "findIndex", "findLast", "findLastIndex", "reverse", "slice", "sort", "includes", "indexOf", "keys", "entries", "forEach", "filter", "map", "every", "set", "some", "subarray", "reduce", "reduceRight", "join", "lastIndexOf", "values", "toLocaleString", "toString", "toReversed", "toSorted", "with"])
     }
 
     /// Type of a JavaScript function.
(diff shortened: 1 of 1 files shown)


saelo:
Thanks a lot! Just one comment. Maybe just drop the first commit of this chain?

0xedward:
Just to make sure - do you mean this commit, 25efa6e899e5d0908ff3a0767495d302960c9dd5?

If so, I made those changes since it seems like the [`index` parameter for `Array.prototype.at` accepts any type, which will be coerced to integer for the operation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#index). I thought specifying `.anything` for the `index` parameter would result in a larger set of possible inputs for fuzzing than having `index` typed to only accept integers.

I'm happy to make the change :), but I wanted to double check before dropping the commit and also changing the type information for `TypedArray.prototype.at` in d4e15eb3a982075e18eea8074f4d2205415bd283.

saelo (commented):
Ah sorry, I realize my review comment hadn't been released, my bad...

Yeah we should document this approach better. I guess a rule of thumbs would be that if the documentation says something like "parameter will be converted to X" than we want type X in our signature. JavaScript is loosely typed so will usually try to convert whatever it gets to the desired type, but for the majority of possible values, that type conversion will be fairly meaningless (e.g. result in `NaN`) so we don't want to have a ton of those samples. The generative engine (which is the only thing using these signatures) should try to generate correct _and_ meaningful code, and the mutators will take care of e.g. flipping inputs to call functions with unexpected/different values.

0xedward:
> Ah sorry, I realize my review comment hadn't been released, my bad...

No problem, I can relate - I've been burned many times by the draft comment feature lol

> JavaScript is loosely typed so will usually try to convert whatever it gets to the desired type, but for the majority of possible values, that type conversion will be fairly meaningless (e.g. result in NaN) so we don't want to have a ton of those samples. 

> The mutators (in particular, the InputMutator) will take care of changing the input types, and if that's interesting (e.g. increases coverage) we'll still keep it and mutate it further.

Oh I see, that makes sense that we want to generate meaningful inputs otherwise the search space might be too large, and then leave it up to other mutators to mutate the `index` parameter.

Thanks for taking the time to share the rationale behind why it's better to have a more narrow type here :)

> I guess a rule of thumbs would be that if the documentation says something like "parameter will be converted to X" than we want type X in our signature. 

I'll keep that in mind for future PRs

Sources/Fuzzilli/Environment/JavaScriptEnvironment.swift
```diff
@@ -835,6 +837,10 @@ public extension ObjectGroup {
             "flatMap"        : [.function(), .opt(.anything)] => .jsArray,
             "toString"       : [] => .jsString,
             "toLocaleString" : [.opt(.string), .opt(.object())] => .jsString,
+            "toReversed"     : [] => .jsArray,
+            "toSorted"       : [.opt(.function())] => .jsArray,
+            "toSpliced"      : [.integer, .opt(.integer), .anything...] => .jsArray,
+            "with"           : [.anything, .anything] => .jsArray,
```
saelo, line 843: I think the first parameter should also be a `.integer` here since it should be an index IIUC

0xedward: Oh right, I think it should be `.integer`. Sorry I missed that earlier. I also missed that TypedArrays have a `with` method too, so I added that method's type information in the latest change

PR #419, opened 2024-02-22
Allow adding additional ObjectGroups to environment using Profile

0xedward:
As far as I'm aware, currently the only way to add additional ObjectGroups to the JS environment requires maintaining a separate fork of Fuzzilli. I thought allowing users of Fuzzilli to supply additional ObjectGroups in their Profiles would remove the need to maintain a separate fork to have a different JS environment from the main branch of Fuzzilli.

An example of a fork that could've benefited from this change is [fuzzilli4wasm](https://github.com/P1umer/fuzzilli4wasm/blob/5ecdc393366f9c4920eeda2c0bd934d0919a02f5/Sources%2FFuzzilli%2FCore%2FJavaScriptEnvironment.swift#L563-L616)

## Example use case
For Hermes, we want to fuzz the [TextEncoder web API](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) (https://github.com/facebook/hermes/commit/3863a36a53005dd1e6d39ea0d4ef5573bafde910, https://github.com/facebook/hermes/commit/7f9d9d5e44e2df5170f33d1604bea3b085f40e39, https://github.com/facebook/hermes/commit/14790c925cadce5eff7e604d7640d6d5451fb7cc):

```swift
// HermesProfile.swift
import Fuzzilli

let TextEncoderConstructor = ILType.object(ofGroup: "TextEncoder", withProperties: ["encoding"], withMethods: ["encodeInto", "encode"])

let hermesProfile = Profile(

    // ...

    additionalBuiltins: [
        "TextEncoder"           : TextEncoderConstructor,
    ],

    additionalObjectGroups: [
        ObjectGroup(
            name: "TextEncoder",
            instanceType: TextEncoderConstructor,
            properties: [
                "encoding"      : .jsString,
            ],
            methods: [
                "encodeInto"    : [.string] => .jsTypedArray("Uint8Array"),
                "encode"        : [.string, .iterable] => .object(), // Just an example, but .iterable should be a Uint8Array instead
            ]),
    ],

    // ...
)

```

## Test plan
1. Updated `Fuzzilli.JavaScriptEnvironment.groups` and `Fuzzilli.JavaScriptEnvironment.builtinTypes` access control modifiers from `private` to `public`
2. Added print statements to print `environment.groups` and `environment.builtinTypes` in [FuzzilliCli/main.swift](https://github.com/googleprojectzero/fuzzilli/blob/9d56c4b08685e5dd82199eae9ebb68145a435f49/Sources/FuzzilliCli/main.swift#L442)
3. Ran Fuzzilli - `swift run FuzzilliCli --profile=hermes ../fuzzilli_build/bin/fuzzilli --storagePath=./corpus --logLevel=verbose`
4. Checked that `TextEncoder` is in the output for `environment.groups` and `environment.builtinTypes`


Sources/FuzzilliCli/main.swift
```diff
@@ -438,7 +438,13 @@ func makeFuzzer(with configuration: Configuration) -> Fuzzer {
     }
 
     // The environment containing available builtins, property names, and method names.
-    let environment = JavaScriptEnvironment(additionalBuiltins: profile.additionalBuiltins, additionalObjectGroups: [])
+    let environment = JavaScriptEnvironment(additionalBuiltins: profile.additionalBuiltins, additionalObjectGroups: profile.additionalObjectGroups)
+    if !profile.additionalBuiltins.isEmpty {
+        logger.verbose("Loaded additional builtins from profile: \(profile.additionalBuiltins.map { $0.key })")
+    }
+    if !profile.additionalObjectGroups.isEmpty {
+        logger.verbose("Loaded additional ObjectGroups from profile: \(profile.additionalObjectGroups.map { $0.name })")
```
0xedward, line 446:
```suggestion
        logger.info("Loaded additional builtins from profile: \(profile.additionalBuiltins.map { $0.key })")
    }
    if !profile.additionalObjectGroups.isEmpty {
        logger.info("Loaded additional ObjectGroups from profile: \(profile.additionalObjectGroups.map { $0.name })")
```

What are your thoughts on using `logger.info` instead `logger.verbose` here?

carl-smith: I don't think this logging is necessary as we will also print all registered builtins in the `initialize` function in the [JavaScriptEnvironment](https://github.com/googleprojectzero/fuzzilli/blob/5f268e1a64f474fac402c79b4f83f7f8c8eeb24a/Sources/Fuzzilli/Environment/JavaScriptEnvironment.swift#L432C9-L433C84). So if you want to keep this for visibility then we should use `.verbose`, otherwise I think we could also drop this, your choice :)

carl-smith (commented):
Nice!

Sources/FuzzilliCli/main.swift
0xedward: Ah, I noticed that `initialize` in JavaScriptEnvironment prints all the registered builtins and objectGroups. Since we fuzz using quite a few Fuzzilli profiles, I think we'd find it helpful to have which of the objectGroups and builtins were loaded from the profile without having to search through the full list of objectGroups and builtins in our logs

0xedward:
Thanks for the review, Carl! :)

PR #425, opened 2024-03-22
fixed some typos

tin-z:
fixed some typos


saelo:
Thanks! And sorry for the slow reply... :(

PR #427, opened 2024-03-29
XS code generators, object groups, etc.

phoddie:
This is the profile we use for XS with Fuzzilli. It includes code generators to stress unique aspects of the XS runtime. It has two ObjectGroups and several additional build-ins to fuzz Hardened JavaScript features. It also pulls in the `RegExpFuzzer` from V8.


phoddie:
Oops.... failed the CLA check because of wrong email. Closing.....

PR #428, opened 2024-03-29
XS code generators, object groups, etc

phoddie:
This is the profile we use for XS with Fuzzilli. It includes code generators to stress unique aspects of the XS runtime. It has two ObjectGroups and several additional build-ins to fuzz Hardened JavaScript features. It also pulls in the RegExpFuzzerfrom V8.

### DIFF
--- a/Sources/FuzzilliCli/Profiles/XSProfile.swift
+++ b/Sources/FuzzilliCli/Profiles/XSProfile.swift
@@ -14,12 +14,306 @@
 
 import Fuzzilli
 
+// swift run FuzzilliCli --profile=xs --jobs=8 --storagePath=./results --resume --inspect=history --timeout=100 $MODDABLE/build/bin/mac/debug/xst
+// swift run -c release FuzzilliCli --profile=xs --jobs=8 --storagePath=./results --resume --timeout=200 $MODDABLE/build/bin/mac/debug/xst
+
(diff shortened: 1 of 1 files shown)


Sources/FuzzilliCli/Profiles/XSProfile.swift
```diff
@@ -44,21 +338,38 @@ let xsProfile = Profile(
...
     disabledMutators: [],
 
     additionalBuiltins: [
         "gc"                  : .function([] => .undefined),
+        "memoryFail"          : .function([.number] => .number),
         "print"               : .function([.string] => .undefined),
-        "placeholder"         : .function([] => .undefined),
+
+        // hardened javascript
+        "Compartment"         : .function([] => .jsCompartmentConstructor),
```
carl-smith, line 364: Same here as with the signatures in the object groups, please provide the types that these constructors expect.

phoddie:
@carl-smith, thanks for the thorough review. I'll try to get all that cleaned up.

FWIW – XSProfile is an admittedly a bit messy. It has evolved over time while I've been learning a (little) bit about Swift and Fuzzilli APIs along the way. Despite the imperfections, it has been incredibly valuable in finding subtle issues in XS. Thanks so much for your work on this project.

phoddie:
Closing. Replaced by #431.

PR #430, opened 2024-04-03
Add support for running on macOS with the target jsc build

Eziodeng:
Cheak whether this cannot run on macOS and restore the fuzzbuild shell for macOS build which use the build-jsc shell without not officially supported "--jsc-only" port

Also fix the REPRL and main fuzzer code to deal with the jsc executable path input which lead to the executeable file cannot run on macOS for the env not properly set

The problem is that on macOS with the test tag I have used is "releases/Apple/Safari-14.0.3-macOS-11.2.3", and on the ubuntu it has coverage edges almost 960000+ and only 12000+ on mac.（Here I have found that this problem is caused the JavascriptCore.framework build, I now try to change it like WTF to build a .a library）

### DIFF
--- a/Sources/Fuzzilli/Execution/REPRL.swift
+++ b/Sources/Fuzzilli/Execution/REPRL.swift
@@ -49,6 +49,14 @@ public class REPRL: ComponentBase, ScriptRunner {
         self.maxExecsBeforeRespawn = maxExecsBeforeRespawn
         super.init(name: "REPRL")
 
+        #if os(macOS)
+            // Make sure that the DYLD_FRAMEWORK_PATH have been correctly set
+            // Only when the name of executable is jsc
+            if executable.suffix(4) == "/jsc" {
+                env.append("DYLD_FRAMEWORK_PATH=" + executable.components(separatedBy: "jsc")[0])
+            }
+        #endif
+
         for (key, value) in processEnvironment {
             env.append(key + "=" + value)
         }
--- a/Sources/Fuzzilli/Modules/NetworkSync.swift
+++ b/Sources/Fuzzilli/Modules/NetworkSync.swift
@@ -48,7 +48,7 @@ fileprivate protocol MessageHandler {
 /// A connection to a network peer that speaks the above protocol.
 fileprivate class Connection {
     /// The file descriptor on POSIX or SOCKET handle on Windows of the socket.
-    let socket: libsocket.socket_t
+    let socket: libsocket.socket_fd
 
     /// The UUID of the remote end.
     let localId: UUID
@@ -78,7 +78,7 @@ fileprivate class Connection {
     /// Pending outgoing data. Must only be accessed on this connection's dispatch queue.
     private var sendQueue: [Data] = []
 
-    init?(socket: libsocket.socket_t, localId: UUID, handler: MessageHandler) {
+    init?(socket: libsocket.socket_fd, localId: UUID, handler: MessageHandler) {
         self.socket = socket
         self.localId = localId
         self.handler = handler
@@ -331,7 +331,7 @@ public class NetworkParent: DistributedFuzzingParentNode {
         unowned let fuzzer: Fuzzer
 
         /// File descriptor or SOCKET handle of the server socket.
-        private var serverFd: libsocket.socket_t = INVALID_SOCKET
+        private var serverFd: libsocket.socket_fd = INVALID_SOCKET
 
         /// Address and port on which to listen for connections.
         private let address: String
@@ -346,7 +346,7 @@ public class NetworkParent: DistributedFuzzingParentNode {
         private var serverQueue: DispatchQueue? = nil
 
         /// Active workers indexed by the socket used to communicate with them.
-        private var clientsBySocket = [libsocket.socket_t: Client]()
+        private var clientsBySocket = [libsocket.socket_fd: Client]()
 
         /// Active workers indexed by their id.
         private var clientsById = [UUID: Client]()
@@ -445,7 +445,7 @@ public class NetworkParent: DistributedFuzzingParentNode {
             onChildDisconnectedCallback = callback
         }
 
-        private func handleNewConnection(_ socket: libsocket.socket_t) {
+        private func handleNewConnection(_ socket: libsocket.socket_fd) {
             guard socket > 0 else {
                 return logger.error("Failed to accept client connection")
             }
--- a/Sources/FuzzilliCli/Profiles/JSCProfile.swift
+++ b/Sources/FuzzilliCli/Profiles/JSCProfile.swift
@@ -49,8 +49,12 @@ let jscProfile = Profile(
             "--thresholdForFTLOptimizeAfterWarmUp=1000",
             "--thresholdForFTLOptimizeSoon=1000",
             // Enable bounds check elimination validation
-            "--validateBCE=true",
+            //"--validateBCE=true",
             "--reprl"]
+        
+        #if !os(macOS)
+        args.append("--validateBCE=true")
+        #endif
 
         guard randomize else { return args }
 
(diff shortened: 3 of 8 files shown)


Eziodeng:
Maybe I should first try to solve the static library of JavascriptCore.
Close this PR.

Eziodeng:
Successfully build a static library version with some mentally settings.
Now try running for a test
Reopen it.

Sources/Fuzzilli/Execution/REPRL.swift
```diff
@@ -49,6 +49,14 @@ public class REPRL: ComponentBase, ScriptRunner {
         self.maxExecsBeforeRespawn = maxExecsBeforeRespawn
         super.init(name: "REPRL")
 
+        #if os(macOS)
```
saelo, line 52: This is a bit of an unfortunate workaround. If there's no way to avoid this with JSC build options, then I think it should go into the [JSCProfile](https://github.com/googleprojectzero/fuzzilli/blob/main/Sources/FuzzilliCli/Profiles/JSCProfile.swift)

Sources/Fuzzilli/Modules/NetworkSync.swift
```diff
@@ -48,7 +48,7 @@ fileprivate protocol MessageHandler {
 /// A connection to a network peer that speaks the above protocol.
 fileprivate class Connection {
     /// The file descriptor on POSIX or SOCKET handle on Windows of the socket.
-    let socket: libsocket.socket_t
+    let socket: libsocket.socket_fd
```
saelo, line 51: Not necessarily opposed to this rename, but why is it necessary?

Sources/FuzzilliCli/Profiles/JSCProfile.swift
```diff
@@ -49,8 +49,12 @@ let jscProfile = Profile(
             "--thresholdForFTLOptimizeAfterWarmUp=1000",
             "--thresholdForFTLOptimizeSoon=1000",
             // Enable bounds check elimination validation
-            "--validateBCE=true",
+            //"--validateBCE=true",
```
saelo, line 52: What's the reason for this change? Is that feature only available on linux? I'm not sure how important it even is, maybe we can drop it entirely

Targets/JavaScriptCore/fuzzbuild.sh
```diff
@@ -18,6 +18,10 @@ export WEBKIT_OUTPUTDIR=FuzzBuild
 
 if [ "$(uname)" == "Linux" ]; then
     ./Tools/Scripts/build-jsc --jsc-only --debug --cmakeargs="-DENABLE_STATIC_JSC=ON -DCMAKE_C_COMPILER='/usr/bin/clang' -DCMAKE_CXX_COMPILER='/usr/bin/clang++' -DCMAKE_CXX_FLAGS='-fsanitize-coverage=trace-pc-guard -O3 -lrt'"
+elif [ "$(uname)" == "Darwin" ]; then
```
saelo, line 21: While you're here, is our [jsc patch](https://github.com/googleprojectzero/fuzzilli/tree/main/Targets/JavaScriptCore/Patches) still necessary? I believe JSC now has native support for Fuzzilli, so maybe we can drop that patch entirely? In that case, please also update the [README](https://github.com/googleprojectzero/fuzzilli/blob/main/Targets/JavaScriptCore/README.md)

Sources/REPRLRun/main.swift
```diff
@@ -37,7 +37,12 @@ if ctx == nil {
 }
 
 let argv = convertToCArray(Array(CommandLine.arguments[1...]))
+#if os(macOS)
+// Set the DYLD_FRAMEWORK_PATH env variable for the shell to run
```
saelo, line 41: This also feels misplaced here. Maybe we could just forward our environment to the child instead? Then you'd have to `DYLD_FRAMEWORK_PATH=xyz ./REPRLRun path/to/jsc` which would be ok I guess

Targets/JavaScriptCore/fuzzbuild.sh
```diff
@@ -18,6 +18,10 @@ export WEBKIT_OUTPUTDIR=FuzzBuild
...
+    unset WEBKIT_OUTPUTDIR # remove the outputdir env on macOS for the build without cmake will be in the Source directory(maybe the build-jsc have problem)
```
saelo, line 22: I'm not sure I understand this. So if that environment variable exist, the build artifacts will be placed in Source/? Or the other way around?

saelo (commented):
Thanks! A couple of questions and suggestions

Sources/Fuzzilli/Execution/REPRL.swift
Eziodeng: Yes, truly it should .
(With the next problem, I find

Sources/Fuzzilli/Modules/NetworkSync.swift
Eziodeng: Either old or new version of macOS with the build of the swift just have encountered the error with bellow 
```shell
fuzzilli/Sources/libsocket/include/libsocket.h:39:13: note: candidate found by name lookup is 'socket_t'
typedef int socket_t;
            ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/include/sys/kernel_types.h:96:41: note: candidate found by name lookup is 'socket_t'
typedef struct __socket*                socket_t;
```
Maybe there is no problem on Linux. So I suggest avoid the `reference to ' socket_t' is ambiguous` error

Sources/FuzzilliCli/Profiles/JSCProfile.swift
Eziodeng: I guess this just do not support for some old version, so just take care.
Here with the comment of `// Enable bounds check elimination validation`, I think there should be some function for that? So I donot handle with that.
Or just check the macOS sdk version to validate if this can be used for the build version of jsc? (For purpose of the old version? maybe this is not necessary)

Targets/JavaScriptCore/fuzzbuild.sh
Eziodeng: I dont find that until you tell me. I just follow with the old one.   XD 
Yes we should just update the README for the workflow of the build and use for fuzzilli

But for the settings on Darwin, I think we still need to keep this shell with some args be set by `set-webkit-configuration`

Sources/REPRLRun/main.swift
Eziodeng: Yes it should be fine, but I just find that with the static build of JavascriptCore of libJavascriptCore.a, we now do not need the DYLD_FRAMEWORK_PATH for the shell to run, so we can discard this commit at all if we properly solve the static build (Indeed a mess especially for some version of WebKit code fro some code between release tag). 

Also for the first problem

Targets/JavaScriptCore/fuzzbuild.sh
Eziodeng: In the past, I have just set the sub project to build with xcode settings that reletive to source folder directly within opening the JavascriptCore proj

With the debug of my env, maybe my wrong settings for the sub xcode project make the problem, I will try with that again and check what the workflow of build with the perl script and Makefile.

Try with the export WEBKIT_OUTPUTDIR to FuzzBuild.
And the build script just lead to this pic which means if this env exist, the artifacts will just in the `Source/` folder
I think this should be the problem caused by the Makefile
image

Eziodeng:
I now need some time to check if there is some difference between the static build on macOS  and the Linux one

Eziodeng:
Working for other things now and take it off until free for sometime

PR #432, opened 2024-04-16
Add SpiderMonkey CVEs to README

bernhl:
Bugs mentioned here: https://www.mozilla.org/en-US/security/advisories/mfsa2024-18/


saelo:
Awesome! Congrats!

Issue #437, opened 2024-07-22
Improve FuzzIL Compiler


saelo:
The FuzzIL compiler is not yet feature complete as it does not support a number of JavaScript features. For most features, it should be enough to add the necessary support in the [parser](https://github.com/googleprojectzero/fuzzilli/blob/main/Sources/Fuzzilli/Compiler/Parser/parser.js) and [compiler](https://github.com/googleprojectzero/fuzzilli/blob/main/Sources/Fuzzilli/Compiler/Compiler.swift), which should be mostly straight-forward. However, some features may require changes to the FuzzIL itself, in case the respective language feature is not yet implemented in the IL.

To identify missing features, the following script can be used with an existing corpus of .js files.
```
import subprocess                              
import glob
import sys                                            
import re
import multiprocessing                       
from dataclasses import dataclass
                                                                                           
if len(sys.argv) != 3:
    print("Usage: compile.py path/to/FuzzILTool path/to/javascript_corpus_directory")
    sys.exit(-1)
                                                                                                                                                                                      
fuzzilTool = sys.argv[1]  
corpusDirectory = sys.argv[2]

def parseOutput(output):
    m = re.search("Unhandled node type (\w+)", output)
    if m:
        return f"Unhandled node {m.group(1)}"

    m = re.search("SyntaxError:", output)
    if m:
        return f"Syntax Error"

    m = re.search("Expected variable declaration as init part of a for-in loop", output)
    if m:
        return f"Invalid for-in loop"

    m = re.search("Unknown property key type: (\w+)", output)
    if m:
        return f"Unknown property key type: {m.group(1)}"

    m = re.search("Unsupported class declaration field: (\w+)", output)
    if m:
        return f"Unsupported class declaration field: {m.group(1)}"

    m = re.search("Assertion failed", output)
    if m:
        return f"Unknown assertion failure"

    m = re.search("Maximum call stack size exceeded", output)
    if m:
        return f"Stack overflow during parsing"

    m = re.search("Failed to parse .*: (\w+)", output)
    if m:
        return f"Other failure: {m.group(1)}"

    return output

def compileToFuzzIL(path):
    try:
        result = subprocess.run([fuzzilTool, "--compile", path], timeout=10, capture_output=True)
        print(".", end='')
        sys.stdout.flush()
        output = result.stdout.decode('utf-8')
        output = parseOutput(output).strip()
        return (result.returncode, output, path)
    except subprocess.TimeoutExpired as e:
[...]

PR #439, opened 2024-07-26
Feature/with statement js compiler

TobiasWienand:
Addresses #437 
Note to reviewer: https://github.com/TobiasWienand/fuzzilli/blob/7ac6ae5fc777ed469413682640f33a3f7d9a40a3/Sources/Fuzzilli/Compiler/Compiler.swift#L487
needs attention. Is this the correct way to iterate over the block of the with-statement?

### DIFF
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -479,6 +479,16 @@ public class JavaScriptCompiler {
             let value = try compileExpression(throwStatement.argument)
             emit(ThrowException(), withInputs: [value])
 
+        case .withStatement(let withStatement):
+            let object = try compileExpression(withStatement.object)
+            let beginWith = BeginWith()
+            emit(beginWith, withInputs: [object])
+            try enterNewScope {
+                for statement in withStatement.body.blockStatement.body {
+                    try compileStatement(statement)
+                }
+            }
+            emit(EndWith())
         }
     }
 
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -483,9 +483,7 @@ public class JavaScriptCompiler {
             let object = try compileExpression(withStatement.object)
             emit(BeginWith(), withInputs: [object])
             try enterNewScope {
-                for statement in withStatement.body.blockStatement.body {
-                    try compileStatement(statement)
-                }
+                try compileBody(withStatement.body)
             }
             emit(EndWith())
         }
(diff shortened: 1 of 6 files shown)


Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -479,6 +479,16 @@ public class JavaScriptCompiler {
             let value = try compileExpression(throwStatement.argument)
             emit(ThrowException(), withInputs: [value])
 
+        case .withStatement(let withStatement):
+            let object = try compileExpression(withStatement.object)
+            let beginWith = BeginWith()
+            emit(beginWith, withInputs: [object])
```
saelo, line 485: Nit: maybe combine into one line: `emit(BeginWith(), ...)`

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -479,6 +479,16 @@ public class JavaScriptCompiler {
...
+            try enterNewScope {
+                for statement in withStatement.body.blockStatement.body {
```
saelo, line 487: I think instead of this, you can just do `try compileBody(withStatement.body)`. Does that work?

saelo (commented):
Nice, thanks!

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: That's what I tried initially, however VSCode says: 

> For-in loop requires 'Compiler_Protobuf_Statement' to conform to 'Sequence'

Below is an example js code using the with-operator and the AST generated from the JS parser
# basic.js
```
var person = {
    firstName: "John",
  };
  
with (person) {
    console.log(firstName);
  }
  
```
# AST of basic.js
```
Fuzzilli.Compiler_Protobuf_AST:
statements {
  variableDeclaration {
    declarations {
      name: "person"
      value {
        objectExpression {
          fields {
            property {
              name: "firstName"
              value {
                stringLiteral {
                  value: "John"
                }
              }
            }
          }
        }
      }
    }
  }
}
statements {
  withStatement {
    object {
      identifier {
        name: "person"
      }
    }
    body {
      blockStatement {
        body {
          expressionStatement {
            expression {
              callExpression {
                callee {
                  memberExpression {
                    object {
                      identifier {
                        name: "console"
                      }
                    }
                    name: "log"
                  }
                }
                arguments {
                  identifier {
                    name: "firstName"
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

saelo: Hm which for-in loop does it complain about?
What I mean is something like
```swift
case .withStatement(let withStatement):
    let object = try compileExpression(withStatement.object)
    emit(BeginWith(), withInputs: [object])
    try enterNewScope {
        try compileBody(withStatement.body)
    }
    emit(EndWith())
```
`compileBody` requires only a `StatementNode`, so I think that should work there.

TobiasWienand: I see. That works 👍

saelo:
Thanks!

PR #442, opened 2024-08-23
Implements robust break context identification and tests

TobiasWienand:
The two tests succeeded for me. Are any other tests for break context identification interesting?

### DIFF
--- a/Sources/Fuzzilli/FuzzIL/Analyzer.swift
+++ b/Sources/Fuzzilli/FuzzIL/Analyzer.swift
@@ -169,7 +169,8 @@ struct ContextAnalyzer: Analyzer {
                 newContext.formUnion(contextStack.secondToTop)
             }
             
-            // If we are in a loop, we don't want to propagate the switch context and vice versa.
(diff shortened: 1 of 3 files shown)


Sources/Fuzzilli/FuzzIL/Analyzer.swift
```diff
@@ -168,6 +168,14 @@ struct ContextAnalyzer: Analyzer {
 
                 newContext.formUnion(contextStack.secondToTop)
             }
+            
+            // If we are in a loop, we don't want to propagate the switch context and vice versa.
```
saelo, line 172: Can you add something like
`Otherwise, we could determine which type of break operations to emit. TODO consider building a generic mechanism for this if we need similar logic in other cases as well. For example, we could potentially add a instr.op.contextClosed list to generalize this behavior.`

saelo (commented):
Thanks! Also please remove this comment now: https://github.com/googleprojectzero/fuzzilli/blob/aeed429d56bd771ffa5a315f57f3413fa4732b55/Sources/Fuzzilli/Compiler/Compiler.swift#L449 :)

PR #447, opened 2024-09-22
Feature/array spread

TobiasWienand:
Implements compilation of arrays with spread operator and implements various tests

### DIFF
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -833,17 +833,31 @@ public class JavaScriptCompiler {
         case .arrayExpression(let arrayExpression):
             var elements = [Variable]()
             var undefined: Variable? = nil
+            var spreads = [Bool]()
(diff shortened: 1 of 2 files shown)


Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -833,17 +833,31 @@ public class JavaScriptCompiler {
...
+                    else {
+                        elements.append(try compileExpression(elem))
+                        spreads.append(false)
+                    }
                 }
             }
-            return emit(CreateArray(numInitialValues: elements.count), withInputs: elements).output
+            if spreads.contains(true) {
+                return emit(CreateArrayWithSpread(spreads: spreads), withInputs: elements).output
+           }
```
saelo, line 857: Nit: can the `else` go onto this line? Also I think the indention is off a bit, the `if` seems to be more intended than the `else`

saelo (commented):
Cool, looks good, just one nit!

saelo:
Thanks!

PR #449, opened 2024-10-09
Implements compilation of new parameter types

TobiasWienand:
This initial commit enables compilation of objectPattern and arrayPattern parameters

### DIFF
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -468,7 +468,14 @@ public class JavaScriptCompiler {
                 try enterNewScope {
                     let beginCatch = emit(BeginCatch())
                     if tryStatement.catch.hasParameter {
-                        map(tryStatement.catch.parameter.name, to: beginCatch.innerOutput)
+                        let parameter = tryStatement.catch.parameter
+                        switch parameter.parameter {
+                            case .identifierParameter(let identifier):
+                                map(identifier.name, to: beginCatch.innerOutput)
+
+                            default:
+                                throw CompilerError.unsupportedFeatureError("Unsupported parameter type")
+                        }
                     }
                     for statement in tryStatement.catch.body {
                         try compileStatement(statement)
@@ -1071,14 +1078,69 @@ public class JavaScriptCompiler {
     }
 
     private func mapParameters(_ parameters: [Compiler_Protobuf_Parameter], to variables: ArraySlice<Variable>) {
-        assert(parameters.count == variables.count)
-        for (param, v) in zip(parameters, variables) {
-            map(param.name, to: v)
+        var flatParameters: [String] = [] 
+        var expectedVariableCount = 0
+        for param in parameters {
+            switch param.parameter {
+            case .identifierParameter(let identifier):
+                flatParameters.append(identifier.name)
+                expectedVariableCount += 1
+            case .objectParameter(let object):
+                for subParam in object.parameters {
+                    flatParameters.append(subParam.name)
+                    expectedVariableCount += 1
+                }
+            case .arrayParameter(let array):
+                for element in array.elements {
+                    flatParameters.append(element.name)
+                    expectedVariableCount += 1
+                }
+            default:
+                break
+            }
+        }
+        assert(expectedVariableCount == variables.count, "The number of variables does not match the number of parameters.")
+        for (name, v) in zip(flatParameters, variables) {
+            map(name, to: v)
         }
     }
 
     private func convertParameters(_ parameters: [Compiler_Protobuf_Parameter]) -> Parameters {
-        return Parameters(count: parameters.count)
+        var totalParameterCount = 0
+        var parameterTypes = [Parameters.ParameterType]()
+        var objectPropertyNames = [[String]]()
+        
+        for param in parameters {
+            switch param.parameter {
+            case .identifierParameter(_):
+                totalParameterCount += 1
+                parameterTypes.append(.identifier)
+            case .objectParameter(let object):
+                let objectCount = object.parameters.count
+                totalParameterCount += objectCount
+                if (objectCount == 1) {
+                    parameterTypes.append(.standaloneObject)
+                } else {
+                    parameterTypes.append(.objectStart)
+                    parameterTypes.append(contentsOf: Array(repeating: .objectMiddle, count: max(0, objectCount - 2)))
+                    parameterTypes.append(.objectEnd)
+                }
+                objectPropertyNames.append(object.parameters.map { $0.name })
+            case .arrayParameter(let array):
+                let arrayCount = array.elements.count
+                totalParameterCount += arrayCount
+                if arrayCount == 1 {
+                    parameterTypes.append(.standaloneArray)
+                } else {
+                    parameterTypes.append(.arrayStart)
+                    parameterTypes.append(contentsOf: Array(repeating: .arrayMiddle, count: max(0, arrayCount - 2)))
+                    parameterTypes.append(.arrayEnd)
+                }
+            default:
+                break
+            }
+        }
+        return Parameters(count: totalParameterCount, parameterTypes: parameterTypes, objectPropertyNames: objectPropertyNames)
     }
 
     /// Convenience accessor for the currently active scope.
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -468,14 +468,10 @@ public class JavaScriptCompiler {
                 try enterNewScope {
                     let beginCatch = emit(BeginCatch())
                     if tryStatement.catch.hasParameter {
-                        let parameter = tryStatement.catch.parameter
-                        switch parameter.parameter {
-                            case .identifierParameter(let identifier):
-                                map(identifier.name, to: beginCatch.innerOutput)
-
-                            default:
-                                throw CompilerError.unsupportedFeatureError("Unsupported parameter type")
+                        guard case let .identifierParameter(identifier) = tryStatement.catch.parameter.parameter else {
+                            throw CompilerError.unsupportedFeatureError("Only identifier parameters are supported in catch blocks")
                         }
+                        map(identifier.name, to: beginCatch.innerOutput)
                     }
                     for statement in tryStatement.catch.body {
                         try compileStatement(statement)
@@ -1095,7 +1091,7 @@ public class JavaScriptCompiler {
                     flatParameters.append(element.name)
                     expectedVariableCount += 1
                 }
-            default:
+            case .none:
                 break
             }
         }
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -122,7 +122,7 @@ public class JavaScriptCompiler {
             }
 
         case .functionDeclaration(let functionDeclaration):
-            let parameters = convertParameters(functionDeclaration.parameters)
+            let parameters = try convertParameters(functionDeclaration.parameters)
             let functionBegin, functionEnd: Operation
             switch functionDeclaration.type {
             case .plain:
@@ -218,7 +218,7 @@ public class JavaScriptCompiler {
                     emit(op, withInputs: inputs)
 
                 case .ctor(let constructor):
-                    let parameters = convertParameters(constructor.parameters)
+                    let parameters = try convertParameters(constructor.parameters)
                     let head = emit(BeginClassConstructor(parameters: parameters))
 
                     try enterNewScope {
@@ -233,7 +233,7 @@ public class JavaScriptCompiler {
                     emit(EndClassConstructor())
 
                 case .method(let method):
-                    let parameters = convertParameters(method.parameters)
+                    let parameters = try convertParameters(method.parameters)
                     let head: Instruction
                     if method.isStatic {
                         head = emit(BeginClassStaticMethod(methodName: method.name, parameters: parameters))
@@ -780,7 +780,7 @@ public class JavaScriptCompiler {
                         emit(ObjectLiteralAddComputedProperty(), withInputs: [computedPropertyKeys.removeLast()] + inputs)
                     }
                 case .method(let method):
-                    let parameters = convertParameters(method.parameters)
+                    let parameters = try convertParameters(method.parameters)
 
                     let instr: Instruction
                     if case .name(let name) = method.key {
@@ -861,7 +861,7 @@ public class JavaScriptCompiler {
             }
 
         case .functionExpression(let functionExpression):
-            let parameters = convertParameters(functionExpression.parameters)
+            let parameters = try convertParameters(functionExpression.parameters)
             let functionBegin, functionEnd: Operation
             switch functionExpression.type {
             case .plain:
@@ -892,7 +892,7 @@ public class JavaScriptCompiler {
             return instr.output
 
         case .arrowFunctionExpression(let arrowFunction):
-            let parameters = convertParameters(arrowFunction.parameters)
+            let parameters = try convertParameters(arrowFunction.parameters)
             let functionBegin, functionEnd: Operation
             switch arrowFunction.type {
             case .plain:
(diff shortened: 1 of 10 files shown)


brookate:
Can you also support such code: 
1. const [a0, a1] = f0(t);
2. const { prop1 } = obj.f();

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -468,7 +468,14 @@ public class JavaScriptCompiler {
...
                     let beginCatch = emit(BeginCatch())
                     if tryStatement.catch.hasParameter {
-                        map(tryStatement.catch.parameter.name, to: beginCatch.innerOutput)
+                        let parameter = tryStatement.catch.parameter
+                        switch parameter.parameter {
+                            case .identifierParameter(let identifier):
+                                map(identifier.name, to: beginCatch.innerOutput)
+
+                            default:
+                                throw CompilerError.unsupportedFeatureError("Unsupported parameter type")
```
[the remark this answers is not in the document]
TobiasWienand: Fixed with 6e8fde4f29a6aac7caa92e8f4674b069ff3f5956

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -468,7 +468,14 @@ public class JavaScriptCompiler {
                 try enterNewScope {
                     let beginCatch = emit(BeginCatch())
                     if tryStatement.catch.hasParameter {
-                        map(tryStatement.catch.parameter.name, to: beginCatch.innerOutput)
+                        let parameter = tryStatement.catch.parameter
+                        switch parameter.parameter {
```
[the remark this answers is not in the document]
TobiasWienand: Fixed with 6e8fde4f29a6aac7caa92e8f4674b069ff3f5956. Theoretically we could allow other parameter types than IdentifierParameter but practically this would mean making changes to the core of Fuzzilli for a case that will happens very rarely. Thus, I have changed this to a reduced version that only allows the case with IdentifierParameter

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1059,9 +1059,25 @@ public struct Parameters {
...
+    enum ParameterType {
+        case identifier
+        case objectStart
+        case objectEnd
+        case objectMiddle
+        case standaloneObject
+        case arrayStart
+        case arrayEnd
+        case arrayMiddle
+        case standaloneArray
```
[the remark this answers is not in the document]
TobiasWienand: I have implemented a new structure that comes with a bit of overhead in terms of memory but should provide good maintanability / robustness in comparison with a more lightweight approach, such as enums or maybe saving indices for object/array braces. 

The new implementation also allows to compile nested arrays/object which was not possible with the old approach

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1059,9 +1059,25 @@ public struct Parameters {
...
+    }
+    
+    var parameterTypes: [ParameterType]
+    var objectPropertyNames: [[String]]
+    init(count: Int, hasRestParameter: Bool = false, parameterTypes: [ParameterType] = [], objectPropertyNames: [[String]] = []) {
         self.numParameters = UInt32(count)
         self.hasRestParameter = hasRestParameter
+        self.parameterTypes = parameterTypes
```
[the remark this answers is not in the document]
TobiasWienand: I think the new structure guarantees correctness because it mimics the parameter structure. If you don't agree - I'll implement a test. No problem

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1059,9 +1059,25 @@ public struct Parameters {
         return Int(numParameters)
     }
 
-    init(count: Int, hasRestParameter: Bool = false) {
+    enum ParameterType {
```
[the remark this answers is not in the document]
TobiasWienand: Regarding this in particular - I would appreciate if you look over my solution

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -1071,14 +1074,68 @@ public class JavaScriptCompiler {
...
+            case .objectParameter(let object):
+                for property in object.parameters {
+                    extractIdentifiers(from: property.parameterValue)
+                }
+            case .arrayParameter(let array):
+                for element in array.elements {
+                    extractIdentifiers(from: element)
+                }
+            case .none:
+                break
```
saelo, line 1091: Do we ever legitimately expect `.none` for a parameter? I guess this value is autogenerated by the protobuf compiler and there's probably no way to avoid that? But maybe we should just `fatalError` in all these cases or something like that?

Sources/Fuzzilli/FuzzIL/JSTyper.swift
```diff
@@ -170,7 +170,36 @@ public struct JSTyper: Analyzer {
...
+    }
+
+    private func inferParameterType(from pattern: ParameterPattern) -> ILType {
+        switch pattern {
+        case .identifier:
+            return .anything
+        case .array:
+            return .iterable
+        case .object:
+            return .object()
```
saelo, line 201: Maybe leave a TODO here that we could be more precise (IIUC), e.g. by adding properties with the parameter names (e.g. `.object(withProperties: ["a"])`, etc.)

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1047,21 +1047,37 @@ final class TestIn: JsOperation {
 
 }
 
+enum ParameterPattern {
+    case identifier
+    case object(properties: [ObjectPatternProperty])
```
saelo, line 1052: Could this also be a `[String: ParameterPattern]` dictionary or would that make things more complicated? (asking because that "feels" a bit more natural without having thought about it deeply)

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1047,21 +1047,37 @@ final class TestIn: JsOperation {
...
-    /// Whether the last parameter is a rest parameter.
     let hasRestParameter: Bool
 
-    /// The total number of parameters. This is equivalent to the number of inner outputs produced from the parameters.
     var count: Int {
         return Int(numParameters)
     }
 
-    init(count: Int, hasRestParameter: Bool = false) {
+    var objectPropertyNames: [[String]]
```
saelo, line 1070: Is this still used?

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1047,21 +1047,37 @@ final class TestIn: JsOperation {
 
 }
 
+enum ParameterPattern {
```
saelo, line 1050: I think it would also be good to add support to our code generation machinery to generate functions with more complex parameters. The place for that would be here: https://github.com/googleprojectzero/fuzzilli/blob/7b2850d6831535aedba74da7f8811dda6ee59abd/Sources/Fuzzilli/Base/ProgramBuilder.swift#L425 Maybe this makes most sense as a follow-up change though, so we can land this one first to get the compiler support and then see if it's easy to also get codegen support.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1047,21 +1047,37 @@ final class TestIn: JsOperation {
...
+
+struct ObjectPatternProperty {
+    let key: String
+    let value: ParameterPattern
+}
+
 // The parameters of a FuzzIL subroutine.
 public struct Parameters {
-    /// The total number of parameters.
     private let numParameters: UInt32
```
saelo, line 1063: What is this now? Is that the number of patterns or the number of variables needed for the parameters or something else? Maybe add a short comment.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1047,21 +1047,37 @@ final class TestIn: JsOperation {
...
-    /// Whether the last parameter is a rest parameter.
     let hasRestParameter: Bool
```
saelo, line 1064: Maybe leave a TODO here to investigate if we want to have this also be a pattern

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -1071,14 +1074,68 @@ public class JavaScriptCompiler {
...
+            }
+        }
+        for param in parameters {
+            let pattern = try processParameter(param)
+            patterns.append(pattern)
+        }
+        var params = Parameters(
+            count: totalParameterCount
+        )
+        params.patterns = patterns
```
saelo, line 1137: Nit: can the constructor take `patterns` and then keep them immutable (i.e. as a `let`) instead?

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1047,21 +1047,37 @@ final class TestIn: JsOperation {
...
 
-    /// The total number of parameters. This is equivalent to the number of inner outputs produced from the parameters.
     var count: Int {
```
saelo, line 1066: Same here, is that still the total number of inner outputs?

Sources/Fuzzilli/FuzzIL/JsOperations.swift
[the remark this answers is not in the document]
saelo: Nice! Yeah I think that's a really good solution here!

saelo (commented):
Really nice! And sorry for the slow review!

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -1071,14 +1074,68 @@ public class JavaScriptCompiler {
     }
 
     private func mapParameters(_ parameters: [Compiler_Protobuf_Parameter], to variables: ArraySlice<Variable>) {
```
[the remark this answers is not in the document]
TobiasWienand: Yes, absolutely correct. I have added descriptive comments.

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: No, we do not expect that. The case .none is autogenerated and switches need to be exhaustive. That is why this case is handled with just a break - it will not be reached anyway. I have included a fatal Error anyway for better clarity.

Sources/Fuzzilli/FuzzIL/JSTyper.swift
TobiasWienand: Good idea. I am wondering if being precise here is something that we must do. What happens if we do not specify details about an object ILType?

Sources/Fuzzilli/FuzzIL/JsOperations.swift
TobiasWienand: Unfortunately switching to a dictionary based approach causes problem because it does not contain information about the order of the object property keys. The dictionary tells us which value belongs to which key but not in which order the keys appear in the object pattern. This is why I opted for an array.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
TobiasWienand: Good catch, thanks for pointing that out. That is an artifact of a previous version where I did not have the enum and instead only saved indices of braces, which was not very maintainable/intuitive.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
TobiasWienand: Referenced this in a new issue

Sources/Fuzzilli/FuzzIL/JsOperations.swift
TobiasWienand: In a nutshell numParameters for f( a1, [ a2, a3 ] ) would be 3, not 2. So we are counting the parameters after destructuring - not before. I realize numParameters is probably not the appropriate name, that is why I have renamed it to numVariables.

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: Will work on this now.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
TobiasWienand: I hope this questions was answered with the newly added comment and renaming of the numParameters variable to numVariables.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
saelo: Thanks!

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -1355,15 +1355,33 @@ public class JavaScriptLifter: Lifter {
 
     private func liftParameters(_ parameters: Parameters, as variables: [String]) -> String {
```
saelo, line 1356: Maybe another useful test to add would be in the LifterTests, around here: https://github.com/googleprojectzero/fuzzilli/blob/3f81d5aa6e187f9211aa8be44cf8c121373a5776/Tests/FuzzilliTests/LifterTest.swift#L832

To try a few different types of parameters for functions and make sure that they lift to the expected JavaScript code.

Sources/Fuzzilli/FuzzIL/JSTyper.swift
TobiasWienand: Ah ok that answers the question

Sources/Fuzzilli/FuzzIL/JsOperations.swift
TobiasWienand: Technically yes. The order of the parameters doesn't matter. However the mapping from parameter -> variable is lost this way (IIUC). I.e. the order of variables is fixed but the order of object properties is permutated. 

We can do it anyway by using an array of tuples instead of a dictionary. This way we can retain the orderedness.

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: The more I think about it, the more I'm starting to feel this might be a bigger problem.

Changing the constructor requires updating all places in the codebase where we instantiate Patterns. Even if we use a default value for backward compatibility, shouldn’t we still review every instance in the entire codebase where Parameters is referenced and ask: “What should be the _patterns_ argument for the Parameters constructor?”

I've also noticed there are separate data structures, Parameter and ParameterList, which might also need revision—along with all their references in the codebase.

These extensive changes might be needed again once rest parameters and default values are supported.

The point I’m trying to make is that it feels like a lot of things might need to change for this, but the effort should also be justified and prioritized. I’m thinking it would make the most sense for me to focus on implementing the delete operator in Compilation and string object keys first, as they are likely easier to implement and will together enable hundreds of new seeds to become compilable.

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
TobiasWienand: Great idea for testing. Implemented this and it works.

In [advanced_param.js](https://github.com/TobiasWienand/fuzzilli/blob/d4169b71b553b085e292258e3e174750fc3d74c1/Tests/FuzzilliTests/CompilerTests/advanced_params.js) I also changed parameter names and invocation order to minimize the likelihood that lifting to JS is successful - just because these two things are ordered.

saelo: Great, thanks!

Sources/Fuzzilli/FuzzIL/JsOperations.swift
saelo: Right yeah that's a good point. It also needs to be consistent across serialization and deserialization (see https://github.com/googleprojectzero/fuzzilli/pull/449#issuecomment-2468034053) so I guess a list of tuples is the way to go here 👍

Sources/Fuzzilli/Compiler/Compiler.swift
saelo: +1 to working on delete operator and string object keys, those should allow for some quick wins!

I think it would make sense to have a convenience constructor of Parameters that creates a simple list of parameters: no patterns, no rest parameter, etc. just `function foo(a1, a2, a3) {`. Then we can have a second constructor for all the fancy stuff (patterns, rest param, etc.). I would think that the majority of use cases in Fuzzilli want the first constructor. Most of the time that should be the right thing to do. Only if you actually want to stress the logic for handling the special parameter types in the target engine would you want the other constructor. And that should then mostly be limited to the Compiler and a few CodeGenerators (or I guess the [`randomParameters()`](https://github.com/googleprojectzero/fuzzilli/blob/f6b1917fd6ed87ccb44c714483b6384b12e4eb11/Sources/Fuzzilli/Base/ProgramBuilder.swift#L425) helper function).

Regarding the ParameterList, I *think* this one doesn't need to change, but maybe its documentation should be updated. The way I see it, there are basically two use cases for the types of a function's parameters:
* The "outer" view: the argument types that a caller need to use, and
* The "inner" view: the types of the parameters that the callee can use

The ParameterList struct (which is part of a [Signature](https://github.com/googleprojectzero/fuzzilli/blob/f6b1917fd6ed87ccb44c714483b6384b12e4eb11/Sources/Fuzzilli/FuzzIL/TypeSystem.swift#L903)) is used for both of these cases. In both cases, we still only need a "flat" list of types: one for each argument or parameter. However, up until now, the two "views" would be identical: if the 2nd parameter was of type .integer, then that's what the caller had to pass and what the callee could assume. Now, for the caller the 2nd parameter might be of type `.object(withProperties: ["a", "b"])` while for the callee, the 2nd parameter might be of type `.integer` (because it's something like `function foo(p, {a: v3, b: v4})`. So the same function can have different ParameterLists for the two "views". I think it's fine to keep using the same data structures for that, just the distinction should maybe be described in a comment? Alternatively, we could try to force the distinction by using two different data structures. I'm not sure that will work well though, but might be worth a try. WDYT?

PR #455, opened 2024-10-24
Fix non-iteratrable input of yieldEach

Yi2255:
Logic:
1. Try to use existing variables and avoid creating new ones proactively, as this can lead to uninteresting and meaningless code.
2. If the variable passed to the current `CodeGenerator` is iterable, use it directly.
3. If it is not iterable, randomly generate an array composed of variables from the context, which, in the worst case, will still be empty, ensuring that what follows `yieldEach` is an iterable object, thus eliminating the need for additional `guard` checks.

### DIFF
--- a/Sources/Fuzzilli/CodeGen/CodeGenerators.swift
+++ b/Sources/Fuzzilli/CodeGen/CodeGenerators.swift
@@ -769,13 +769,13 @@ public let CodeGenerators: [CodeGenerator] = [
         // These are "typically" used as arguments, so we don't directly generate a call operation here.
     },
 
-    RecursiveCodeGenerator("GeneratorFunctionGenerator") { b in
+    RecursiveCodeGenerator("GeneratorFunctionGenerator", inputs: .preferred(.iterable)) { b, it in
         let f = b.buildGeneratorFunction(with: b.randomParameters(), isStrict: probability(0.1)) { _ in
             b.buildRecursive()
             if probability(0.5) {
                 b.yield(b.randomVariable())
             } else {
-                b.yieldEach(b.randomVariable())
+                b.yieldEach(b.buildIteratorVariable(b, it))
             }
             b.doReturn(b.randomVariable())
         }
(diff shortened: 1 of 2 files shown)


Sources/Fuzzilli/CodeGen/CodeGenerators.swift
```diff
@@ -1085,7 +1085,7 @@ public let CodeGenerators: [CodeGenerator] = [
             }
         } else {
             // TODO only do this when the value is iterable?
-            b.yieldEach(val)
+            b.yieldEach(b.buildIteratorVariable(b, val))
```
saelo, line 1088: I think here it would make sense to split this into two generators:

```
    CodeGenerator("YieldGenerator", inContext: .generatorFunction, inputs: .one) { b, val in
        assert(b.context.contains(.generatorFunction))
        if probability(0.9) {
            b.yield(val)
        } else {
            b.yield()
        }
    },

    CodeGenerator("YieldEachGenerator", inContext: .generatorFunction, inputs: .required(.iterable)) { b, val in
        assert(b.context.contains(.generatorFunction))
        b.yieldEach(val)
    },
```

That way it's easy to guarantee that we get a .iterable.

Sources/Fuzzilli/CodeGen/CodeGenerators.swift
```diff
@@ -769,13 +769,13 @@ public let CodeGenerators: [CodeGenerator] = [
...
 
-    RecursiveCodeGenerator("GeneratorFunctionGenerator") { b in
+    RecursiveCodeGenerator("GeneratorFunctionGenerator", inputs: .preferred(.iterable)) { b, it in
         let f = b.buildGeneratorFunction(with: b.randomParameters(), isStrict: probability(0.1)) { _ in
             b.buildRecursive()
             if probability(0.5) {
                 b.yield(b.randomVariable())
             } else {
-                b.yieldEach(b.randomVariable())
+                b.yieldEach(b.buildIteratorVariable(b, it))
```
saelo, line 778: Here and in the AsyncGeneratorFunctionGenerator, I would recommend doing this differently. For one, I think we would usually prefer to yield something that we create in this function (i.e. in the `buildRecursive()` call), but that's not possible if the `it` comes from the outside. I would probably recommend just doing something like this:
```
} else {
    let randomVariables = b.randomVariables(Int.random(in: 1...5))
    let array = b.createArray(with: randomVariables)
    b.yieldEach(array)
}
```
This will then cause us to always yield a plain array, but we have the other CodeGenerator below for yielding other stuff. Also the InputMutator will quickly shuffle this around to yield from other things.

saelo (commented):
Thanks!

PR #458, opened 2024-10-29
Increases assert error message coverage in parser.js

TobiasWienand:
This makes it easier to identify the most valuable targets for feature addition to compiler/parser by categorizing logs from insuccessful compilations based on assert message information


saelo:
Ah nice, thanks!

PR #460, opened 2024-11-09
Fixes Switch Break Bug

TobiasWienand:
Bug can be seen when trying to compile Tests/FuzzilliTests/CompilerTests/switch_statements.js, which fails because "break statement outside of loop or switch", even though it isn't. This happens because breaks are expected in switch cases, rather than the switch block. 

This was missed because compilation was succesful before moving the Analyzer changes to a separate branch and verification of the new ContextAnalyzer was only done with Tests/FuzzilliTests/AnalyzerTest.swift and not with Tests/CompilerTests/switch_statements.js


saelo:
Ah good catch! Yeah I guess this happens in part because the .switchBlock context is not propagated into the cases, but I think that needs to be this way, otherwise we might end up with SwitchCase blocks inside other SwitchCase blocks, which isn't valid. Anyway, thanks for the fix!

PR #461, opened 2024-11-13
Implements compilation of the delete operator

TobiasWienand:
This branches enables compilation of the delete operator. Some thoughts I think are worth mentioning

### 1) 

With this change, approximately 300 new seeds from test/mjsunit and the SpiderMonkey JIT test files will become compilable. If Fuzzilli treats the delete operation as a standard unary operator, delete might be applied to non-object properties, such as simple variables (var). If I understand correctly, semantic correctness will not be significantly impacted because delete does not cause a crash when called on unintended targets. For example, if applied to var, let, or parameters, it simply returns false and has no effect.

### 2) 

It can happen that an object has a key that is a number. In that case DeleteElement will be called, even though it is not an array.
Example:
```
const obj = { 1: 'value' };
delete obj[1];
```
Similarly, we can delete elements from an array based on a computed index - in that case DeleteComputedProperty is called, not DeleteElement.

Example
```
const arr = [1, 2, 3];
delete arr[Math.floor(Math.random() * arr.length)];
```
 Is that a problem? IIUC, the main difference between DeleteElement and Delete(Computed)Property In FuzzIL is only syntactical, not semantical.

### DIFF
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -983,9 +983,8 @@ public class JavaScriptCompiler {
             }
 
         case .unaryExpression(let unaryExpression):
-            let argument = try compileExpression(unaryExpression.argument)
-
             if unaryExpression.operator == "typeof" {
+                let argument = try compileExpression(unaryExpression.argument)
                 return emit(TypeOf(), withInputs: [argument]).output
             } else if unaryExpression.operator == "delete" {
                 guard case .memberExpression(let memberExpression) = unaryExpression.argument.expression else {
@@ -1031,6 +1030,7 @@ public class JavaScriptCompiler {
                 guard let op = UnaryOperator(rawValue: unaryExpression.operator) else {
                     throw CompilerError.invalidNodeError("invalid unary operator: \(unaryExpression.operator)")
                 }
+                let argument = try compileExpression(unaryExpression.argument)
                 return emit(UnaryOperation(op), withInputs: [argument]).output
             }
 
--- a/Sources/Fuzzilli/Compiler/Compiler.swift
+++ b/Sources/Fuzzilli/Compiler/Compiler.swift
@@ -987,11 +987,53 @@ public class JavaScriptCompiler {
 
             if unaryExpression.operator == "typeof" {
                 return emit(TypeOf(), withInputs: [argument]).output
+            } else if unaryExpression.operator == "delete" {
+                let argumentExpression = unaryExpression.argument.expression!
+
+                if case .memberExpression(let memberExpression) = argumentExpression {
+                    let obj = try compileExpression(memberExpression.object)
+                    let isGuarded = memberExpression.isOptional
+
+                    if memberExpression.name != "" {
+                        // Deleting a non-computed property (e.g., delete obj.prop)
+                        let propertyName = memberExpression.name
+                        let instr = emit(
+                            DeleteProperty(propertyName: propertyName, isGuarded: isGuarded),
+                            withInputs: [obj]
+                        )
+                        return instr.output
+                    } else {
+                        // Deleting a computed property (e.g., delete obj[expr])
+                        let propertyExpression = memberExpression.expression
+                        let propertyExpr = propertyExpression.expression
+                        let property = try compileExpression(propertyExpression)
+
+                        if case .numberLiteral(let numberLiteral) = propertyExpr {
+                            // Delete an element (e.g., delete arr[42])
+                            let index = Int64(numberLiteral.value)
+                            let instr = emit(
+                                DeleteElement(index: index, isGuarded: isGuarded),
+                                withInputs: [obj]
+                            )
+                            return instr.output
+                        } else {
+                            // Use DeleteComputedProperty for other computed properties (e.g., delete obj["key"])
+                            let instr = emit(
+                                DeleteComputedProperty(isGuarded: isGuarded),
+                                withInputs: [obj, property]
+                            )
+                            return instr.output
+                        }
+                    }
+                } else {
+                    throw CompilerError.invalidNodeError("Invalid argument for delete operator; expected a member expression")
+                }
+            } else {
+                guard let op = UnaryOperator(rawValue: unaryExpression.operator) else {
+                    throw CompilerError.invalidNodeError("invalid unary operator: \(unaryExpression.operator)")
+                }
+                return emit(UnaryOperation(op), withInputs: [argument]).output
             }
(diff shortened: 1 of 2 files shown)


Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -987,11 +987,53 @@ public class JavaScriptCompiler {
 
             if unaryExpression.operator == "typeof" {
                 return emit(TypeOf(), withInputs: [argument]).output
+            } else if unaryExpression.operator == "delete" {
+                let argumentExpression = unaryExpression.argument.expression!
+
+                if case .memberExpression(let memberExpression) = argumentExpression {
```
saelo, line 993: Nit: this might be nicer as a `guard`:
```
            guard case .memberExpression(let memberExpression) = unaryExpression.argument.expression else {
                throw CompilerError.invalidNodeError("unary operator must be applied to a member expression")
            }
```
I think then you also don't need to force-unwrap the expression

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -987,11 +987,53 @@ public class JavaScriptCompiler {
...
+                    let obj = try compileExpression(memberExpression.object)
+                    let isGuarded = memberExpression.isOptional
```
saelo, line 995: I think this could use a comment. I guess you are effectively turning
```
delete a?.b;
```
into
```
try { delete a.b; } catch {}
```
?
I think that makes sense, but a short comment might be nice :)

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -987,11 +987,53 @@ public class JavaScriptCompiler {
...
+
+                    if memberExpression.name != "" {
```
saelo, line 997: Maybe `!memberExpression.isEmpty` instead

saelo (commented):
Nice! Overall looks good already, with a few suggestions

saelo:
Regarding you comments:

1) So IIUC we currently cannot compile something like `delete a` because that's not a MemberExpression? It's probably not super important, but maybe it's easy enough to fix. We could either add a `DeleteName` instruction, or, probably even simpler, just compile `delete a` to a `delete this.a` (which IIUC is effectively what it's doing). WDYT?

2) I think this all works correctly the way you describe/implemented it. `delete x[1]` should always be a `DeleteElement` and `delete x[foo]` should always be a `DeleteComputedProperty`, regardless of what the object or the computed property name is.

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: Implemented

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: Here is an example of before and after:

### Before
```
const nestedObj = { a: { b: 2 } };
delete nestedObj.a?.c;
```

### After
```
const o1 = {
    "b": 2,
};
const o2 = {
    "a": o1,
};
o2.a?.c;
const t6 = o2.a;
delete t6?.c;
```

So, no try catch enveloping which is nice in my opinion :)

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: Do you mean `!memberExpression.name.isEmpty` ? Have implemented that

TobiasWienand:
Regarding 1), I initially had this in the code, but then I found out that an implicit global variable does not become a property of `this` in all engines. For example, in Node.js, it becomes a property of `global`. I am not sure how this is handled in other JS engines, so I avoided it. Also, another problem would be to distinguish between implicit global variables and `let, var, etc`. The data structure that is processed by Compiler.swift only says that it is an `identifier` and what name it has.

I did a quick check of how many files in test/mjsunit/* have `delete` but are not compilable. We are talking about 28 files, and therefore they are in position 12 on the priority list. So perhaps it is best not to invest time in this right now and come back to it later.

Tests/FuzzilliTests/CompilerTests/basic_delete.js
```diff
@@ -0,0 +1,22 @@
...
+const arr = [1, 2, 3];
+console.log(delete arr[1]);
+console.log(arr);
+
+const index = 0;
+console.log(delete arr[index]);
+console.log(arr);
+
+const nestedObj = { a: { b: 2 } };
+console.log(delete nestedObj?.a?.b);
```
saelo, line 21: Maybe also add a test for something like
```
try {
    delete null.a;
} catch(e) {
    console.log(e.message);
}
console.log(delete null?.a);
```

Sources/Fuzzilli/Compiler/Compiler.swift
saelo: Looks good! Though were does the `o2.a?.c;` line come from? Is that maybe a bug in our lifting, or is that property access also there in the IL?

saelo:
Cool, just a question and one more suggestion. And sounds good, let's not bother with `delete a` stuff for now :)

Sources/Fuzzilli/Compiler/Compiler.swift
TobiasWienand: Ah yes, good catch! This happened because 

` 
let argument = try compileExpression(unaryExpression.argument)
` 
was called, even though `argument` was not used in the delete case. This is fixed now. I will be looking more carefully at the FuzzIL translation in the future!

Tests/FuzzilliTests/CompilerTests/basic_delete.js
TobiasWienand: This is compilable. However, this would cause DeleteProperty to be called, even though null is not an object. In Compiler.swift, I believe it is currently not possible to check [here](https://github.com/TobiasWienand/fuzzilli/blob/530e5099881982589be546fa4c00af7c5a6e786f/Sources/Fuzzilli/Compiler/Compiler.swift#L994) if object is truly an object. We could either
- Try preventing this in parser.js, although there are possibly more edge cases than just this
- Ignore it and add a TODO
- Introduce a new DeleteAny JSOperation

What would you prefer?

saelo: No I think this is all working exactly as it should :) We  We should just assume that the LHS of the MemberExpression for `delete` is an object.

Sources/Fuzzilli/Compiler/Compiler.swift
saelo: Ah great thanks, and no worries, this is what code reviews are for :)

PR #462, opened 2024-11-13
Implements compilation of string literal keys

TobiasWienand:
Currently we support only this:

`const obj = { a : 1 };`

This is **not** supported:

`const obj = { "a" : 1 };`

Even though it is the same thing, just with a different representation. See this:
https://www.reddit.com/r/typescript/comments/1ac7b68/in_javascript_object_keysproperties_are_strings/?rdt=50433


saelo:
Ah very nice, thanks!

PR #463, opened 2024-11-14
Fix non-iteratrable input of yieldEach

Yi2255:
1. The input of `yieldEach` must be an iterable object, so separate `yieldEach` from YieldGenerator and use it as a separate code generator which requires iterable object.
2. For other code generators that need to use `yieldEach`, get random objects from the context, and then put those objects into an array as the input to `yieldEach`.


Sources/Fuzzilli/CodeGen/CodeGeneratorWeights.swift
```diff
@@ -91,6 +91,7 @@ public let codeGeneratorWeights = [
     "ConstructorGenerator":                     5,
     "SubroutineReturnGenerator":                3,
     "YieldGenerator":                           2,
```
saelo, line 93: While you're here, can you bump these weights to maybe 5 each? These generators will only run when we're inside a generator function, but then we probably do want to run them with a decent probability, so 2 seems a bit low (some day we might find a better way to assign these weights than manual estimates...)

Yi2255: Sure！

Sources/Fuzzilli/CodeGen/CodeGeneratorWeights.swift
```diff
@@ -89,8 +89,8 @@ public let codeGeneratorWeights = [
     "AsyncArrowFunctionGenerator":              1,
     "AsyncGeneratorFunctionGenerator":          1,
     "ConstructorGenerator":                     5,
-    "SubroutineReturnGenerator":                3,
-    "YieldGenerator":                           2,
+    "SubroutineReturnGenerator":                5,
```
saelo, line 92: Ah sorry, I meant the weights of `YieldGenerator` and `YieldEachGenerator` not the `SubroutineReturnGenerator` (that can probably stay at 3) 😅

Yi2255: Got it, thanks for the correction!

saelo:
Thanks a lot!

PR #464, opened 2024-11-14
feature/Nullish coalescing operator and assignment

Yi2255:
Add some feature about nullish coalescing operator and assignment

### DIFF
--- a/Sources/Fuzzilli/FuzzIL/JsOperations.swift
+++ b/Sources/Fuzzilli/FuzzIL/JsOperations.swift
@@ -1403,11 +1403,18 @@ public enum BinaryOperator: String, CaseIterable {
     case RShift   = ">>"
     case Exp      = "**"
     case UnRShift = ">>>"
+    // Nullish coalescing operator (??)
+    case NCO = "??"
 
     var token: String {
         return self.rawValue
     }
+
+    static public func allCaseWithoutNCO() -> Array<BinaryOperator> {
+        return BinaryOperator.allCases.filter { $0 != .NCO }
+    }
 }
+
 
 final class BinaryOperation: JsOperation {
     override var opcode: Opcode { .binaryOperation(self) }
--- a/Sources/Fuzzilli/Mutators/RuntimeAssistedMutator.swift
+++ b/Sources/Fuzzilli/Mutators/RuntimeAssistedMutator.swift
@@ -186,6 +186,7 @@ public class RuntimeAssistedMutator: Mutator {
         case LogicalAnd = "LOGICAL_AND"
         case LogicalOr = "LOGICAL_OR"
         case LogicalNot = "LOGICAL_NOT"
+        case NCO = "NCO"
         case BitwiseAnd = "BITWISE_AND"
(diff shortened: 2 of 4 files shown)


Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1403,12 +1403,19 @@ public enum BinaryOperator: String, CaseIterable {
...
     case Exp      = "**"
     case UnRShift = ">>>"
+    // Nullish coalescing operator (??)
+    case NCO = "??"
 
     var token: String {
         return self.rawValue
     }
+
+    static public func allCaseWithoutNCO() -> Array<BinaryOperator> {
```
saelo, line 1413: What's the idea behind this? Why is the new operator special?

Sources/Fuzzilli/Mutators/RuntimeAssistedMutator.swift
```diff
@@ -385,6 +386,8 @@ extension RuntimeAssistedMutator.Action {
             try translateBinaryOperation(.BitOr)
         case .BitwiseXor:
             try translateBinaryOperation(.Xor)
+        case .NCO:
+            try translateBinaryOperation(.NCO)
```
saelo, line 390: You will also need to update the corresponding JavaScript code, see here: https://github.com/googleprojectzero/fuzzilli/blob/d6f77d7b9425c21a0b36b6d2275c21cf5bba1ced/Sources/Fuzzilli/Lifting/JavaScriptRuntimeAssistedMutatorLifting.swift#L391

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -1403,12 +1403,19 @@ public enum BinaryOperator: String, CaseIterable {
     case RShift   = ">>"
     case Exp      = "**"
     case UnRShift = ">>>"
+    // Nullish coalescing operator (??)
+    case NCO = "??"
```
saelo, line 1407: 'NCO' is maybe a little weird because none of the other operators have "operator" in their name. 'NC' is maybe a bit short/unintuitive though, so how about just calling it `NullCoalesce`?

saelo (commented):
Thanks!
It would be nice to also add support in the compiler for this. See https://github.com/googleprojectzero/fuzzilli/pull/461 or similar PRs for examples. But also fine to land like this and maybe leave a TODO or file an issue :)

Sources/Fuzzilli/Mutators/RuntimeAssistedMutator.swift
Yi2255: Thanks very much. I missed something here.

Sources/Fuzzilli/FuzzIL/JsOperations.swift
Yi2255: Good point. Thanks!

saelo:
I guess this can be closed now after #471? If not, please reopen

PR #465, opened 2024-11-14
feature/void operator

Yi2255:
Add void operator feature. The specification ref [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void)

### DIFF
--- a/Sources/Fuzzilli/Base/JavaScriptLifterUtil.swift
+++ b/Sources/Fuzzilli/Base/JavaScriptLifterUtil.swift
@@ -1,0 +1,4 @@
+
+func insertVoid() -> String{
+    return probability(0.05) ? "void " : ""
+}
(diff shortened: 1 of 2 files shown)


saelo:
Thanks! So I think this isn't quite the right way to do this. For one, we generally need the output of lifting to be deterministic (i.e. the same FuzzIL program lifted twice must result in the same output text), and this breaks that (and you can see various tests failing due to that here: https://github.com/googleprojectzero/fuzzilli/actions/runs/11836407403/job/33049141843?pr=465). This is bad because now the same program might behave differently between executions and so cause unreliability. Also, this way only a few selected operations can be prefixed with `void`, while in theory, every expression can use it. I think instead the right way to do this would either be to add another unary operator (should be fairly easy) or another JSOperation (a bit more work, but should be fairly straightforward). A reason to go with a new JSOperation might be that this operation is probably fairly uninteresting (compared with the other unary operators) as it basically just means the engine will throw away the return value. So maybe that would be the better approach? WDYT?

Yi2255:
You are right, all expressions can use `void`. Using unary is a very clever way to do it. Thanks a lot for the correction!

PR #468, opened 2024-11-15
Feature/super expression

TobiasWienand:
With this, 7 new JSOperations become compilable:

1) CallSuperConstructor
2) SetSuperProperty
3) GetSuperProperty 
4) SetComputedSuperProperty
5) GetComputedSuperProperty
6) CallSuperMethod
7) UpdateSuperProperty

I verified correctness with [Tests/FuzzilliTests/CompilerTests/super.js
](https://github.com/TobiasWienand/fuzzilli/blob/72cbfe6df1b6d0eb4f6aaf823102f0441ea9b7fe/Tests/FuzzilliTests/CompilerTests/super.js)
Lifts to semantically equivalent code. (No side effects like with delete).

It is probably not a lot of effort to implement the remaining super cases but it is lower down the priority list.


Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -957,6 +994,18 @@ public class JavaScriptCompiler {
...
+            let (arguments, spreads) = try compileCallArguments(callSuperConstructor.arguments)
+            let isSpreading = spreads.contains(true)
+            
+            if isSpreading {
+                throw CompilerError.unsupportedFeatureError("Spread arguments are not supported in super constructor calls")
+            }
+            guard !callSuperConstructor.isOptional else {
+                throw CompilerError.unsupportedFeatureError("Optional chaining is not supported in super constructor calls")
+            }
+            return emit(CallSuperConstructor(numArguments: arguments.count), withInputs: arguments).output
```
saelo, line 1007: Is this why you need `CallSuperConstructor` to now return a value? I wonder if we should just do this instead:
```
// The "return" value of invoking the super constructor is just |this|
emit(CallSuperConstructor(numArguments: arguments.count), withInputs: arguments)
return emit(LoadThis())
```
WDYT?

saelo (commented):
Very nice, thanks! Just one question/suggestion

Sources/Fuzzilli/Compiler/Compiler.swift
saelo: Ah even simpler, I think you can just do `return lookupIdentifier("this");` since `this` must be available inside class constructors/methods (see https://github.com/googleprojectzero/fuzzilli/blob/0f521821d55f000acb5e63e9251274fc478ff456/Sources/Fuzzilli/Compiler/Compiler.swift#L226)

Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -957,6 +994,20 @@ public class JavaScriptCompiler {
...
+            let isSpreading = spreads.contains(true)
+            
+            if isSpreading {
+                throw CompilerError.unsupportedFeatureError("Spread arguments are not supported in super constructor calls")
+            }
+            guard !callSuperConstructor.isOptional else {
+                throw CompilerError.unsupportedFeatureError("Optional chaining is not supported in super constructor calls")
+            }
+            emit(CallSuperConstructor(numArguments: arguments.count), withInputs: arguments)
+            return lookupIdentifier("this")! // we can force unwrap because |this| always exists in the context where |super| exists
```
saelo, line 1008: Can you also add a short comment about why we're just returning `this` here? E.g. something like `// In JS, the result of calling the super constructor is just |this|, but in FuzzIL the operation doesn't have an output (because |this| is always available anyway)`

saelo:
Nice, thanks! 🎉

PR #469, opened 2024-11-17
Feature/await expression

TobiasWienand:


Sources/Fuzzilli/Compiler/Compiler.swift
```diff
@@ -1040,6 +1040,14 @@ public class JavaScriptCompiler {
         case .v8IntrinsicIdentifier:
             fatalError("V8IntrinsicIdentifiers must be handled as part of their surrounding CallExpression")
 
+        case .awaitExpression(let awaitExpression):
+                // TODO await is also allowed at the top level of a module
```
saelo, line 1044: Ah yeah that's a good point. That's probably a bigger project and we need to support module fuzzing for that, so leaving this TODO and throwing an error sounds good.

saelo:
Thanks!

PR #471, opened 2024-11-18
Support feature/nullish coalesce

Yi2255:
Add the new operator `NullCoalesce`

### DIFF
--- a/Sources/Fuzzilli/Lifting/JavaScriptRuntimeAssistedMutatorLifting.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptRuntimeAssistedMutatorLifting.swift
@@ -389,6 +389,7 @@ struct JavaScriptRuntimeAssistedMutatorLifting {
         const OP_LOGICAL_AND = 'LOGICAL_AND';
         const OP_LOGICAL_OR = 'LOGICAL_OR';
         const OP_LOGICAL_NOT = 'LOGICAL_NOT';
+        const OP_LOGICAL_NullCoalesce = 'LOGICAL_NullCoalesce';
 
         const OP_BITWISE_AND = 'BITWISE_AND';
         const OP_BITWISE_OR = 'BITWISE_OR';
@@ -511,6 +512,7 @@ struct JavaScriptRuntimeAssistedMutatorLifting {
           [OP_LOGICAL_AND]: (inputs) => inputs[0] && inputs[1],
           [OP_LOGICAL_OR]: (inputs) => inputs[0] || inputs[1],
           [OP_LOGICAL_NOT]: (inputs) => !inputs[0],
+          [OP_LOGICAL_NullCoalesce]: (inputs) => inputs[0] ?? inputs[1],
           [OP_BITWISE_AND]: (inputs) => inputs[0] & inputs[1],
           [OP_BITWISE_OR]: (inputs) => inputs[0] | inputs[1],
           [OP_BITWISE_XOR]: (inputs) => inputs[0] ^ inputs[1],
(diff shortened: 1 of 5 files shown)


saelo:
To clarify, what's the relationship between this PR and #464? Is this a replacement?

Yi2255:
Yes, this PR is a revised version based on your feedback on [#464.](https://github.com/googleprojectzero/fuzzilli/pull/464)

Sources/Fuzzilli/Mutators/RuntimeAssistedMutator.swift
```diff
@@ -186,6 +186,7 @@ public class RuntimeAssistedMutator: Mutator {
         case LogicalAnd = "LOGICAL_AND"
         case LogicalOr = "LOGICAL_OR"
         case LogicalNot = "LOGICAL_NOT"
+        case NullCoalesce = "NullCoalesce"
```
saelo, line 189: Let's stay consistent here, e.g. `NULL_COALESCE`

Sources/Fuzzilli/Lifting/JavaScriptRuntimeAssistedMutatorLifting.swift
```diff
@@ -389,6 +389,7 @@ struct JavaScriptRuntimeAssistedMutatorLifting {
         const OP_LOGICAL_AND = 'LOGICAL_AND';
         const OP_LOGICAL_OR = 'LOGICAL_OR';
         const OP_LOGICAL_NOT = 'LOGICAL_NOT';
+        const OP_LOGICAL_NullCoalesce = 'LOGICAL_NullCoalesce';
```
saelo, line 392: Same here, this should probably be `OP_NULL_COALESCE`

Sources/Fuzzilli/Lifting/JavaScriptRuntimeAssistedMutatorLifting.swift
```diff
@@ -511,6 +512,7 @@ struct JavaScriptRuntimeAssistedMutatorLifting {
           [OP_LOGICAL_AND]: (inputs) => inputs[0] && inputs[1],
           [OP_LOGICAL_OR]: (inputs) => inputs[0] || inputs[1],
           [OP_LOGICAL_NOT]: (inputs) => !inputs[0],
+          [OP_LOGICAL_NullCoalesce]: (inputs) => inputs[0] ?? inputs[1],
```
saelo, line 515: This is good, but to be useful, you also need to create these operations. I think mainly this is useful for the ExplorationMutator which creates new instructions depending on the runtime type of variables. Concretely, what I would suggest is replacing this here: https://github.com/googleprojectzero/fuzzilli/blob/5798638831dbc3347a0aa557f14ef63ebed704b3/Sources/Fuzzilli/Lifting/JavaScriptExploreLifting.swift#L250 and here: https://github.com/googleprojectzero/fuzzilli/blob/5798638831dbc3347a0aa557f14ef63ebed704b3/Sources/Fuzzilli/Lifting/JavaScriptExploreLifting.swift#L425 to call a new `exploreNullish` function which then generates this operation. Something like:
```
function exploreNullish(v) {
    // Best thing we can do with nullish values is a NullCoalescing (??) operation.
    return new Action(OP_NULL_COALESCE, [exploredValueInput, Inputs.randomArgument()])
}
```

Then this mutator should be able to generate these new operations :)

Yi2255: Thank you for your valuable advice. I did miss many details. I have modified and then pushed it.

saelo:
Cool, thanks a lot!

PR #472, opened 2024-11-18
feature/void

Yi2255:
Add a new UnaryOperator `void`


Sources/Fuzzilli/FuzzIL/JSTyper.swift
```diff
@@ -664,6 +664,8 @@ public struct JSTyper: Analyzer {
                 set(instr.output, .boolean)
             case .BitwiseNot:
                 set(instr.output, maybeBigIntOr(.integer))
+            case .Void:
+                set(instr.output, .anything)
```
saelo, line 668: This needs to be `.undefined` I think (the output of the void operator is always `undefined` IIRC)

saelo:
Nice! This looks good, I'm just wondering what happened to this commit for the compiler support https://github.com/googleprojectzero/fuzzilli/commit/bb872e1b2422540e7290db058c9eb2807c1f1170 @vi3tL0u1s? We can also just land this CL as it is and add the compiler support in a follow-up PR. WDYT?

vi3tL0u1s:
Hi @saelo, I’ve added the compiler support as you suggested. Thank you so much for the reminder!

saelo:
Ok great, thanks! I think there's a merge conflict though, and also I think there used to be a compiler test as well (https://github.com/googleprojectzero/fuzzilli/commit/bb872e1b2422540e7290db058c9eb2807c1f1170#diff-e3c74f57e01c1d64375830a153bdede28b838e44615cad0ba7085d5754390c2e), can you add that again, too?

vi3tL0u1s:
Hi @saelo, I’ve added the compiler test. Regarding the conflict, I believe it might be due to this PR not including the logic for the delete feature, which was merged earlier. I’m not entirely sure how to resolve this. Could you provide some guidance?

saelo:
Ah yes I see, let me try to figure out how to resolve that :)

saelo:
Ok, that worked. I guess I'll just merge these commits, then locally squash the merge commit, then force push. A bit hacky but oh well :)

saelo:
Okay done! I had to change the test a little bit because it ends up redefining the `output` constant from the prefix: https://github.com/googleprojectzero/fuzzilli/blob/d4796eddf8ebfc901dd11db610dc949d0226dbb2/Sources/Fuzzilli/Util/JavaScriptExecutor.swift#L21 but otherwise, this seems to have worked. Thanks! :)

vi3tL0u1s:
Thank you so much @saelo!

PR #483, opened 2024-12-17
bugfix/livetests

TobiasWienand:
The threshold in `LiveTests.swift` for the failure rate seems to be too low. The test fails or succeeds by chance. I have set the threshold to the mean plus three standard deviations, reducing the false positive rate to approximately 0.3%.


saelo:
Thanks! Carl, could you take a quick look at this? I think you've also recently looked at these failures

TobiasWienand:
I believe this was addressed with ad4b1ec208866e317c6ca4ab2518b160dbf23461

PR #486, opened 2024-12-25
Redesign named variables in FuzzIL

saelo:
This change replaces the three operations LoadNamedVariable, StoreNamedVariable, and DefineNamedVariable with a single new operation: CreateNamedVariable. This operation now simply creates a new FuzzIL variable that will be assigned a specific identifier during lifting. Optionally, the named variable can be declared using any of the available variable declaration modes: global, var, let, or const.

With this, we now have a single, flexible way of creating variables that have a specific name. We now use this:

* For builtins, which are effectively just existing named variables in the global scope. This also now makes it (easily) possible to overwrite builtins. As such, The LoadBuiltin operation was removed.
* During code generation, to sometimes create variables with specific names in generated code (we use random property names).
* In the compiler. This is the main user of named variables and this is where this change has the most impact: we now compiler _every_ variable declaration to a CreateNamedVariable operation. This now makes it possible to correctly compiler any code that relies on variable names, for example due to using `eval`, with statements, or similar constructs. See some of the added/modified tests for examples. The consequence of this is that compiled code will now often have a lot of CreateNamedVariable operations. However, as these are now just regular FuzzIL variables, this change should not significantly affect mutability of the programs. In the future, we could consider implementing a specific minimizer (that we could also run during corpus import) to remove unneeded CreateNamedVariable operations. However, it will likely be somewhat difficult to determine when such an operation is not needed.


saelo:
Had some time on the train today so I gave the idea from https://github.com/googleprojectzero/fuzzilli/pull/474#discussion_r1896696197 a try. @carl-smith  and @TobiasWienand, could you both take a look at this and see what you think? This should make named variables much more flexible and powerful and I think it should now make it easy to handle issues such as the one we're encountering in PR #474 now. But it's also a somewhat deep change and for example also removes `LoadBuiltin`.

TobiasWienand:
Overall it looks really awesome :) 
I noticed that
### before
```javascript
for (const a of ["a"]) {}
for (let b of ["b"]) {}
for (var c of ["c"]) {}
```
### after
```javascript
for (const v2 of ["a"]) {}
for (const v5 of ["b"]) {}
for (const v8 of ["c"]) {}
```

Apart from that, my small testbench regarding scoping of variables passed. I am currently working on a more comprehensive test bench for semantic correctness based on differential testing. I don't know if that already exists or has been done before but I think we will gain more confidence in any case. Will keep you updated

PR #496, opened 2025-02-03
feature/forawaitof

p0-tato:
I found that the current version of Fuzzilli does not generate `for await... of` statements. I implemented it by referring to MDN and the existing implementation of the for of statement within Fuzzilli. Additionally, I fixed a typo in `Sources/Fuzzilli/Compiler/Parser/parser.js`.

This is my first time making changes to Fuzzilli's code, so there might be mistakes. I would appreciate any feedback.


p0-tato:
The current PR contains commits with two separate purposes—one for typo corrections and another for modifications to the for-await-of behavior. Therefore, I will split them and submit a new PR.

PR #512, opened 2025-04-30
feature/loopNestedContinue

Yi2255:
Hi Saelo,

First of all, I sincerely apologize for the long delay regarding the implementation of `LabelStatement`.
Following your suggestion #479 , I have split this feature into two separate PRs, `LoopNestedContinue `and `NestedBreak`.
This PR implements `LoopNestedContinue`. As discussed in #479, I add a new JsOperation `loopNestedContinueand`  and a new CodeGenerator `LoopNestedContinueGenerator`. Beside, the code design in `JavaScriptLifter.swift` is as follows:
1. Use `LabelStack` to trace the nested code block. 
2. The element in` LabelStack` is `LabelPin`, which has two fields: `beginPos` and `hasLabel`. `beginPos` is the index of the starting point of the js code block, and `hasLabel ` indicates whether `beginPos ` has generated a label (with a small probability of generating N lines of break or continue in the same block). 
3. Because there is no need to label every code entry with a label, which will expand the code size, it is only inserted when `continueNested ` appears. Therefore, the core operation of `LabelStack ` is `insertLabel`, which has three operations: ① Insert label string into js code; ② mark that a label has been inserted here; ③Move the positions of all LabelPins behind the stack at the corresponding depth back by the corresponding size.

Please let me know if any changes are needed; Thanks a lot!

### DIFF
--- a/Sources/Fuzzilli/CodeGen/CodeGenerators.swift
+++ b/Sources/Fuzzilli/CodeGen/CodeGenerators.swift
@@ -54,7 +54,7 @@ public let CodeGenerators: [CodeGenerator] = [
     ValueGenerator("StringGenerator") { b, n in
         for _ in 0..<n {
             b.loadString(b.randomString())
-        }
+        }   
     },
 
     ValueGenerator("BooleanGenerator") { b, n in
@@ -1866,6 +1866,10 @@ public let CodeGenerators: [CodeGenerator] = [
             b.callFunction(f, withArgs: args)
         }, catchBody: { _ in })
     },
+
+    CodeGenerator("LoopNestedContinueGenerator", inContext: .loop) { b in
+        b.loopNestedContinue(Int.random(in: 0...10))
+    }
 ]
 
 extension Array where Element == CodeGenerator {
--- a/Sources/Fuzzilli/FuzzIL/JsOperations.swift
+++ b/Sources/Fuzzilli/FuzzIL/JsOperations.swift
@@ -2256,7 +2256,7 @@ final class BeginFinally: JsOperation {
 
 final class EndTryCatchFinally: JsOperation {
     override var opcode: Opcode { .endTryCatchFinally(self) }
-
+ 
     init() {
         super.init(attributes: [.isBlockEnd])
     }
@@ -2461,6 +2461,17 @@ class BindMethod: JsOperation {
         self.methodName = methodName
         // TODO(cffsmith): We probably want to expand this in the future to also bind arguments at some point.
         super.init(numInputs: 1, numOutputs: 1, requiredContext: .javascript)
+    }
+}
+
+final class LoopNestedContinue: JsOperation {
+    override var opcode: Opcode { .loopNestedContinue(self) }
+
+    let depth: Int
+
+    init(_ depth: Int) {
+        self.depth = depth
+        super.init(attributes: [.isJump], requiredContext: [.javascript, .loop])
     }
 }
 
--- a/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
+++ b/Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
@@ -101,6 +101,8 @@ public class JavaScriptLifter: Lifter {
 
         var w = JavaScriptWriter(analyzer: analyzer, version: version, stripComments: !options.contains(.includeComments), includeLineNumbers: options.contains(.includeLineNumbers))
 
+        var loopLabelStack = LabelStack(type: .loopblock)
+
         var wasmCodeStarts: Int? = nil
 
         if options.contains(.includeComments), let header = program.comments.at(.header) {
@@ -1150,14 +1152,19 @@ public class JavaScriptLifter: Lifter {
 
             case .beginWhileLoopBody:
                 let COND = handleEndSingleExpressionContext(result: input(0), with: &w)
+                loopLabelStack.push(currentCodeLength: w.code.count)
                 w.emitBlock("while (\(COND)) {")
                 w.enterNewBlock()
+
 
             case .endWhileLoop:
                 w.leaveCurrentBlock()
                 w.emit("}")
+                loopLabelStack.pop()
+
 
             case .beginDoWhileLoopBody:
+                loopLabelStack.push(currentCodeLength: w.code.count)
                 w.emit("do {")
                 w.enterNewBlock()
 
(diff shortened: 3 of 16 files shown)


Sources/Fuzzilli/CodeGen/CodeGenerators.swift
```diff
@@ -54,7 +54,7 @@ public let CodeGenerators: [CodeGenerator] = [
     ValueGenerator("StringGenerator") { b, n in
         for _ in 0..<n {
             b.loadString(b.randomString())
-        }
+        }
```
saelo, line 57: Nit unnecessary whitespace

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -2256,7 +2256,7 @@ final class BeginFinally: JsOperation {
 
 final class EndTryCatchFinally: JsOperation {
     override var opcode: Opcode { .endTryCatchFinally(self) }
-
+
```
saelo, line 2259: Nit same here

Sources/Fuzzilli/FuzzIL/JsOperations.swift
```diff
@@ -2464,6 +2464,17 @@ class BindMethod: JsOperation {
     }
 }
 
+final class LoopNestedContinue: JsOperation {
```
saelo, line 2467: Maybe here would be a good place for a short comment describing how this operation works, in particular how `depth` is interpreted (e.g. that it wraps around if there are fewer open loops)

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -1150,14 +1152,19 @@ public class JavaScriptLifter: Lifter {
 
             case .beginWhileLoopBody:
                 let COND = handleEndSingleExpressionContext(result: input(0), with: &w)
+                loopLabelStack.push(currentCodeLength: w.code.count)
                 w.emitBlock("while (\(COND)) {")
                 w.enterNewBlock()
 
+
```
saelo, line 1159: More whitespace here and below

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -101,6 +101,8 @@ public class JavaScriptLifter: Lifter {
 
         var w = JavaScriptWriter(analyzer: analyzer, version: version, stripComments: !options.contains(.includeComments), includeLineNumbers: options.contains(.includeLineNumbers))
 
+        var loopLabelStack = LabelStack(type: .loopblock)
```
saelo, line 104: Can we `assert` at the end of the lifting that this stack is now empty? Just to catch any cases where we forget the `pop()`

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -2268,4 +2304,89 @@ public class JavaScriptLifter: Lifter {
...
+
+    // Every possible position of label
+    struct LabelPin {
+        var beginPos: Int
+        var hasLabel: Bool
+    }
+
+    enum LabelType {
+        case loopblock
+        case ifblock
```
saelo, line 2316: I guess for this PR, only `loopBlock` would be supported? The others should probably move to the corresponding PR

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -2268,4 +2304,89 @@ public class JavaScriptLifter: Lifter {
...
+
+            stack[index].hasLabel = true
+
+            let delta = labelContent.count
+            for i in index+1..<stack.count {
+                stack[i].beginPos += delta
+            }
+        }
+
+        mutating func translateBreakLabel(w: inout ScriptWriter, expDepth: Int){
```
saelo, line 2368: Nit: missing space before `{` (also below)

Sources/Fuzzilli/Lifting/ScriptWriter.swift
```diff
@@ -84,4 +84,12 @@ struct ScriptWriter {
         assert(currentIndention.count >= indent.count)
         currentIndention.removeLast(indent.count)
     }
+
+    mutating func insert(_ pos: Int, _ content: String){
+        if code.index(code.startIndex, offsetBy: pos, limitedBy: code.endIndex) != nil {
```
saelo, line 89: Can this ever not be the case? Wouldn't that be a bug?

Tests/FuzzilliTests/LifterTest.swift
```diff
@@ -3095,4 +3095,35 @@ class LifterTests: XCTestCase {
         """
         XCTAssertEqual(actual, expected)
     }
-}
+
+    func testLoopNestedContinueLifting(){
```
saelo, line 3099: I'm wondering, does the label insertion work correctly if we also print line numbers? Did this test: https://github.com/googleprojectzero/fuzzilli/blob/f31876fff984ef7050adcdbe7f4c8bb3255ed8d0/Tests/FuzzilliTests/LifterTest.swift#L51 ever fail? Maybe you could add an extra test like this one that inserts labels but also prints line numbers?

saelo: If this change breaks line numbering, maybe we could just add the label to the same line as the loop begin? So lift to something like `label1: for (...) {` instead?

Sources/Fuzzilli/Lifting/JavaScriptLifter.swift
```diff
@@ -2268,4 +2304,89 @@ public class JavaScriptLifter: Lifter {
...
+        case loopblock
+        case ifblock
+        case switchblock
+        case tryblock
+        case codeBlock
+        case withblock
+    }
+
+    /// A structure that manages label positions within a specific control flow block (e.g., loop, if, switch, etc.).
+    public struct LabelStack {
```
saelo, line 2324: I wonder if this struct should actually be part of the ScriptWriter or the JavaScriptWriter itself. It feels a bit awkward that you need to pass in the ScriptWriter or internal state of the script writer (the current position). I think this would all go away if the ScriptWriter/JavaScriptWriter itself maintained this stack. And I think it would also remove the need for the `withScriptWriter` method above. Would that work? That way, the ScriptWriter also wound't need to expose the `insert` method which probably shouldn't be used for any other purposes.

saelo (commented):
Thanks a lot! Did a first round of review now

PR #519, opened 2025-08-02
refactor: move common parts from (V8Profile|V8HoleFuzzingProfile|V8SandboxProfile).swift to V8CommonProfile.swift

simbahax:
Multiple profiles for V8 fuzzing exist. 
Currently, common functionality is re-implemented in every profile, therefore there is an open todo in some of the profiles.

This PR implements this to ease the implementation of new profiles :) 

Point for discussion: The namespace is somewhat polluted here, e.g.,  other profiles also implement a `GcGenerator`.
I prefixed these where required, there might be better ways around that.


simbahax:
(rebased to current main)

Sources/FuzzilliCli/Profiles/V8CommonProfile.swift
```diff
@@ -0,0 +1,570 @@
...
+    // `gc()` takes a `type` parameter. If the value is 'async', gc() returns a
+    // Promise. We currently do not really handle this other than typing the
+    // return of gc to .undefined | .jsPromise. One could either chain a .then
+    // or create two wrapper functions that are differently typed such that
+    // fuzzilli always knows what the type of the return value is.
+    b.callFunction(gc, withArgs: b.findOrGenerateArguments(forSignature: b.fuzzer.environment.type(ofBuiltin: "gc").signature!)) 
+}
+
+// Insert random GC calls throughout our code.
+public let V8GcSandboxGenerator = CodeGenerator("GcGenerator") { b in
```
saelo, line 55: I think we can drop this one in favor of the above. I don't think we intended to have two different versions of this generator

saelo (commented):
Nice, thanks! I think this looks good, just one comment.

saelo:
Thanks a lot!

PR #535, opened 2025-10-17
Pass environment variables to the engine in startup tests

### LINKED ISSUE #534: Environment variable `SHM_ID` is not passed to the engine in startup tests
Fuzzing SpiderMonkey currently produces this warning at the beginning:

```plain-text
[Fuzzer] Initialized
[Fuzzer] Recommended timeout: [...]
[Fuzzer] Runner has non-empty output for empty program! This might indicate that some flags are wrong.
[Fuzzer] Output:
[COV] no shared memory bitmap available, skipping
[COV] edge counters initialized. Shared memory: (null) with 398403 edges
```

This is because the environment variable `SHM_ID` is not passed to the engine when running startup tests. While this does not currently pose a problem, this could be bad if some engine depends on the existence of shared memory for coverage in the future.

Related code:

https://github.com/googleprojectzero/fuzzilli/blob/11402cc6e9eb21237a5acbfe2465a936939d78df/Sources/Fuzzilli/Util/JavaScriptExecutor.swift#L93-L99

CSharperMantle:
This PR helps avoid potential mismatches in behavior between a startup test and subsequent fuzzing based on REPRL sessions.

After this patch, the startup tests on SpiderMonkey produce the following output:

```plain-text
[Fuzzer] Initialized
[Fuzzer] Recommended timeout: at least 1000ms. Current timeout: 10000ms
[Fuzzer] Runner has non-empty output for empty program! This might indicate that some flags are wrong.
[Fuzzer] Output:
[COV] edge counters initialized. Shared memory: shm_id_230610_0 with 398403 edges
```

Resolves #534.

### DIFF
--- a/Sources/Fuzzilli/Util/JavaScriptExecutor.swift
+++ b/Sources/Fuzzilli/Util/JavaScriptExecutor.swift
@@ -38,9 +38,12 @@ public class JavaScriptExecutor {
 
     let arguments: [String]
 
+    let env: [(String, String)]
+
     /// Depending on the type this constructor will try to find the requested shell or fail
-    public init?(type: ExecutorType = .any, withArguments maybeArguments: [String]? = nil) {
+    public init?(type: ExecutorType = .any, withArguments maybeArguments: [String]? = nil, withEnv maybeEnv: [(String, String)]? = nil) {
         self.arguments = maybeArguments ?? []
+        self.env = maybeEnv ?? []
         let path: String?
 
         switch type {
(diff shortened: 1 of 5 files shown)


Sources/Fuzzilli/Util/JavaScriptExecutor.swift
```diff
@@ -89,13 +93,19 @@ public class JavaScriptExecutor {
         // Close stdin
         try inputPipe.fileHandleForWriting.close()
 
+        let environment = ProcessInfo.processInfo.environment.merging(
+            Dictionary(env, uniquingKeysWith: { _, new in new }),
```
Liedtke: This shouldn't have to be a `Dictionary`?
```suggestion
            env,
```

PR #544, opened 2026-01-23
Add more recent builtin additions

linusg:
Adds builtins from the following stage 4 proposals:

- https://github.com/tc39/proposal-array-from-async
- https://github.com/tc39/proposal-float16array
- https://github.com/tc39/proposal-math-sum
- https://github.com/tc39/proposal-promise-try
- https://github.com/tc39/proposal-regex-escaping


Sources/Fuzzilli/Environment/JavaScriptEnvironment.swift
```diff
@@ -1771,6 +1772,7 @@ public extension ObjectGroup {
             "getUint16"  : [.integer] => .integer,
             "getInt32"   : [.integer] => .integer,
             "getUint32"  : [.integer] => .integer,
+            "getFloat16" : [.integer] => .float,
```
Liedtke: It is not enough to add it to the `ObjectGroup`. It will also need to be added to the `ILType` (around line 1072).
This may seem like an unnecessary duplication (and I think it is) but that's how it works right now. :)

